PR #21, opened 2015-03-28
Fix bug with typed_fad_gradient!()

krjackvinator:
Current implementation of `typed_fad_gradient!()` contains a bug related to access to non-existent field of a structure and is inconsistent with `typed_fad_gradient()`. When trying to call `forwarddiff_jacobian!()` of a function that returns `Vector`, an error occurs. In this example:

```Julia
using ForwardDiff

F(x) = [x[1]^2, x[1]*x[2], x[2]^2]

J! = forwarddiff_jacobian!(F, Float64, fadtype=:typed)

A = zeros(3, 2)
J!([1., 1.], A)
```
after execution got
```
`g!` has no method matching g!(::Array{Float64,1}, ::Array{Float64,2})
````

If try to make `A=zeros(6)` got
```
type Array has no field g
```

After that fix, one can use `Matrix` of the same shape as in the output of `forwarddiff_jacobian()`.

### DIFF
--- a/src/typed_fad/GraDual.jl
+++ b/src/typed_fad/GraDual.jl
@@ -170,7 +170,7 @@ atanh{T<:Real, n}(x::GraDual{T, n}) = GraDual{T, n}(atanh(x.v), x.g/(1-x.v*x.v))
 function typed_fad_gradient!{T<:Real}(f::Function, ::Type{T})
   function g!(x::Vector{T}, gradient_output::Matrix{T})
     fvalue = f(GraDual(x))
-    m = size(gradient_output)[1]
+    m = length(fvalue)
     for i in 1:m
       gradient_output[i, :] = fvalue[i].g[:]
     end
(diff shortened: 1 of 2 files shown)


scidom:
Thanks for the fix @krjackvinator, I will look at this in a few days as I am currently travelling.

scidom:
@krjackvinator apologies for taking me a while to look at this, do you want to fix your PR so that it doesn't fail, and then I can check it and merge it?

krjackvinator:
@scidom i apologize for not replying for so long time(i had been working on my graduate thesis). I had to change tests a bit for `forwarddiff_jacobian!()` because this function should take type `Array{Float64, 2}` instead of `Array{Float64, 1}`(it doesn't know if argument function returns one value or array of `Float64`). Unfortunately, now this PR contains merge conflicts. I could merge master with my branch, but then commits from master will get into PR, so i'd better wait for your decision.

jrevels:
This now fixed on `master` as of the merging of #27.

PR #23, opened 2015-04-16
Fix #20: Add isless(a,b) for GraDual type

### LINKED ISSUE #20: A definition of isless for GraDual?
Shouldn't GraDual{T} be comparable to T so that `isless(graDual, x) == isless(graDual.v, x)`?

Otherwise it's not possible to differentiate `x > 0 ? x : -x`

vrld:
Add isless(x1::GraDual, x2) and isless(x1, x2::GraDual) so that functions taking a minimum, maximum, etc are differentiable.


scidom:
@vrld I will merge this; I was thinking of changing 

```
isless(x1::GraDual, x2) = isless(value(x1), x2)
```

to 
```
isless(x1::GraDual, x2::Number) = isless(value(x1), x2)
```

I will do this after merging.

PR #25, opened 2015-05-29
Fix #24

### LINKED ISSUE #24: Bug in `typed_fad_gradient`
Consider the following function 

```
using ForwardDiff;
srand(1)
y = randn(100);
x = randn(100,2);
g(theta)  = x.*(y-x*theta);
gn(theta) = vec(sum(g(theta), 1));
gn([.1,.1])
```

I want to use `typed_fad_gradient` to calculate the derivative of `gn` with respect to `theta`. This code 

```
    h1 = ForwardDiff.typed_fad_gradient(gn, Float64);
    h1([.1,.1])
```
provides the correct gradient
``` 
 2x2 Array{Float64,2}:
  -103.869     11.1779
    11.1779  -119.393
```
as it can be seen by comparing it with the analytic gradient
```
    -x'x
```
	
Suppose now the function to be derived is defined using a linear algebra
operation, i.e.,

```
 p = ones(100);
 wgn(theta) = g(theta)'p;
 h2 = ForwardDiff.typed_fad_gradient(wgn, Float64);
 h2([.1,.1])
```
Although `wgn([.1,.1]).==gn([.1,.1])`, the gradient returned by `type_fad_gradient` is
```
 2x2 Array{Float64,2}:
  103.869   -11.1779
  -11.1779  119.393  
```
which is different from `h1([.1,.1])` and from `-x'x` by a factor of `-1`.

gragusa:
I tracked the problem mentioned on #24 on the lack of a transpose/ctranspose methods for GradDual. I added those. Test still pass, but I am not sure I am doing it right. (The problem with #24 is gone though).

### DIFF
--- a/src/typed_fad/GraDual.jl
+++ b/src/typed_fad/GraDual.jl
@@ -67,6 +67,16 @@ isless(x1::Number, x2::GraDual) = isless(x1, value(x2))
 
 show(io::IO, x::GraDual) = print(io, "GraDual(", value(x), ",\n", grad(x), ")")
 
+function transpose{T<:Real, n}(x::GraDual{T, n})
+  g = Array(T, n)
+  for i in 1:n
+    g[i] = x.g[i]
+  end
+  GraDual{T, n}(x.v, g)
+end
+
+ctranspose{T<:Real, n}(x::GraDual{T, n}) = transpose(x)
+
 function conj{T<:Real, n}(x::GraDual{T, n})
   g = Array(T, n)
   for i in 1:n


scidom:
Thanks, @gragusa. I think you need to correct only one thing; use `g = Array(T, 1, n)` instead of `g = Array(T, n)`. So, your code would be:

```
function transpose{T<:Real, n}(x::GraDual{T, n})
  g = Array(T, 1, n)
  for i in 1:n
    g[i] = x.g[i]
  end
  GraDual{T, n}(x.v, g)
end

ctranspose{T<:Real, n}(x::GraDual{T, n}) = transpose(x)
```

But then, looking more closely, my suggestion can't work, because we have defined the `g` field of a `GraDual` to be a vector, and by transposing it we turn it to a matrix in Julia terms. Your definition on the other hand works, though I am trying to convince myself that it is reasonable to define transposition in a way that does not adhere with the more general way Julia treats vector transposition.

scidom:
P.S. An alternative option would be to define the `g` field of `GraDual` as `g::Union(Vector{T}, Matrix{T})` but I am not sure this is a good idea.

scidom:
P.S.2: On a second note, your definition makes sense if we decide to consider/define transposition only in terms of the scalar value field `v` of a `GraDual`, in a similar way we confined ourselves on the `v` field in the case of functions such as `isless`. Then we can proceed using your definition.

gragusa:
@scidom I get your points --- but beside style, what do you think are the side effects of my definition? I can't see any, but that's just me being near sighted I guess!

gragusa:
My stance is that derivatives involving  linear algebra ops boil down to sums of scalar linear functions. As such, considering --- as you said --- the transposition in terms of the scalar value field v of a GraDual

scidom:
I get your point @gragusa. My main consideration is that there is no natural way of defining uniquely some concepts for `GraDual`, same goes for other functions such as `isless`, not only transposition. There is some ambiguity since there is no obvious mathematical interpretation or setting for such definitions (why for example not transpose the gradient field `g` along with `v` or why to transpose it - that's the ambiguity I am referring to). That being said, I will merge your PR, given that my consideration is not specific to transposition and it has previously being surpassed when it come to ordering of `GraDual` objects.

PR #38, opened 2015-08-17
RFC: Add Partials type to separate ForwardDiffNumber implementations from container implementation

jrevels:
This cleans up a lot of the code that was previously in the `gradnum_math` folder by adding a layer of abstraction over the container type used for `GradientNumber`s. In addition, [I ran these benchmarks](https://github.com/JuliaDiff/ForwardDiff.jl/blob/master/benchmarks/benchmarks.jl#L20) and saw between a 10% and 30% speedup when taking the gradient compared to master.

This abstraction layer may also be useful in the future for providing `HessianNumber`s (or even `TensorNumber`s) with `Tuple` support, though that's not really the point of this PR.

@mlubin This is essentially ready for merging as soon as I see that tests pass and code coverage hasn't been messed up by it (I think this is the implementation we should benchmark against for this week).


mlubin:
Is there a quick explanation for why this is faster?

jrevels:
Well, the process of separating everything out causes the code to do a lot less "manual" type-inferencing (before, the code was doing a lot of type promotion/calling inner constructors, which you can see happening all over [here](https://github.com/JuliaDiff/ForwardDiff.jl/commit/873f7d3b95a753bfa108958470cb36699a61338d#diff-fc0ff6be0044f3f461cc21198fe3303c)). I don't know why exactly, but maybe the cleaner code is easier for Julia's codegen to reason about? Normally, I would guess that master might've had some "superfluous" conversions happening by accident, and maybe this PR reduces their occurrence, but I really didn't think there were any extra conversions going on behind-the-scenes in the first place.

The division-related code is much cleaner as well, and possibly faster, but that wouldn't have been reflected by the `ackley_sum` benchmark...

mlubin:
Okay, looks good to merge

PR #39, opened 2015-08-18
draft benchmark with c++

mlubin:
This implements the ``ackley`` function in C++ with gradient calculations via dual numbers, which corresponds to ``chunk_size=1``.
On my laptop:
```
0.000392914 sec for function value
9.04767 sec for gradient
```

compared with
```
julia> bench_fad(ackley,xlens=16000,repeat=2)
7x4 DataFrames.DataFrame
| Row | time       | func | xlen  | chunk_size |
|-----|------------|------|-------|------------|
| 1   | 0.00044988 | 'f'  | 16000 | -1         |
| 2   | 1.88277    | 'g'  | 16000 | 0          |
| 3   | 21.625     | 'g'  | 16000 | 1          |
| 4   | 12.0676    | 'g'  | 16000 | 2          |
| 5   | 8.3323     | 'g'  | 16000 | 4          |
| 6   | 6.14265    | 'g'  | 16000 | 8          |
| 7   | 5.4309     | 'g'  | 16000 | 16         |
```

So our function evaluations are about the same, and the base dual number code is a factor of two off of C. That's pretty respectable but there might be room in the future to squeeze a bit more performance out of the tuple-based implementation. It's surprising how much faster the vector implementation is.

### DIFF
--- a/benchmarks/benchmarks.cpp
+++ b/benchmarks/benchmarks.cpp
@@ -1,0 +1,129 @@
+#include <iostream>
+#include <cmath>
+#include <vector>
+#include <sys/time.h>
+
+using namespace std;
+
+// g++ -std=c++11 -O2 -o benchmarks benchmarks.cpp /path/to/julia/usr/lib/libopenlibm.a
+
+
+double clock_now()
+{
+    struct timeval now;
+    gettimeofday(&now, NULL);
+    return (double)now.tv_sec + (double)now.tv_usec/1.0e6;
+}
+
+class Dual {
+    double real;
+    double eps;
+    public:
+    Dual(double real, double eps) : real(real), eps(eps) {}
+    Dual(double real) : real(real), eps(0.0) {}
+    Dual() : real(0.0), eps(0.0) {}
+
+    friend ostream& operator<<(ostream& os, const Dual& dt) {
+        os << "Dual(" << dt.real << "," << dt.eps << ")";
+        return os;
+    }
+
+    Dual operator+(const Dual& rhs) const {
+        return Dual(real+rhs.real,eps+rhs.eps);
+    }
+
+    Dual operator-(const Dual& rhs) const {
+        return Dual(real-rhs.real,eps-rhs.eps);
+    }
+
+    Dual operator*(const Dual& rhs) const {
+        return Dual(real*rhs.real, real*rhs.eps + eps*rhs.real);
+    }
+
+    double real_part() const { return real; }
+    double eps_part() const { return eps; }
+
+};
+
+Dual operator*(double x, const Dual &rhs) {
+    return Dual(x*rhs.real_part(), x*rhs.eps_part());
+}
+
+Dual sin(const Dual& x) {
+    return Dual(sin(x.real_part()), x.eps_part()*cos(x.real_part()));
+}
+
+Dual cos(const Dual& x) {
+    return Dual(cos(x.real_part()), -x.eps_part()*sin(x.real_part()));
+}
+
+Dual sqrt(const Dual& x) {
+    return Dual(sqrt(x.real_part()), x.eps_part()/(2*sqrt(x.real_part())));
+}
+
+Dual exp(const Dual& x) {
+    double expval = exp(x.real_part());
+    return Dual(expval, x.eps_part()*expval);
+}
+
+/*
+function ackley(x)
+    a, b, c = 20.0, -0.2, 2.0*π
+    len_recip = inv(length(x))
+    sum_sqrs = zero(eltype(x))
+    sum_cos = sum_sqrs
+    for i in x
+        sum_cos += cos(c*i)
+        sum_sqrs += sqr(i)
+    end
+    return (-a * exp(b * sqrt(len_recip*sum_sqrs)) -
+            exp(len_recip*sum_cos) + a + e)
+end
+*/
+
+template<typename T> T ackley(const vector<T> &x) {
+    double a = 20.0, b = -0.2, c = 2*M_PI;
+    double len_recip = 1.0/x.size();
+    T sum_sqrs(0);
+    T sum_cos(0);
+
+    for (auto i : x) {
+        sum_cos = sum_cos + cos(c*i);
+        sum_sqrs = sum_sqrs + i*i;
+    }
+    return (-a * exp(b * sqrt(len_recip*sum_sqrs)) -
+            exp(len_recip*sum_cos) + a + M_E);
+}
+
+
+int main() {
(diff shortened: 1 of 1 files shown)


jrevels:
> That's pretty respectable but there might be room in the future to squeeze a bit more performance out of the tuple-based implementation. It's surprising how much faster the vector implementation is.

To see where we might improve, it would be cool to compare with Optim's `autodiff` vs. ForwardDiff with `chunk_size=1`. We should be able to make them equivalently fast, but I don't know whether they are at the moment.

jrevels:
Did you want to merge this now? Feel free to merge whenever you're ready.

jrevels:
The results I'm getting on my laptop are quite different...then again, I'm a very rookie C++ user:

```zsh
➜  benchmarks git:(cppbench) g++ benchmarks.cpp
benchmarks.cpp:90:10: warning: 'auto' type specifier is a C++11 extension [-Wc++11-extensions]
    for (auto i : x) {
         ^
benchmarks.cpp:90:17: warning: range-based for loop is a C++11 extension [-Wc++11-extensions]
    for (auto i : x) {
                ^
2 warnings generated.

➜  benchmarks git:(cppbench) ✗ ./a.out #ran a couple of times, this was the best result
20
0.000397921 sec for function value
25.7708 sec for gradient
```

My julia benchmarks are fairly close to yours, though:

```julia
julia> bench[:ackley]
21x4 DataFrames.DataFrame
| Row | time        | func | xlen  | chunk_size |
|-----|-------------|------|-------|------------|
| 1   | 2.74e-6     | 'f'  | 16    | -1         |
| 2   | 5.7223e-5   | 'g'  | 16    | 0          |
| 3   | 5.6552e-5   | 'g'  | 16    | 1          |
| 4   | 4.9823e-5   | 'g'  | 16    | 2          |
| 5   | 4.6942e-5   | 'g'  | 16    | 4          |
| 6   | 4.5825e-5   | 'g'  | 16    | 8          |
| 7   | 4.4029e-5   | 'g'  | 16    | 16         |
| 8   | 4.1718e-5   | 'f'  | 1600  | -1         |
| 9   | 0.0328962   | 'g'  | 1600  | 0          |
| 10  | 0.171796    | 'g'  | 1600  | 1          |
| 11  | 0.0966284   | 'g'  | 1600  | 2          |
| 12  | 0.0679323   | 'g'  | 1600  | 4          |
| 13  | 0.0484395   | 'g'  | 1600  | 8          |
| 14  | 0.0420936   | 'g'  | 1600  | 16         |
| 15  | 0.000400089 | 'f'  | 16000 | -1         |
| 16  | 1.87736     | 'g'  | 16000 | 0          |
| 17  | 17.0998     | 'g'  | 16000 | 1          |
| 18  | 9.59618     | 'g'  | 16000 | 2          |
| 19  | 6.74976     | 'g'  | 16000 | 4          |
| 20  | 4.80788     | 'g'  | 16000 | 8          |
| 21  | 4.1689      | 'g'  | 16000 | 16         |
```

jrevels:
And if you're curious, here are the results I got from AlgoPy:

```python
>>> df = pandas.DataFrame.from_dict(bench['ackley'])
>>> df
  func       time   xlen
0    f   0.000056     16
1    g   0.001969     16
2    f   0.004461   1600
3    g   0.549358   1600
4    f   0.044596  16000
5    g  89.519750  16000
```

mlubin:
Make sure you use the compile statement in the comment at the top of the
file.

jrevels:
Ah, didn't see that, thanks.

```
➜  benchmarks git:(cppbench) ✗ ./benchmarks
20
0.000397921 sec for function value
4.74544 sec for gradient
```

So on my machine, it seems like ForwardDiff is off by a factor of ~3.

mlubin:
What methods does algopy use?
Why is the function evaluation 100x slower but gradient evaluation isn't?

jrevels:
> What methods does algopy use?
> Why is the function evaluation 100x slower but gradient evaluation isn't?

The relevant class in AlgoPy is [here](https://github.com/b45ch1/algopy/blob/master/algopy/utpm/utpm.py), the methods for initializing/extracting the gradient [here](https://github.com/b45ch1/algopy/blob/master/algopy/utpm/utpm.py#L1678). 

It appears that the formulation is based on storing a rank 4 tensor and then have the overloaded methods act on it via calls to numpy. Pretty cool stuff, this is actually similar to what I had in mind with my `PartialsNum` idea we discussed a while back, but obviously actually fleshed out. 

So instead of the formulation being rooted in the algebraic properties of dual numbers,  it's rooted in "univariate Taylor polynomial matrix arithmetic". There's some slides from the main author of the package [here](https://pythonhosted.org/algopy/_downloads/walter_euroad2010_paderborn_univariate_taylor_polynomial_arithmetic_applied_to_matrix_factorizations_in_the_forward_and_reverse_mode_of_algorithmic_differentiation.pdf) which go into a bit of detail on the subject (I'm guessing you know of it, but of course it's new to me).

mlubin:
Looks like a vector forward mode then?

mlubin:
So we should be comparing with our vector forward mode, where we're about 50x faster.

jrevels:
> Looks like a vector forward mode then?

Yes, and I believe by extension their forward-mode Hessian/tensor calculations are roughly equivalent to our own when `N == length(x)`.

> So we should be comparing with our vector forward mode, where we're about 50x faster.

Yeah, that's what I was looking at when making comparisons. Our "slowdown" ratio for `ackley` is worse than AlgoPy's by a factor of 2. Could that factor of 2 be coming from the fact that the calls to numpy probably end up falling back to C, or do you think it's a deeper issue?

This makes me want to consider benchmarking the slowdown ratio of our Hessian calculation vs. AlgoPy's. As a future side project, it might be worth it to prototype a UTPM implementation in Julia to see how it stacks up to the dual number implementation. If it's just as performant (or more so), it might be worth switching to - it appears to generalize to higher orders far more easily (obviously that's not immediately relevant to these benchmarks).

mlubin:
> Could that factor of 2 be coming from the fact that the calls to numpy probably end up falling back to C, or do you think it's a deeper issue?

That factor of 2 is probably coming from the fact that scalar operations in python are super slow, while the gradient calculation is using vector operations. So in the gradient mode, the relative overhead of python is smaller.

jrevels:
> That factor of 2 is probably coming from the fact that scalar operations in python are super slow, while the gradient calculation is using vector operations. So in the gradient mode, the relative overhead of python is smaller.

Ah, that makes sense, thanks.

PR #40, opened 2015-09-01
Allow access to lower-order values by providing a `ForwardDiffResult`s interface

### LINKED ISSUE #37: Offer a way for users to access the lower-order values that end up being calculated by FAD methods
Previously discussed [here](https://github.com/JuliaDiff/ForwardDiff.jl/pull/27#issuecomment-130843307).

For example, if somebody calls `hessian(f,x)` they should be able to tell the function to return `∇f(x)` and/or `f(x)` as well, since those values are already calculated by way of the Hessian calculation.

Here some ideas for what this might look like:

1) Have a keyword argument that specifies that the function should return the calculated `ForwardDiffNumber`, then provide methods in the API to extract the data you want from it. Something like this:

```julia
julia> data = hessian(f, x, alldata=true) # returns ForwardDiffNumber
julia> gradient(data) # extract gradient from data
julia> hessian(data) # extract hessian from data
```

I don't really like `alldata` as a name for the keyword argument, but you get my drift.

2) Pass symbols to the method saying what you want up front (once again, via a keyword argument):

```julia
# returns in the same order as it's given, with Hessian first 
julia> hess, val, grad = hessian(f, x, also=(:value, :gradient)) 
```

Thoughts?

jrevels:
This PR resolves #37.

### What it does:

* Adds a `ForwardDiffResult` wrapper type which users can interact with without having to understand `ForwardDiffNumber`s

* Centralizes all result-handling methods under `ForwardDiffResult`

* If the abstract type ~~`AllInfo`~~ `AllResults` is passed in to API methods, the "full" results of the calculation are returned along with the normal output. For example:

    ```julia
    julia> using ForwardDiff
    
    julia> f(x) = 3 * sum(x)^5;
    
    julia> x = [1,2,3,4];
    
    julia> f(x)
    300000
   
    julia> ForwardDiff.gradient(f, x) # normal calling behavior is unchanged
    4-element Array{Int64,1}:
    150000
    150000
    150000
    150000

    julia> out, allresults = ForwardDiff.gradient(f, x, AllResults);
    
    julia> out
    4-element Array{Int64,1}:
    150000
    150000
    150000
    150000
    
    julia> value(allresults)
    300000
    
    julia> ForwardDiff.gradient(allresults)
    4-element Array{Int64,1}:
    150000
    150000
    150000
    150000
    ```

    ~~I don't particularly like `AllInfo` as the keyword for triggering this behavior. It's not horrible, but can anybody come up with something better?~~ I changed it to `AllResults`, which I'm still not really a fan of but can live with (unless somebody comes up with a better name).

* Not strictly necessary for this PR, but I changed the closure-generating methods so that all "configuration" now occurs in the call to the generator rather than in calls to the closures. After sitting on the old behavior for a while, this feels like the more intuitive way to do things. Also, taking optional arguments out of the closures' signatures appears to help type inference for some unknown reason.

### To Do:

- [x] Implement `ForwardDiffResult` and related methods
- [x] Refactor `derivative!`/`derivative`
- [x] Refactor `gradient!`/`gradient`
- [x] Refactor `jacobian!`/`jacobian`
- [x] Refactor `hessian!`/`hessian`
- [x] Refactor `tensor!`/`tensor`

### DIFF
--- a/src/ForwardDiff.jl
+++ b/src/ForwardDiff.jl
@@ -42,7 +42,7 @@ module ForwardDiff
     include("TensorNumber.jl")
     include("fad_api/fad_api.jl")
 
-    export AllInfo,
+    export AllResults,
            ForwardDiffCache,
            value,
            value!,
--- a/src/ForwardDiff.jl
+++ b/src/ForwardDiff.jl
@@ -42,7 +42,11 @@ module ForwardDiff
     include("TensorNumber.jl")
     include("fad_api/fad_api.jl")
 
-    export derivative!,
+    export AllInfo,
+           ForwardDiffCache,
+           value,
+           value!,
+           derivative!,
            derivative,
            gradient!,
            jacobian!,
@@ -50,7 +54,6 @@ module ForwardDiff
            hessian!,
            hessian,
            tensor!,
-           tensor,
-           ForwardDiffCache
+           tensor
 
 end # module ForwardDiff
--- a/src/TensorNumber.jl
+++ b/src/TensorNumber.jl
@@ -123,14 +123,14 @@ end
 #   t₁ = grad(t, i) # coeff of ϵ₁
 #   t₂ = grad(t, j) # coeff of ϵ₂
 #   t₃ = grad(t, k) # coeff of ϵ₃
-#   t₄ = hess(t, a) = hess(t, t_inds_2_h_ind(i, j)) # coeff of ϵ₁ϵ₂
-#   t₅ = hess(t, b) = hess(t, t_inds_2_h_ind(i, k)) # coeff of ϵ₁ϵ₃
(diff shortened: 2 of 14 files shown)


mlubin:
Seems pretty reasonable

jrevels:
@mlubin @scidom @KristofferC I changed `AllInfo` to `AllResults`, which is only slightly less vague...anybody have a better name for this? If not, this PR should be good to merge. Documentation will be slightly out of date once this lands, but I'm planning on tackling new docs in a subsequent PR (hopefully the last PR before the new version of ForwardDiff.jl can be tagged and released).

mlubin:
I'm okay with the name, can always be deprecated if someone comes up with a better one.

KristofferC:
This is perfect for me. 

This package is really awesome!

PR #41, opened 2015-09-03
User Documentation

jrevels:
This PR provides extensive user documentation for the latest version of this package.

It's written in ReST and built using Sphinx. My intention was to put these docs on [readthedocs](https://readthedocs.org/) (RTD), a popular documentation platform for Julia projects. However, RTD still hosts the old documentation for this package under @scidom's account (see [here](https://readthedocs.org/projects/forwarddiffjl/); you may notice that I recently triggered a new build to see what would happen, but it only rebuilt the old docs).

I don't know the right way propagate this repo's information to RTD when I'm not the account holder for the project...is just set up to reflect what's on the master branch? I only have a little prior experience with RTD, so any advice would be great.


mlubin:
We've actually been migrating away from RTD for various reasons. @IainNZ, any suggestions?

IainNZ:
My money for the future is actually http://www.mkdocs.org/ hosted on a gh-pages branch for the package (generated manually offline). Thats Markdown though...

jrevels:
I can translate this to Markdown if that's the route we want to go. I'm more familiar with MkDocs than I  am with Sphinx anyway.

Funny story, I had never written ReST before this. Even if it doesn't get used, I have to say, I'm pretty impressed with ReST as a markup syntax. C'est la vie.

@mlubin Just let me know what you think is the best way forward. I've don't have any experience making GitHub project pages, but deploying the docs via GitHub rather than having to deal with a whole different site sounds pretty appealing.

jrevels:
Also, it seems to be the case that we could, in fact, build the current Sphinx documentation using the RTD theme, then simply deploy that from a `gh-pages` branch. That's my vote, anyway.

jrevels:
Just pushed the build from the `docs` branch to a [`gh-pages` branch](https://github.com/JuliaDiff/ForwardDiff.jl/tree/gh-pages) and the it appears to work quite nicely; go to [http://www.juliadiff.org/ForwardDiff.jl/](http://www.juliadiff.org/ForwardDiff.jl/) to see for yourself.

mlubin:
Looks pretty good to me, who needs RTD. How did you get juliadiff.org working?

IainNZ:
Yeah thats the way to got @jrevels. @mlubin I think its just the default behaviour for orgs on Github.

mlubin:
Oh, fancy

jrevels:
Yup, what @IainNZ said. Should I merge this now, then?

Once this is merged, I'm basically ready to officially tag and release ForwardDiff v0.1.0. How do we want to handle versioning? Just stick with tags or have separate (also tagged) branches (e.g. `release-0.1`, `release-0.2`, etc.)?

The more relevant question for this PR: How do we maintain separate doc versions if we go the `gh-pages` route? Maybe we don't even want to bother with that...if people *really* need it, they can always build the documentation from an earlier version themselves.

mlubin:
Not sure there's a need for versioned docs or branches at this point. A ``NEWS.md`` file would be helpful though.

jrevels:
Sounds good.

> A NEWS.md file would be helpful though.

Maybe we should just have a `News` section at the bottom of the README?

mlubin:
Sure, whichever you prefer

mlubin:
@scidom, could you give me admin access to the ForwardDiff readthedocs page? My username is mlubin.

mlubin:
@jrevels, may want to mention at the top of the readme that only 0.4 is supported, and installing the package on 0.3 will get you an old version.

jrevels:
> @jrevels, may want to mention at the top of the readme that only 0.4 is supported, and installing the package on 0.3 will get you an old version.

Done.

PR #45, opened 2015-09-07
Jacobian of functions of the form `f!(output, x)`

### LINKED ISSUE #44: Differentiation of functions of the form `f!(x, y)` where `x` is the input and `y` is the output
If memory serves, a pervious version of ForwardDiff supported functions that looked like this:

```julia
function sphere2cart!(xin::Vector, xout::Vector)
    rho, theta, phi = xin
    rho_sin_phi = rho * sin(phi)
    xout[1] = rho_sin_phi * cos(theta)
    xout[2] = rho_sin_phi * sin(theta)
    xout[3] = rho * cos(phi)
end
```

However now (according to the docs, anyway), it seems I am only allowed to write:

```julia
function sphere2cart(xin::Vector)
    rho, theta, phi = xin
    rho_sin_phi = rho * sin(phi)
    x = rho_sin_phi * cos(theta)
    y = rho_sin_phi * sin(theta)
    z = rho * cos(phi)
    return [x, y, z]
end
```

Sadly, this second form is considerably slower, due to the allocation of the result vector.  Is there any way take the Jacobian of a vector-valued function written using the *result placement* style?

jrevels:
This PR resolves #44.

### DIFF
--- a/docs/source/perf_diff.rst
+++ b/docs/source/perf_diff.rst
@@ -1,6 +1,21 @@
 Performing Differentiation
 ==========================
 
+Restrictions on the target function
+-----------------------------------
+
+ForwardDiff.jl can only differentiate functions that adhere to the following rules:
+
+- **The function must be univariate** (an exception to this is found in the ``jacobian`` method below).
+
+- **The function's argument types cannot be too restrictively annotated.** In this case, "too restrictive" means more restrictive than ``x::Vector`` or ``x::Number``.
+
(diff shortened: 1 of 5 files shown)


docs/source/perf_diff.rst
```diff
@@ -1,6 +1,21 @@
 Performing Differentiation
 ==========================
 
+Restrictions on the target function
+-----------------------------------
+
+ForwardDiff.jl can only differentiate functions that adhere to the following rules:
+
+- **The function must be univariate** (an exception to this is found in the ``jacobian`` method below).
```
mlubin: Univariate usually means scalar-valued input. Taking only one argument (which itself could be a vector) is something else, I think.

PR #51, opened 2015-09-11
Dev Docs - Overview and Types

jrevels:
This PR provides some much-needed developer documentation.

I've currently addressed the basics of the ForwardDiffNumber types, and added a small section for explaining why Julia is useful for FAD (@mlubin will probably recognize this).

~~I still want to explain the basics of how the API works, and the caching layer. After covering those two topics, I think this will be ready to merge.~~ I'm going to merge this as-is, and cover further topics in new PRs.

### DIFF
--- a/docs/source/types.rst
+++ b/docs/source/types.rst
@@ -123,7 +123,7 @@ A ``HessianNumber`` can be intepreted as an ensemble of hyper-dual numbers. The 
       |       |       |        /        |        \       |       |      |       |
     c₁ϵ₁ϵ₂  c₂ϵ₁ϵ₂  c₃ϵ₁ϵ₂  c₄ϵ₁ϵ₂    c₅ϵ₁ϵ₂   c₆ϵ₁ϵ₂  c₇ϵ₁ϵ₂  c₈ϵ₁ϵ₂  c₉ϵ₁ϵ₂  c₉ϵ₁ϵ₂
 
-Each root-to-leaf path represents an individual hyper-dual number. Labeling the indices of :math:`b` values on the second level with :math:`i`, and those on the third level with :math:`j`, the definition for an individual hyper-dual number :math:`h_{ij}` (where :math:`i \geq j`) in the above tree is:
+Each root-to-leaf path represents an individual hyper-dual number. Labeling the indices of :math:`b` values on the :math:`\epsilon_1` level with :math:`i`, and those on the :math:`\epsilon_2` level with :math:`j`, the definition for an individual hyper-dual number :math:`h_{ij}` (where :math:`i \geq j`) in the above tree is:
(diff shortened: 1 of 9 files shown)


docs/source/types.rst
```diff
@@ -0,0 +1,231 @@
...
+
+    i \leq j \leq N
+
+    i \leq k \leq j
+
+    M = \frac{N(N+1)(N+2)}{6}
+
+    d_p \in \{d_1, d_2, d_3...d_{M-1}, d_M\}
+
+    p_{ijkN} &= k + \left[\sum_{\alpha=i}^{j}\alpha-i+1\right] + \left[\sum_{\alpha=1}^{i-1} \sum_{\beta=\alpha}^{N}\beta-\alpha+1\right] \\
```
jrevels: This line appears to be incorrect. It's supposed to be formula for going from (i,j,k) --> p in the loop

```julia
function loop(N)
    p = 1
    for i in 1:N
        for j in i:N
            for k in i:j
                println((i,j,k), "-->", p)
                p += 1
            end
        end
    end
end
```

Must've made a math error somewhere...

KristofferC:
Would it be possible to give some advice to what someone who wants to start hacking on this should focus on. Any performance bottle necks that would be good to look at our possible new features?  I always had the experience that just a few lines of what to look into can greatly reduce the barrier for a first commit.

jrevels:
> Would it be possible to give some advice to what someone who wants to start hacking on this should focus on.

I think having a "How to Contribute" section is a good idea, thanks for the suggestion. Specifically, there's a lot of low-hanging fruit for anybody willing to implement new functions on ForwardDiffNumbers, or hand-optimize some of the auto-generated functions. I'll be sure to include a short tutorial for how to do so.

docs/source/types.rst
```diff
@@ -0,0 +1,231 @@
...
+.. _`hyper-dual numbers`: https://adl.stanford.edu/hyperdual/Fike_AIAA-2011-886.pdf
+
+``Partials{T,C}`` 
+-----------------
+
+.. code-block:: julia
+    
+    immutable Partials{T,C}
+        data::C
+        Partials{N}(data::NTuple{N,T}) = new(data)
```
mlubin: Why do these need to be inner constructors?

mlubin:
Did a quick read of the text and LGTM

docs/source/types.rst
jrevels: To enforce `T == eltype(C)`. It also allows gives newcomers to the code an idea of what `C` is actually supposed to be. I could've used outer constructors here, I guess, but then the `T == eltype(C)` restriction wouldn't be inherent to the type definition.

PR #52, opened 2015-09-16
make norm on vectors of ForwardDiffs work

KristofferC:
This fixes that `norm` on a vector of ForwardDiff numbers no longer returns a float.

Before:

```julia
a = ForwardDiff.GradientNumber{9,Float64,Tuple{Float64,Float64,Float64,Float64,Float64,Float64,Float64,Float64,Float64}}(2.0202020202020195e8,ForwardDiff.Partials{Float64,Tuple{Float64,Float64,Float64,Float64,Float64,Float64,Float64,Float64,Float64}}((1.5384615384615384e11,-7.69230769230769e10,-7.69230769230769e10,0.0,0.0,0.0,0.0,0.0,0.0)));

norm([a,a,a])

# 3.4990925405431855e8
```

Now:

```jl
norm([a,a,a])

# ForwardDiff.GradientNumber{9,Float64,Tuple{Float64,Float64,Float64,Float64,Float64,Float64,Float64,Float64,Float64}}(3.4990925405431855e8,ForwardDiff.Partials{Float64,Tuple{Float64,Float64,Float64,Float64,Float64,Float64,Float64,Float64,Float64}}((2.6646935501059653e11,-1.3323467750529822e11,-1.3323467750529822e11,0.0,0.0,0.0,0.0,0.0,0.0)))

norm(a_vec) == sqrt(dot(a_vec, a_vec))

# true
```

Maybe

```jl
float(n::ForwardDiffNumber) = n
```

should be a bit better and try to convert all components to float?

### DIFF
--- a/src/ForwardDiff.jl
+++ b/src/ForwardDiff.jl
@@ -1,5 +1,5 @@
 module ForwardDiff
-    
+
     if VERSION < v"0.4-"
         warn("ForwardDiff.jl is only officially compatible with Julia v0.4-. You're currently running Julia $VERSION.")
     end
@@ -7,11 +7,11 @@ module ForwardDiff
     import Calculus
     import NaNMath
     import Base: *, /, +, -, ^, getindex, length,
-                 hash, ==, isequal, copy, zero, 
-                 one, rand, convert, promote_rule, 
-                 read, write, isless, isreal, isnan, 
-                 isfinite, eps, conj, transpose, 
-                 ctranspose, eltype, abs, abs2, start, 
+                 hash, ==, isequal, copy, zero,
+                 one, rand, convert, promote_rule,
+                 read, write, isless, isreal, isnan,
+                 isfinite, isinf, eps, conj, transpose,
+                 ctranspose, float, eltype, abs, abs2, start,
                  next, done, atan2
 
     const supported_unary_funcs = map(first, Calculus.symbolic_derivatives_1arg())
--- a/src/ForwardDiffNumber.jl
+++ b/src/ForwardDiffNumber.jl
@@ -5,11 +5,11 @@ abstract ForwardDiffNumber{N,T<:Number,C} <: Number
 #    eltype(::Type{F}) --> T from ForwardDiffNumber{N,T,C}
 #    value(n::F) --> the value of n
 #    grad(n::F) --> a container corresponding to all first order partials
-#    hess(n::F) --> a container corresponding to the lower 
-#                   triangular half of the symmetric 
+#    hess(n::F) --> a container corresponding to the lower
+#                   triangular half of the symmetric
 #                   Hessian (including the diagonal)
-#    tens(n::F) --> a container of corresponding to lower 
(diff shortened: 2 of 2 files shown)


mlubin:
Yes, ``float`` should convert a ForwardDiffNumber of integers into a ForwardDiffNumber of floats.

jrevels:
There's actually a different bug going on here...`float(a)` is already correctly defined by the provided `convert` methods, and in this instance, should be throwing an `InexactError`. I just found this and [fixed it on master](https://github.com/JuliaDiff/ForwardDiff.jl/commit/495b018c20aeef398fabe61a516f319b121e6f2c).

On the most recent master:

```julia
julia> using ForwardDiff

julia> a = ForwardDiff.GradientNumber{9,Float64,Tuple{Float64,Float64,Float64,Float64,Float64,Float64,Float64,Float64,Float64}}(2.0202020202020195e8,ForwardDiff.Partials{Float64,Tuple{Float64,Float64,Float64,Float64,Float64,Float64,Float64,Float64,Float64}}((1.5384615384615384e11,-7.69230769230769e10,-7.69230769230769e10,0.0,0.0,0.0,0.0,0.0,0.0)));

julia> norm([a,a,a])
ERROR: InexactError()
 in float at float.jl:121
 in generic_vecnormInf at linalg/generic.jl:94
 in generic_vecnorm2 at linalg/generic.jl:111
 in vecnorm at linalg/generic.jl:168
 in norm at linalg/generic.jl:184
```

This is the behavior I expected to see when looking at this PR, but of course you're right in that `norm` should work without error. The key thing to fix here is that the float conversion is occurring at all...

Fixing this behavior might come down to writing a more generic `norm` definition in Base, or at least overloading it in ForwardDiff.jl to be more generic.

jrevels:
> Yes, float should convert a ForwardDiffNumber of integers into a ForwardDiffNumber of floats.

I'm not sure this is the right way to go. The return type of `float` should be `T<:AbstractFloat`. It's even defined that way in the docs:

```julia
help?> float
search: float float64 float32 float16 Float32 Float16 Float64 FloatRange FloatingPoint float64_isvalid float32_isvalid

  float(x)

  Convert a number, array, or string to a AbstractFloat data type. For numeric data, the smallest suitable
  AbstractFloat type is used. Converts strings to Float64.
```

mlubin:
``float`` isn't the same as ``convert(Float64,...)``. ``float([1,2,3])`` doesn't return an ``AbstractFloat``

jrevels:
Oh, you're right. I guess the docs should be cleared up then...

In that case, yes, I'm all for `float{N}(::ForwardDiffNumber{N})` being a conversion to a ForwardDiffNumber of floats.

KristofferC:
Okay, I'm not super confident with how to do the conversion best so I'll leave it to you.

jrevels:
> Okay, I'm not super confident with how to do the conversion best so I'll leave it to you.

Let's track this in #53, then. Thanks for finding this and bringing it up, @KristofferC!

PR #53, opened 2015-09-16
WIP: Add `float` and `isinf` methods for ForwardDiffNumbers

jrevels:
This PR resolves #52.

It's basically done, but I tagged it a WIP because I still need to add tests before I'm ready to merge this.

### DIFF
--- a/src/ForwardDiff.jl
+++ b/src/ForwardDiff.jl
@@ -4,13 +4,16 @@ module ForwardDiff
         warn("ForwardDiff.jl is only officially compatible with Julia v0.4-. You're currently running Julia $VERSION.")
     end
 
+    ###########
+    # imports #
+    ###########
     import Calculus
     import NaNMath
     import Base: *, /, +, -, ^, getindex, length,
                  hash, ==, isequal, copy, zero, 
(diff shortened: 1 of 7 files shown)


jrevels:
The most recent build failed due to how `abs` is currently defined for `GradientNumber`s (not directly related to this PR, but popped up on that build due to lucky RNG):

```julia
Testing GradientNumbers and ForwardDiff.gradient...ERROR: LoadError: LoadError: test failed: 
ForwardDiff.GradientNumber{3,Float64,Tuple{Float64,Float64,Float64}}(-0.0,ForwardDiff.Partials{Float64,Tuple{Float64,Float64,Float64}}((-0.62,-0.84,-0.7))) 
== 
ForwardDiff.GradientNumber{3,Float64,Tuple{Float64,Float64,Float64}}(0.0,ForwardDiff.Partials{Float64,Tuple{Float64,Float64,Float64}}((0.62,0.84,0.7)))
```

The current definition for `abs` is:

```julia
@inline abs(g::GradientNumber) = (value(g) >= 0) ? g : -g
```

I think I'll change it to use `signbit` instead.

PR #60, opened 2015-09-28
Refactor promotion/conversion rules for ForwardDiffNumbers

### LINKED ISSUE #59: Calculation of the gradient of a function that contains derivatives of other function freezes
The minimal example that i could find:

```julia
julia> f(x) = x/2

julia> dfdx = derivative(f)
d (generic function with 1 method)

julia> g(x) = dfdx(x[1])*f(x[2])

julia> ForwardDiff.gradient(g, [1., 2.])
# Does not return, freezes
```

Note that this bug doesn't appear if `f` is identity function or something like `2x`, also when `g` takes only one argument everything is OK, but the bug appears in all other, more complex cases. I use Julia  `v0.4.0-rc2`, ForwardDiff `v0.1.0`.

jrevels:
This PR resolves #59.

### DIFF
--- a/src/ForwardDiff.jl
+++ b/src/ForwardDiff.jl
@@ -33,16 +33,28 @@ module ForwardDiff
     include("TensorNumber.jl")
     include("api/api.jl")
 
-    ###########################
-    # Misc. Utility Functions #
-    ###########################
+    #######################
+    # Promotion Utilities #
+    #######################
     @inline switch_eltype{T,S}(::Type{Vector{T}}, ::Type{S}) = Vector{S}
     @inline switch_eltype{N,T,S}(::Type{NTuple{N,T}}, ::Type{S}) = NTuple{N,S}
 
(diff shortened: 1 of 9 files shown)


jrevels:
To summarize this PR a bit better:

- Previously, results of `GradientNumber` arithmetic would be fully promoted, which led to trouble when one tried to promote the type of a `GradientNumber` with a nested `GradientNumber` of differing `npartials` (see #59)

- Promoting to `Vector` storage with `Tuple` storage is now allowed, in case the user calls nested API functions in different modes (i.e. vector-mode on a nested chunk-mode function). Mixing modes will likely result in some really slow code, but at least it will work (I still want to add tests for this before merging this PR).

- I gave the other `ForwardDiffNumber`'s conversion/promotion code a quick bath to ensure it would work with the above changes, and realized it could be coded more generally/succinctly if conversion/promotion correctly worked in the Partials layer, so I implemented that.

- I ran our benchmarks, and didn't see any sign of regression from these changes.

PR #62, opened 2015-09-29
RFC: Precompile ALL the things

KristofferC:
This is mostly an experiment to see how much time can be saved on avoiding compiling commonly used functions at runtime. I ran [SnoopCompile](https://github.com/timholy/SnoopCompile.jl) on the test files in `ForwardDiff` to generate a list of functions target for precompilation.

Using this example script (taken straight from the README):

```jl
using ForwardDiff
x = rand(5);
f(x::Vector) = sum(sin, x) + prod(tan, x) * sum(sqrt, x);
@time g = ForwardDiff.gradient(f);
@time g(x);
@time j = ForwardDiff.jacobian(g);
@time j(x);
@time ForwardDiff.hessian(f, x);
```

The times before this PR is

```
0.016770 seconds (12.88 k allocations: 606.140 KB)
0.306491 seconds (381.09 k allocations: 17.221 MB, 1.26% gc time)
0.013600 seconds (12.01 k allocations: 564.342 KB)
0.331584 seconds (277.92 k allocations: 12.411 MB, 1.27% gc time)
0.151430 seconds (168.46 k allocations: 7.667 MB)
```

and after

```
0.012115 seconds (704 allocations: 35.953 KB)
0.215817 seconds (243.39 k allocations: 11.462 MB, 2.15% gc time)
0.009060 seconds (491 allocations: 24.859 KB)
0.233938 seconds (128.73 k allocations: 5.882 MB)
0.109585 seconds (94.10 k allocations: 4.241 MB, 3.47% gc time)
```

Now, is doing something like this even worth it at all? The extra time to compile the package seems to be quite low, it might however be a nuisance to keep parity between the API and `precompile.jl`

I don't expect this to be merged, just food for thought.


jrevels:
Thanks for trying this out, it's definitely interesting!

> Now, is doing something like this even worth it at all?

I think that snoop-precompilation would be most useful in the case that a naive user has a very costly function that they only wish to run once, and don't understand how to properly "warm up" the function and weed out the JIT overhead themselves. A counterpoint to this is that the JIT overhead for ForwardDiff's functions would probably by negligible compared to the JIT overhead/runtime of the target function itself. 

If we find a couple of reasonable benchmarks where snoop-precompiling vastly improves first-call performance in such a case, I'm game to try it out. 

> it might however be a nuisance to keep parity between the API and precompile.jl

Yeah...it also wrecks the code coverage heuristic. Is there any way to ignore this file for coverage purposes?

KristofferC:
Sorry, I didn't see you had commented on this.

For now, I don't think we should go down this route since ForwardDiff seems to not have any significant problems with compilation times. Please comment if you think otherwise and I could rebase it,

Issue #63, opened 2015-10-10
jacobians with output_length takes a long time to create


KristofferC:
Some background of my use case: 
I need to solve a smallish nonlinear problem in a lot of points where each points have a set of different parameters. For a specific point I make a closure from the parameters of the point and use `ForwardDiff` to compute the jacobian of that closure. I then pass the closure and the jacobian to an external solver and everything works great.

I now wanted to try a solver that only allows for in place updates of the function values. This means I should use the `output_length` option to `jacobian`. When I do that, the analysis performance takes a large hit because from what I believe is the compilation time in generating the function [here](https://github.com/JuliaDiff/ForwardDiff.jl/blob/33f61b3e77afcdcdd20b938e3a6b2def677c6b41/src/api/jacobian.jl#L50) in each new point. 

Since I have a constant `output_length` it should in theory be possible to cache something similar to the `newf` function that  takes the function I want to compute the jacobian of as an extra argument instead of creating the function from scratch.

Would this be possible/sensible?

mlubin:
What's the need for an ``@eval`` on that line?

KristofferC:
Another thing, wouldn't the `output_cache` after the `@eval` be created in global scope and therefore benefit of having a `const` on it?

jrevels:
>Since I have a constant output_length it should in theory be possible to cache something similar to the newf function that takes the function I want to compute the jacobian of as an extra argument instead of creating the function from scratch.

This makes sense, and should definitely be an option in some capacity.

> What's the need for an eval on that line?

It was used so that output_length would be considered a hard-coded literal at the time of compilation. Closing over output_length as a *variable* was causing type instability in the calc_jacobian method, and expanding it as a literal in the expression fixed it (found this out during benchmarking when I originally wrote this). 

I just tried this now on the latest v0.4 release, and it seems like the situation is reversed - type inference now works *without* the eval. 

I'll put in a PR to remove the eval junk 
sometime in the upcoming week (I'm without Internet atm - I'm happy to report that GitHub has nice mobile app).

> Another thing, wouldn't the output_cache after the eval be created in global scope and therefore benefit of having a const on it?

The eval occurs within the scope of the jacobian method, so output_cache is a variable local to the jacobian method.

mlubin:
> The eval occurs within the scope of the jacobian method, so output_cache is a variable local to the jacobian method.

``eval`` is always run at module-level scope

jrevels:
[...]

PR #64, opened 2015-10-12
Remove unnecessary eval within jacobian method

jrevels:
See discussion about the usage of `@eval` in #63

### DIFF
--- a/src/api/jacobian.jl
+++ b/src/api/jacobian.jl
@@ -48,7 +48,7 @@ function jacobian{A}(f, ::Type{A}=Void;
     # appropriate closure
     if output_length > 0
         output_cache = ForwardDiffCache()
-        @inline function newf{N,T,C}(x::Vector{GradientNumber{N,T,C}})
+        function newf{N,T,C}(x::Vector{GradientNumber{N,T,C}})
             output = get_workvec!(output_cache,
                                   GradientNumber, T,
                                   Val{output_length},
--- a/src/api/jacobian.jl
+++ b/src/api/jacobian.jl
@@ -43,21 +43,18 @@ function jacobian{A}(f, ::Type{A}=Void;
                      chunk_size::Int=default_chunk_size,
                      cache::ForwardDiffCache=ForwardDiffCache(),
                      output_length::Int=0)
-    # if output_length > 0, assume that f is of 
-    # the form f!(output, x), and generate the 
+    # if output_length > 0, assume that f is of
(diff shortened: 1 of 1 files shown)


src/api/jacobian.jl
```diff
@@ -43,21 +43,18 @@ function jacobian{A}(f, ::Type{A}=Void;
...
-                output = get_workvec!(output_cache, 
-                                      GradientNumber, T,
-                                      Val{$output_length}, 
-                                      Val{N})
-                $f(output, x)
-                return output
-            end
-            return newf
+        output_cache = ForwardDiffCache()
+        @inline function newf{N,T,C}(x::Vector{GradientNumber{N,T,C}})
```
mlubin: What does ``@inline`` mean for inner functions? Does this do anything?

jrevels: My assumption was that inner functions are not always inlined by default (I don't see why they should be), and that `@inline` would force inlining when `newf` gets called [here](https://github.com/JuliaDiff/ForwardDiff.jl/blob/rmeval/src/api/jacobian.jl#L104) and [here](https://github.com/JuliaDiff/ForwardDiff.jl/blob/rmeval/src/api/jacobian.jl#L149).

The reason I want this inlined is because the whole `newf` thing was simply a mechanism to easily support the `f!(output, x)` style without having to write a bunch of new code (a solution to #63 is probably going to involve some new code anyway, though). From that perspective, `newf` really is a "just-insert-some-code-at-the-call-site" kind of function (or at least it seemed that way when I did the original performance tweaking on this implementation).

mlubin: I don't see how inner functions could ever be inlined though

jrevels: Why not? Just replace function call with the body of the inlined function at the call site. Where that function originated from shouldn't change anything, right (my understanding of the whole process is pretty naive)?

Is this one of those weird situations that's hard to compile for (e.g. "the compiler doesn't want to inline functions passed as arguments")?

mlubin: ``newf`` is passed in at runtime though. The function body isn't known at compile time. Same reason why passing functions isn't well optimized in julia.

jrevels: > newf is passed in at runtime though. The function body isn't known at compile time. Same reason why passing functions isn't well optimized in julia.

I just realized that after rereading my own question above! Okay, I see what you mean now

PR #73, opened 2015-12-03
[RFC] Make ForwardDiffNumbers subtypes of Real

### LINKED ISSUE #66: Define ForwardDiffNumber <: Real
See discussion [here](https://github.com/JuliaStats/Distributions.jl/pull/430/files#r42935897).

Does anybody have any practical arguments for why we *shouldn't* do this?

jrevels:
This PR resolve #66 by defining `ForwardDiffNumber <: Real`, and does some definition bookkeeping to resolve ambiguity warnings and get the tests to pass. 

The most notable change (besides the titular one) is that some methods previously defined between `Real` and `ForwardDiffNumber` are now defined between `ExternalReal` and `ForwardDiffNumber`. `ExternalReal` is defined in ForwardDiff as:

```julia
julia> @eval typealias ExternalReal Union{$(subtypes(Real)...)}
Union{AbstractFloat,Integer,Irrational{sym},Rational{T<:Integer}}
```

This means that (in theory) custom `T<:Real` types should work, but only if they're defined *before* ForwardDiff is used/imported. It might be best to add "ForwardDiff doesn't support custom `T<:Real` types" in the docs to avoid this confusion all together. Thoughts? On a related note, it'd be helpful if someone else wanted to give the docs a once-over to make sure nothing else needs to be changed in the wake of this PR.

Additionally, this PR (once again, in theory) allows one to have `Complex{T<:ForwardDiffNumber}` types, so if anybody wants to play around with those, I'd be interested to see how it turns out. I don't think the API methods (`gradient`, `hessian`, etc.) will work with complex functions, but experiments that work directly with the number types should be feasible.

### DIFF
--- a/docs/source/how_it_works.rst
+++ b/docs/source/how_it_works.rst
@@ -8,7 +8,7 @@ As previously stated, ForwardDiff.jl is an implementation of `forward mode autom
 New Number Types
 ----------------
 
-ForwardDiff.jl provides several new number types, which are all subtypes of the abstract type ``ForwardDiffNumber{N,T,C} <: Number``. These number types store both normal values, and the values of partial derivatives.
+ForwardDiff.jl provides several new number types, which are all subtypes of the abstract type ``ForwardDiffNumber{N,T,C} <: Real``. These number types store both normal values, and the values of partial derivatives.
 
 Elementary numerical functions on these types are overloaded to evaluate both the original function, *and* evaluate partials derivatives of the function, storing the results in a ``ForwardDiffNumber``. We can then pass these number types into a general function :math:`f` (which is assumed to be composed of the overloaded elementary functions), and the derivative information is naturally propogated at each step of the calculation by way of the chain rule.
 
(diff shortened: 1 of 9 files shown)


mlubin:
I think it's fine to exclude custom real types until someone requests it.

jrevels:
cc @scidom, since this PR should help you with your Distributions.jl refactor (once it gets merged and tagged)

scidom:
That's awesome, thanks a lot for doing all this work @jrevels. Yes, I will sort out the Distributions as promised, once this gets merged. Moreover, I am so close now to pushing this month the long-awaited changes to Lora, and I will make sure it can work both with ForwardDiff and ReverseSourceDiff.

scidom:
Shall we tag a new version and push to metadata? :+1: 

Before working on Distributions, I will try to integrate ForwardDiff with the new version of Lora, which should take me at most a week, as I am really curious how this would work :)

PR #81, opened 2015-12-17
Fix #78

### LINKED ISSUE #78: Breaking change between v0.1.2 and v0.1.3
With the v0.1.2 I was able to do

```
x = randn(100)
function h(θ)
           m1 = θ[1] - x
           m2 = θ[2]^2 - (x - θ[1]).^2
           m3 = x.^3 - θ[1]*(θ[1]^2 + 3*θ[2]^2)
           return [m1 m2 m3]
       end

hh(theta) = vec(sum(h(theta), 1))

ForwardDiff.jacobian(hh, [.1, .1])
```

With the latest release (v.0.1.3) I get an error:

```
ERROR: InexactError()
 in setindex! at array.jl:313
 in .- at arraymath.jl:120
 in - at arraymath.jl:142
 in h at none:2
 in hh at none:1
 in _calc_jacobian at /Users/gragusa/.julia/v0.4/ForwardDiff/src/api/jacobian.jl:101
 in jacobian at /Users/gragusa/.julia/v0.4/ForwardDiff/src/api/jacobian.jl:27
```

I compared the release and does not seem to be any change that might have affected this. What am I missing?

gragusa:
This PR propose a simple fix to #78 by overloading  `Base.promote_array_type` to return a `Forward.GradientNumber` instead of `Float64`.

### DIFF
--- a/test/test_jacobians.jl
+++ b/test/test_jacobians.jl
@@ -124,13 +124,40 @@ for fsym in ForwardDiff.auto_defined_unary_funcs
         end
     end
 
-    # Test jacobian of functions involving `.-`, `.+`, etc. #
+
+    
+    # Test overloading of `promote_array_type`
     #-------------------------------------------------------#
+
+
+    promtyp = Base.promote_array_type(Base.DotAddFun(),
+                                  ForwardDiff.GradientNumber{2, Float64,
+                                  Tuple{Float64, Float64}}, Float64)
+    
+    gradnum = ForwardDiff.GradientNumber{2,Float64,Tuple{Float64,Float64}}
+    
+    @test promtyp <: gradnum
+    
+    # jacobian of functions involving `.-`, `.+`, etc.      # 
+    #-------------------------------------------------------#
+
     a = ones(N)
-    ops = [:-, :+]
-    for op in ops
-        @eval fn(x) = ($op)(a, x)
-        D = reduce(&, @eval (($op)(eye(N)))-ForwardDiff.jacobian(fn, a) .== 0)
-        @test D
-    end
+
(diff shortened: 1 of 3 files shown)


test/test_jacobians.jl
```diff
@@ -123,4 +123,14 @@ for fsym in ForwardDiff.auto_defined_unary_funcs
...
     end
-end
\ No newline at end of file
+
+    # Test jacobian of functions involving `.-`, `.+`, etc. #
+    #-------------------------------------------------------#
+    a = ones(N)
+    ops = [:-, :+]
+    for op in ops
+        @eval fn(x) = ($op)(a, x)
```
mlubin: It doesn't seem like this macro-fu saves much space and it's pretty hard to read, could you rephrase these tests more simply?

mlubin:
For some reason coveralls is reporting that the new line added isn't covered. That might be an error because the line is inlined, but just double check that the test is performing as intended.

gragusa:
The updated PR now explicitly test the promote_array_type (coverall likes this) and removed the macro-fu to make code more readable..

mlubin:
Thanks! LGTM
@jrevels?

KristofferC:
Fwiw,  with this I don't have any problem running my codes using ForwardDiff.jl

jrevels:
LGTM. I'd like to extend this fix to general `ForwardDiffNumber`s before we tag a new patch, but that shouldn't stop this PR from being merged (unless you wanted to tackle the general definition yourself).

gragusa:
I am going to close this and open a new PR (I am not able to rebase it).

PR #84, opened 2015-12-21
Support <=, floor, and ceil

timholy:
This supports operations used in various packages I use. I'm submitting it for consideration, but I'd suggest these definitions are worth contemplating a bit before you decide to merge this. From a mathematical perspective, since you can think of a GradientNumber similarly to `r+ϵ`, then you'd expect `r+ϵ < r2` if `r < r2`, but it's not obvious that one would want `r+ϵ <= r`. On the other hand, from the standpoint of, e..g., bounds-checking with a GradientNumber (e.g., in Interpolations), it seems that you probably want `<=` defined. `floor` and `ceil` are a bit similar.

### DIFF
--- a/src/ForwardDiffNumber.jl
+++ b/src/ForwardDiffNumber.jl
@@ -115,6 +115,9 @@ for T in (Base.Irrational, AbstractFloat, Real)
 
         <(x::$T, n::ForwardDiffNumber) = isless(x, n)
         <(n::ForwardDiffNumber, x::$T) = isless(n, x)
+        <=(x::ForwardDiffNumber, n::ForwardDiffNumber) = <=(value(x), value(n))
+        <=(x::$T, n::ForwardDiffNumber) = <=(x, value(n))
+        <=(n::ForwardDiffNumber, x::$T) = <=(value(n), x)
     end
 end
 
@@ -127,6 +130,9 @@ isnan(n::ForwardDiffNumber) = isnan(value(n))
 isfinite(n::ForwardDiffNumber) = isfinite(value(n))
 isinf(n::ForwardDiffNumber) = isinf(value(n))
 isreal(n::ForwardDiffNumber) = isconstant(n)
+
+floor{T}(::Type{T}, n::ForwardDiffNumber) = floor(T, value(n))
+ceil{T}(::Type{T}, n::ForwardDiffNumber) = ceil(T, value(n))
 
 ########################
 # Conversion/Promotion #
--- a/src/ForwardDiffNumber.jl
+++ b/src/ForwardDiffNumber.jl
@@ -131,8 +131,10 @@ isfinite(n::ForwardDiffNumber) = isfinite(value(n))
 isinf(n::ForwardDiffNumber) = isinf(value(n))
 isreal(n::ForwardDiffNumber) = isconstant(n)
 
-floor{T}(::Type{T}, n::ForwardDiffNumber) = floor(T, value(n))
-ceil{T}(::Type{T}, n::ForwardDiffNumber) = ceil(T, value(n))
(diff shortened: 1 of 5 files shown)


timholy:
For reference:
```jl
julia> using DualNumbers

julia> dual(2,1) <= 2
true
```

src/ForwardDiffNumber.jl
```diff
@@ -128,6 +131,9 @@ isfinite(n::ForwardDiffNumber) = isfinite(value(n))
 isinf(n::ForwardDiffNumber) = isinf(value(n))
 isreal(n::ForwardDiffNumber) = isconstant(n)
 
+floor{T}(::Type{T}, n::ForwardDiffNumber) = floor(T, value(n))
```
mlubin: Why should these drop the infinitesimals?

timholy: What do you do with `n = ForwardDiff.GradientNumber(3,(1,-2))`? Is it less than 3 or bigger than 3?

mlubin:
Comparisons always ignore the infinitesimal parts, so your definition of ``<=`` is correct. The reason for this is that for AD purposes, the infinitesimal parts should never affect the control flow of the code.

src/ForwardDiffNumber.jl
timholy: Or are you suggesting I should restrict `T<:Real`?

timholy:
Glad to hear there's a good guiding principle; that same idea was in the back of my thoughts, but it's a nice crisp enunciation.

src/ForwardDiffNumber.jl
mlubin: I'm suggesting it should return a ``ForwardDiffNumber`` with floor applied to the realpart, like the implementation of ``mod`` [here](https://github.com/JuliaDiff/DualNumbers.jl/blob/7d4eedfa1bd6c0af7d844d085d0c2ce6b2357aa7/src/dual.jl#L238).

timholy: Shouldn't it return something that's of type `T`?

Consider:
```jl
xi = floor(Int, x)
fx = x - xi
yx = (1-fx)*y[xi] + fx*y[xi+1]
```

timholy: (I would agree with your suggestion for `floor(typeof(x), x)`.)

mlubin: Maybe in this case it's okay because the derivative of floor is zero almost everywhere anyway.

timholy: Seems worth restricting to Real, however. Pushed a new commit.

src/ForwardDiffNumber.jl
```diff
@@ -128,6 +131,9 @@ isfinite(n::ForwardDiffNumber) = isfinite(value(n))
 isinf(n::ForwardDiffNumber) = isinf(value(n))
 isreal(n::ForwardDiffNumber) = isconstant(n)
 
+floor{T<:Real}(::Type{T}, n::ForwardDiffNumber) = floor(T, value(n))
+ceil{T<:Real}(::Type{T}, n::ForwardDiffNumber) = ceil(T, value(n))
```
jrevels: (accidentally commented in your fork instead of on the PR, so copy and pasted from there...)

Thanks for this! Might want to change the restriction here to T<:ExternalReal if your goal is to disallow T<:ForwardDiffNumber. ForwardDiffNumber was recently subtyped to Real, while ExternalReal is the alias for the union of all the other Real types defined at module load time.

timholy: Fixed. I also added support for `trunc` and `round`, just to be comprehensive.

timholy:
Do let me know whether anything else needs doing here.

mlubin:
LGTM

PR #93, opened 2016-02-01
use parameterised value in more loops

KristofferC:
This uses the parameterized value `xval` which is known to the compiler at compile time instead of the unknown `eachindex(x)`.

Since we already parameterize on this value I don't see a reason not to use it.

For the effect on the LLVM IR, see for example: https://gist.github.com/KristofferC/d264a1d8bdeb9717cbc3 where the number of lines went from 90 -> 60 probably due to loop unrolling.

The IR was generated from:

```jl
@code_llvm ForwardDiff._calc_jacobian((x) -> x, rand(5), Float64, Val{2}, Val{2}, ForwardDiffCache())
```

### DIFF
--- a/src/api/gradient.jl
+++ b/src/api/gradient.jl
@@ -79,7 +79,7 @@ end
         # Vector-Mode
         ResultType = switch_eltype(G, S)
         body = quote
-            @simd for i in eachindex(x)
+            @simd for i in 1:xlen
                 @inbounds gradvec[i] = G(x[i], partials[i])
             end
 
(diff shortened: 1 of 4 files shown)


KristofferC:
Maybe there are more places, I didn't look very carefully.

mlubin:
LGTM

KristofferC:
Is `npartials(G)` always equal to `chunk_size`?

jrevels:
Good idea. This can also be applied to various places in the [Hessian API code](https://github.com/JuliaDiff/ForwardDiff.jl/blob/master/src/api/hessian.jl).

> Is npartials(G) always equal to chunk_size?

Not always, it's pretty confusing - a `chunk_size` of `0` specifies that `Vector` storage should be used, such that `npartials(G) == xlen`. I'd like to clean this up in the refactoring I intend to do this week.

The Travis failure is unrelated to this PR (I'm going to push a fix for the v0.5 closure changes today).

jrevels:
Though, in the loops inside the chunk-mode branch, `N` can be replaced with `chunk_size`.

KristofferC:
I updated with some `N -> chunk_size` changes.

KristofferC:
> EDIT: Looks like you beat me to it

;)

jrevels:
LGTM. Feel free to merge this whenever you're ready.

KristofferC:
Thanks for review

PR #95, opened 2016-02-01
Fix closure-related test breakage

jrevels:
Notice that I didn't remove the `@eval`, as was previously suggested. We test `ForwardDiff` against results generated via symbolic differentiation, which requires the actual `Expr`s, not just the closures (I would use finite differencing, instead, but it's not accurate enough to compare against).

### DIFF
--- a/test/test_behaviors.jl
+++ b/test/test_behaviors.jl
@@ -53,61 +53,57 @@ j = x -> ForwardDiff.jacobian(g, x, chunk_size=2)/2 # jacobian in chunk_mode
 # Promote type Issues #
 #######################
 
-# Test overloading of `promote_array_type`
-#-------------------------------------------------------#
+# Test overloading of `promote_array_type` #
+#------------------------------------------#
 
 promtyp = Base.promote_array_type(Base.DotAddFun(),
                                   ForwardDiff.ForwardDiffNumber{2, Float64,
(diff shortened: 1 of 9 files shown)


test/test_behaviors.jl
```diff
@@ -53,61 +53,58 @@ j = x -> ForwardDiff.jacobian(g, x, chunk_size=2)/2 # jacobian in chunk_mode
...
+#-----------------------------------#
 
-## Test jacobian    
+a = ones(4)
 jac0 = reshape(vcat([[zeros(N*(i-1)); a; zeros(N^2-N*i)] for i = 1:N]...), N^2, N)
 
-for op = [:-, :+, :./, :.*]
-    @eval fn(x) = [($op)(x[1], a); ($op)(x[2], a); ($op)(x[3], a); ($op)(x[4], a)]
-    jac = ForwardDiff.jacobian(fn, a)
+println(N)
```
KristofferC: Should this be here?

jrevels: Oops, meant to clean that up. Good catch.

KristofferC:
Commit without whitespace is empty: https://github.com/KristofferC/ForwardDiff.jl/commit/6ab02005eb25282066b0cbcee853cf33f2367f82?w=1

Issue #92, opened 2016-02-01
Current method of caching arrays seems to be inefficient in some scenarios.


KristofferC:
I was profiling my code a bit and after doing some optimization I noticed that for jacobian computations, ~60-70% of the time is spend in the cache functions of ForwardDiff.

Here is a profile screenshot with some annotations:

![image](https://cloud.githubusercontent.com/assets/1282691/12717714/faedb5cc-c8e9-11e5-91f3-73bb92c50d67.png)

It seems that a lot of the time in the cache method is spent in `ht_keyindex` in `dict.jl` which I believe is a key lookup. [This line](https://github.com/JuliaDiff/ForwardDiff.jl/blob/78ebb41df4eb8642723fde26edee9710702af01b/src/api/cache.jl#L81) and the one after will both look up the key in the dictionary which is unnecessary. I am not sure there is a good way to solve the double lookup right now, https://github.com/JuliaLang/julia/issues/12157 is relevant.

Another could be reconsider the whole cache method so that instead of using a dict some other data structure is used. The dict method is very flexible since it is possible to insert and retrieve arrays of arbitrary type and size but it seems that maybe the lookups are a bit too slow..

If you want to repeat the profile data:

Add these packages:

```jl
Pkg.clone("https://github.com/KristofferC/Voigt.jl")
Pkg.clone("https://github.com/KristofferC/ConstLab.jl")
Pkg.add("Parameters")
Pkg.add("Devectorize")
```

Include this file: https://gist.github.com/KristofferC/71a0797b5b9bf42b8a43

And benchmark with (hopefully I didn't forget anything):

```jl
E = 200000.0
ν = 0.3
σy = 800.0
n = 2.0
l = 1.e-2
Hl = 0.1E
Hg = 4.e7
q = 0.1
D = 1.0
tstar = 1000.0
angles = [20, 40]
const mp = CrystPlastMP(E, ν, σy, n, Hl, Hg, q, D, l, tstar, angles)
const ms = CrystPlastMS(length(angles))

comp_res!(y, x) = compute_residual!(y, x, zeros(6), zeros(2), 0.1, mp, ms)
jac = ForwardDiff.jacobian(comp_res!, output_length = 10)
jac(zeros(10))

@profile for i = 1:10^5 jac(zeros(10)) end
```

KristofferC:
I experimented a bit and with some ugly hacking and slashing in the code I now run with
```jl
type JacobianCache{T, Q}
[...]

PR #99, opened 2016-02-02
WIP/RFC optionally use faster cache

KristofferC:
This is a very early WIP that is more a proof of concept and it would be nice with some comments before I continue.

In #92 I noticed that the dict used to cache arrays is sometimes a bit slow. There are many levels of indirections, an anonymous function -> double key lookup in a dict to get back maybe an array of size 10. The point of this PR would be that a user could optionally declare input size + input type when creating the function that computes the derivative. If this is done, we know the exact data types and sizes that are needed and can therefore use a specialized type to cache the arrays instead of the dict method. This can lead to significantly faster run times.

Right now, what is implemented is an extra method to `gradient` where the input size and input type is used. A new cache `GradientCache` is then used instead of `ForwardDiffCache`.

For a quite simple function of size 10, this gives about a 2x speedup (benchmarked on 0.4.3).
```jl
function f{T}(x::Vector{T})
    s = zero(T)
    @inbounds for i in eachindex(x)
        s += sin(x[i]) + tan(x[i]) + sqrt(x[i])
    end
    return s
end

x = rand(10)
J = zeros(10)

@time for i = 1:10^5 g_PR!(J, x) end
#  0.164378 seconds (400.00 k allocations: 36.621 MB, 5.18% gc time)

@time for i = 1:10^5 g!(J, x) end
#  0.348962 seconds (600.00 k allocations: 41.199 MB, 1.57% gc time)
```

What would be left to do is:

- [ ] Deal with the type instability in the return value of the returned AD function. I believe this is because the keyword argument `cache` is no longer concrete.
- [ ] Reduce the code duplication from creating a very similar `gradient` function
- [ ] Implement similar methods for hessians, tensors and jacobians (have to specify output length as well?),
- [ ] Proper benchmarking
- [ ] Documentation

Note, as I said, this is mostly a proof of concept to get the ball going. If you (@jrevels) had somethins else in mind that you want to go with, feel free to do that.

### DIFF
--- a/src/api/gradient.jl
+++ b/src/api/gradient.jl
@@ -4,14 +4,13 @@
 
 # Gradient Cache #
 #----------------#
-type GradientCache{T, Q, R}
+type GradientCache{T, R}
     workvec::Vector{T}
-    zeros::Partials{R}
-    partials::Vector{Q}
+    zeros::R
+    partials::Vector{R}
 end
 
-function GradientCache{T}(input_size::Int, chunk_size::Int, Tv::Type{T}=Float64)
-    G = GradientNumber{chunk_size, Tv, NTuple{chunk_size, Tv}}
+function GradientCache{T}(input_size::Int, G::Type{T})
     workvec = zeros(G, input_size)
     _zeros = build_zeros(G)
     partials = build_partials(G)
@@ -21,6 +20,8 @@ end
 get_workvec(cache::GradientCache) = cache.workvec
 get_partials(cache::GradientCache) = cache.partials
 get_zeros(cache::GradientCache) = cache.zeros
+
+gradientcache_type(input_size, G) = GradientCache{G, partials_type(G)}
 
 typealias GradForwardDiffCache Union{GradientCache, ForwardDiffCache}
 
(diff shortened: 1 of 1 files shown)


src/api/gradient.jl
```diff
@@ -59,11 +83,51 @@ function gradient{A}(f, ::Type{A}=Void;
     end
 end
 
+function gradient{A, T}(f, input_size::Int, input_type::Type{T},
```
KristofferC: This whole 30 line block here is basically only to use a `@genereated` function to compute the type of the GradientCache (to avoid type instability) and create the cache...

jrevels:
You're right that the cache code has a lot of problems, and desperately needs a refactor. Before you get too far down the line, though, you should be aware that I'm in the middle of a hugely breaking refactor targeting Julia v0.5 that I hope will land either this week or next week. Unfortunately, it will very likely result in the code here becoming obsolete - the API,  `ForwardDiffNumber` types, and cache layer are all going to change dramatically. I'm going to open a WIP PR soon that should explain things in more detail.

That being said, please continue with this PR if you're interested in getting better performance using ForwardDiff in Julia v0.4 - we can update the versions separately.

KristofferC:
If you are in progress of some big changes I will let this rest. I already have a local branch that fixes the cache performance for my use case. 

Would be cool to see a wip PR to play with.

PR #101, opened 2016-02-06
fix domain error in pow

mlubin:
This ports https://github.com/JuliaDiff/DualNumbers.jl/pull/34 to ForwardDiff. It fixes the domain errors in the cases that affect JuMP (all partials are zero), but doesn't fully address the issue, since there will still be issues if some of the partials are zero and some are not.

### DIFF
--- a/src/GradientNumber.jl
+++ b/src/GradientNumber.jl
@@ -110,7 +110,10 @@ end
 #----------------#
 for f in (:^, :(NaNMath.pow))
     @eval begin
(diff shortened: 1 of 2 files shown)


mlubin:
@jrevels, okay to merge?

jrevels:
Sorry, didn't see this PR until now.

mlubin:
@jrevels, thanks! Is it safe to tag a new release? JuMP will be failing its tests until we do so.

jrevels:
Go for it, as long as you don't need things working with Julia v0.5 it should be fine.

Issue #98, opened 2016-02-02
Simple reductions of Vectors of ForwardDiff numbers fail to use SIMD


KristofferC:
A standrad reduction function like:

```jl
function simd_sum{T}(x::Vector{T})
    s = zero(T)
    @simd for i in eachindex(x)
        @inbounds s = s + x[i]
    end
    return s
end
```

which vectorizes with `Vector{Float64}` fails to vectorize with ForwardDiff numbers. For example:

```jl
@code_llvm simd_sum(rand(Float64, 10))

G = ForwardDiff.GradientNumber{4, Float64, NTuple{4, Float64}}
G_vec = G[rand(G) for i = 1:10]
@code_llvm simd_sum(G_vec)
```

It is possible that there is nothing that can be done on the ForwardDiff side to solve this but I thought it could be interesting to track nonetheless.

KristofferC:
Unfortunately, I am not well versed enough in LLVM IR to be able to pinpoint exactly what the problem is...

KristofferC:
This generates very similar code and also fails to vectorize:

```jl


function simd_sum{T}(x::Vector{NTuple{4, T}})
    s = (0.0, 0.0, 0.0, 0.0)
    @inbounds @simd for i in eachindex(x)
        x_i = x[i]
        s = (s[1] + x_i[1], s[2] + x_i[2], s[3] + x_i[3], s[4] + x_i[4])
    end
    return s
end

tuple_vec = [(rand(), rand(), rand(), rand()) for i = 1:20]

@code_llvm simd_sum(tuple_vec)
```

mlubin:
If moving from ``NTuple`` to another storage type for the epsilon components will make it easier to get access to SIMD, I'd say it's worth considering.

KristofferC:
I don't know what other storage type that would be. A vector of GradientNumber is already packed efficiently in contiguous memory so I don't think the storage type is wrong, it is just that it is translated to LLVM IR in a non optimal way.

There are some patches pending to LLVM that might help https://github.com/JuliaLang/julia/pull/5355#issuecomment-176335389

It would be interesting to experiment with https://github.com/eschnett/SIMD.jl when it is mature enough.

jrevels:
[...]

Issue #96, opened 2016-02-02
equality should be overloaded


mlubin:
```
julia> ForwardDiff.GradientNumber(1.0,(5.0,)) == ForwardDiff.GradientNumber(1.0,(6.0,))
false

julia> ForwardDiff.GradientNumber(1.0,(5.0,)) == ForwardDiff.GradientNumber(1.0,(5.0,))
true
```
That should be true/true, since the epsilon components shouldn't affect comparisons.

jrevels:
closed by #102

Issue #94, opened 2016-02-01
Shoule eltype really be defined on ForwardDiffNumbers as it currently is?


KristofferC:
I believe the current definition of `eltype` on the subtypes of `ForwardDiffNumber`s is confusing

Numbers in Julia tend to be viewed similar to 1 length containers containing themselves. We can for example do:

```julia
julia> 1[1]
1
```

Putting it that way, it seems that `eltype(Vector{T}) == eltype(T)`, or (equivalently?) `eltype(T) == typeof(T)` should be an identity for any `T <: Number` type. However, for ForwardDiffNumbers, this is not true, c.f (where `G` is a `GradientNumber`):

```jl
julia> eltype(1)
Int64

julia> eltype([1])
Int64

julia> eltype(G)
Float64

julia> eltype([G])
ForwardDiff.GradientNumber{3,Float64,Tuple{Float64,Float64,Float64}}
```

Due to this, code that has been type inferred fine with Base Julia Number types has gotten type inference problems with `ForwardDiffNumbers`.

I propose changing `eltype` to `eltype{T <: ForwardDiffNumber}(::Type{T}) = T` and use some other internal function to get the `T` in `ForwardDiffNumbers{N, T, C}`.

KristofferC:
Also, tell me if I blab too much in too little time :)

jrevels:
> I propose changing eltype to eltype{T <: ForwardDiffNumber}(::Type{T}) = T and use some other internal function to get the T in ForwardDiffNumbers{N, T, C}.

Sounds like a good idea to me. 

> Also, tell me if I blab too much in too little time :)

On the contrary, this is the perfect time for discussion, since I now have some time to spend on refactoring ForwardDiff.

jrevels:
fixed in #102

Issue #89, opened 2016-01-27
Overloading "black box" function.


KristofferC:
I have a function that is computed in a "complicated enough" way for Automatic Differentiation to not be suitable for it. I can however get the derivative in some other way.

I want to use this function as part of another function where the rest can be AD:ed so I need a way to overload the ForwardDiff call to this function and inject the returned derivative.

I have started a bit but I know I am doing it wrong because I don't know how to propagate correctly my computed derivative to the `Partials` correctly.

```jl
using ForwardDiff

# Function that in reality can't be differentiated with AD
function ff(x)
    @assert length(x) == 3
    y = 2*x
    return y
end

# Derivative to above function
function ff_diff(x)
    @assert length(x) == 3
    return 2*eye(3)
end

# Function that I want to use AD on that uses f
function g(x)
    a = 2*x
    b = ff(a)
    return 5*b
end

typealias GradNumFloat3 ForwardDiff.GradientNumber{3,Float64,Tuple{Float64,Float64,Float64}}
typealias PartialFloat3 ForwardDiff.Partials{Float64,Tuple{Float64,Float64,Float64}}

#Overload f
function ff(x::Vector{GradNumFloat3})
    println("I got called")
    # Extract the values
    x_val = ForwardDiff.get_value(x)
    # Call f and get the derivative
    y = ff(x_val)
    dydx = ff_diff(x_val)
    # Insert back 
    # THIS IS WRONG because it ignores the already stored paritals in x
    nums = [GradNumFloat3(y[i], PartialFloat3((dydx[:, i]...))) for i in 1:length(x)]
    return nums
end

ForwardDiff.jacobian(g, rand(3))

# I got called
# 3x3 Array{Float64,2}:
#  10.0   0.0   0.0
#  0.0  10.0   0.0
#  0.0   0.0  10.0

# Above should be 20.0 on diagonal
```

Any tips on how to do this correctly? Maybe we could put it in the documentation if there is a good way to currently do it.

KristofferC:
From the definition `f(x + ε) = f(x) + εf'(x)` I guess the computed derivative must be inserted with a multiplication of the already computed partials... Componentwise..?

jrevels:
> From f(x + ε) = f(x) + εf'(x) I guess the computed derivative must be inserted with a multiplication of the already computed partials... Componentwise..?

[...]

PR #102, opened 2016-02-08
Giant rewrite for Julia v0.5

### LINKED ISSUE #36: Removing the divisibility restriction on `chunk_size`
Currently, the `chunk_size` option only allows for sizes that divide evenly into the length of the input vector.

As @mlubin pointed out, this isn't always the most reasonable restriction, and could prevent some functions with [big, sad input dimensions](http://www.bigprimes.net/archive/prime/1001/) to be unsupported by the package without access to gobs of memory.

jrevels:
There's been a bunch of discussion between ForwardDiff developers about where this package should go, and with v0.5 on the way, I think now is the right time to start implementing some of  these changes. A lot of these changes are breaking and depend on each other, so it's unfortunately going to be one monolithic PR if we want to avoid intermittent breakage.

After this PR, it will be time to upgrade ForwardDiff's version from`v0.1.x` to `v0.2.x`. We can maintain a `release-0.1` branch for backporting purposes.

This PR is only about 2/3 of the way complete, but aims to fix #36, #88, #92, #94, #96. It paves the way for us to address #83, #89, and #98.

**Major Changes:**

- Drop `Vector` storage for epsilons components. This reduces indirection in `Partials` code that had to handle multiple container types. I have yet to see a case where `Tuple` storage wasn't faster due to GC overhead.

- Consolidate all `ForwardDiffNumber` types into the new type `Dual{N,T}`, which is structured similarly to the former `GradientNumber`. `HessianNumber` and `TensorNumber` have been replaced by the ability to nest `Dual`s. This allows for `Tuple` storage of higher-order partial components, cuts out a lot of code, and should allow for more cache-friendly higher-order API methods (since indexing patterns will be more straightforward). 

- All `@generated` functions have been removed, and the API has been simplified by introducing the `Chunk` immutable and allowing subtypes of `ForwardDiffResults` to be passed as mutable arguments to the API functions. The documentation will explain this in more detail, once I write it.

- Remove the function-generating API functions. I think that it's now easier and more transparent for people to define their own closures, e.g. `j(x) = ForwardDiff.jacobian(f, x)`.

- [The code is now generic enough that higher-order/higher-dimensional derivatives can be written using lower-order API functions. Thus, `tensor`/`tensor!` has been removed, but is still easily implementable by users.](https://github.com/JuliaDiff/ForwardDiff.jl/pull/102#issuecomment-225735459)

- Experimental multithreading support for parallel chunk-mode can be enabled on some API functions by passing in `multithread = true`. I'm getting a 2x speed up vs. the single-threaded implementation when using 4 threads to take the gradient of [`rosenbrock`](https://github.com/JuliaDiff/ForwardDiff.jl/blob/master/benchmarks/testfuncs.jl#L12) with large (> 10,000 elements) input vectors. I haven't benchmarked enough to know how this scales, though I'd expect it to asymptote to N-times speed-up for N threads as the input length goes to infinity. 

- The caching layer is thread-safe for multithreaded API functions, and now features some optimizations proposed by @KristofferC to reduce the number of key lookups per API function call.

**TODO:**

- [x] `jacobian`/`jacobian!`
- [x] `hessian`/`hessian!`
- [x] API Tests
- [x] Do some benchmarking. From my naive benchmarks, this branch's gradient methods perform better than master w.r.t. both time and memory when using Julia v0.4, and *way* better when using Julia v0.5. I'd like to get a similar speed-up with the other API methods, which I've yet to re-implement.
- [x] Update documentation, and provide a guide for upgrading from `v0.1.x` to `v0.2.x`


.travis.yml
```diff
@@ -1,6 +1,5 @@
 language: julia
 julia:
-    - 0.4
```
mlubin: Is this branch dropping support for 0.4? Why?

jrevels: `Threads` doesn't exist in 0.4, so trying to load the package will fail. I guess I could wrap all the multithreading code in conditionals against `VERSION`.

mlubin: Yes please, otherwise JuMP (and anyone who has JuMP installed) will be stuck on ForwardDiff 0.1 for a long time

KristofferC:
Coo stuff! I am impressed you managed to reduce the different `Number` types to one.

Some thoughts about the cache implementation. From previous experience I have found that caching arrays in a Dict is in some cases just too slow. Especially now with the removal of the vector gradient numbers the vectors used will tend to be small so the overhead will be significant.

One possible idea to improve performance of the cache without any impact on the API:

Instead of having the cache hold arrays, we let it hold types that contain as many arrays is needed for the problem (jacobian, gradient etc). The accesses to the dict (`get_workvec!, get_partials!, get_zeros!`) would collapse into one call to the dict, i.e. `get_cache`. This would look in the cache to see if a **type** of the combinations of relevant parameters is stored. If not, a new instance of the needed type is created and returned. This type would contain all the needed arrays `workvec, partials, zeros` and they can simply be accessed as a field for free. 

This requires a bit of extra code to implement the different caches but would significantly reduce the number of typeasserts and calls to the dict needed.

My second comment is that https://github.com/JuliaDiff/ForwardDiff.jl/pull/99 still seems relevant and could with the macros now be implemented in a type stable way. The hardest part would be to figure out a clean API for it.

Regarding SIMD there has been some discussions about it starting here: https://github.com/eschnett/SIMD.jl/issues/1#issuecomment-180997930. With https://github.com/JuliaLang/julia/pull/14976 `SIMD.jl` is able to generate very nice code for `Vec`s but there is still some problems when these `Vec`s are put inside a type. ArchRobinson said that he would look a bit on the generated code for this case. 

I am willing to put in work to implement this but would be good to have some discussions with others to know if this is something more people support.

KristofferC:
I implemented the first idea in: https://github.com/KristofferC/ForwardDiff.jl/commit/1a65099220bb811247b0e86af61a0517b4324351

Benchmarks with:

```jl
function rosenbrock(x)
          a = one(eltype(x))
          b = 100 * a
          result = zero(eltype(x))
          for i in 1:length(x)-1
          end
          return result
      end
```

That commit:
```jl
julia> @time for i = 1:10^5 g(rand(5)) end
  0.166368 seconds (900.00 k allocations: 39.673 MB, 5.06% gc time)

julia> @time for i = 1:10^5 g(rand(20)) end
  0.292666 seconds (900.00 k allocations: 61.035 MB, 2.32% gc time)

julia> @time for i = 1:10^5 g(rand(35)) end
  0.624599 seconds (900.00 k allocations: 94.604 MB, 2.74% gc time)
```

This branch:
```jl
julia> @time for i = 1:10^5 g(rand(5)) end
  0.352903 seconds (800.00 k allocations: 35.095 MB, 1.12% gc time)

julia> @time for i = 1:10^5 g(rand(20)) end
  0.508113 seconds (800.00 k allocations: 56.458 MB, 1.23% gc time)

julia> @time for i = 1:10^5 g(rand(35)) end
  0.986225 seconds (800.00 k allocations: 90.027 MB, 1.27% gc time)
```

jrevels:
I liked your idea in #99, I was actually going to eventually use that idea here. Your approach for separating individual threads' cache data also saves `NTHREADS-1` lookups compared to what I was doing (just storing a separate cache for each thread).

If you want to implement your cache idea as follow-up PR to this one, that would be awesome (otherwise, I'll incorporate your ideas later).

jrevels:
> but there is still some problems when these Vecs are put inside a type. ArchRobinson said that he would look a bit on the generated code for this case.

Makes sense. When I started this refactor, I tried rewriting `Partials` to hold a `Vec` instead of an `NTuple` (It looks like you actually [did the same experiment](https://github.com/JuliaDiff/ForwardDiff.jl/issues/98#issuecomment-180862709)), but didn't see any vectorization happening on Julia v0.5 on my machine. SIMD stuff is tricky.

src/gradient.jl
```diff
@@ -0,0 +1,199 @@
...
+    end
+    return calc_gradient_expr(body)
+end
+
+if VERSION >= THREAD_VERSION
+    @generated function multi_calc_gradient!{S,T,N,L}(f, output::Vector{S}, x::Vector{T}, ::Type{Val{N}}, ::Type{Val{L}})
+        if N == L
+            body = VEC_MODE_EXPR
+        else
+            nthreads = Threads.nthreads()
```
KristofferC: These need to be either `Base.Threads` or a `using Base.Threads` is needed, I think.

KristofferC:
Thank you for your comments. I have a few busy days coming up but when I get time I will try flesh out the cache stuff unless you get to it before me. I thought a bit about benchmarking as well. Since I am likely to do some benchmarking during the cache stuff would it make sense to implement those benchmarks on top of `BenchmarkTrackers`? Similar to how `BaseBenchmarks` is done. What do you think?

As a side note I though I would mention that to me the code is a bit harder to read and reason about now than before. There are a few places like 

```jl
@generated function call_gradient!{S,N,L,M}(f, output::Vector{S}, x::Vector, ::Type{Val{N}}, ::Type{Val{L}}, ::Type{Val{M}})
```

where I am having troubles remembering which length or type `L` and `M` are etc. I'm sure I will grok it in time though :)

KristofferC:
Regarding a global `const cache`. Other packages have gotten problems with it, i.e: https://github.com/JuliaOpt/Convex.jl/issues/83. I guess the problem is that you never know how the user will call the code.

We could either document the `clearcache!` function (should probably have a `sizeof` function for the cache then as well) or we just reset the cache after it reaches a certain threshold.

jrevels:
> As a side note I though I would mention that to me the code is a bit harder to read and reason about now than before.

Yeah, there are a lot of mysterious single-letter identifiers at the moment. I'll change the type parameter names so that they reflect their corresponding keywords, should make it slightly less confusing.

> We could either document the clearcache! function (should probably have a sizeof function for the cache then as well) or we just reset the cache after it reaches a certain threshold.

Transparent documentation and a `sizeof` function are both good ideas, if we end up keeping a global cache as is presented here. 

I'm just going to save any additional cache refactoring for when the rest of this PR is complete, which will probably take some time, given my current schedule. If #105 is ready by that point, I'll just defer to that PR. Either way, you're gathering some valuable information on where the pain points are, and what the right way to solve them might be. Thanks for looking into this!

KristofferC:
#105 is shaping up ok.  What is left is adding the functionality to explicitly pass a created cache as a keyword argument to the macros.

What do you think about removing the option to set the chunk size? It would make some stuff simpler because then the chunk size is simply a function of the input length.

jrevels:
> What do you think about removing the option to set the chunk size?

I've thought about it, but we'd have to be really confident in our chunk-picking heuristic. The right chunk size for any given problem depends not only on the dimension of the input, but on the performance characteristics of the target function itself. If we forced the user to defer to our heuristic with no option to override it, I'd want to make sure that the dependence on the target function is near negligible in the general case.

mlubin:
> What do you think about removing the option to set the chunk size?

Given that we have no good sense of how to choose a good chunk size, I think we should leave this as an option.

KristofferC:
Ok. Makes sense!

KristofferC:
The segfault on 0.5 goes away for me with `--inline=no`. Hmm.

jrevels:
> Edit: It goes away when doing many things

Yeah, it's a finicky one. When I added these tests, I played around for a little bit to try to come up with a minimal repro of the segfault, to no avail. Part of the reason I pushed was to see if anybody else could figure it out :stuck_out_tongue: I would guess this bug is similar to JuliaLang/julia#15043, but I'm not totally sure how it actually behaves yet.

IIRC, I found that `call_gradient` wasn't correctly regenerating for different values of `CHUNK`. Instead, `CHUNK` was propagated as a `TypeVar` even when it should've been bound to an `Int` value. Sometimes this would result in `MethodErrors` (e.g. `no method matching calc_gradient!(::f, ::Array{Float64, 1}, ::Array{Float64, 1}, ::Type{Val{CHUNK}}, ::Type{Val{10}})`) and other times `calc_gradient!` actually got called, but segfaulted. 

Substituting uses of `CHUNK` with `CHUNK + 0` in `call_gradient!` fixed the problem in some cases by causing the resulting value to be correctly inferenced as an `Int`, but obviously that's a hack around the underlying problem.

KristofferC:
Maybe it is worth making an issue upstream?

I played around a bit with trying to reduce it and if I removed some tests I stopped getting segfaults and instead got some GC corruption crash. Would be interesting to go back before `jb/functions` and see if that was what made it fail. Since I have no life, maybe I'll do a bisect...

jrevels:
> I played around a bit with trying to reduce it and if I removed some tests I stopped getting segfaults and instead got some GC corruption crash. Would be interesting to go back before jb/functions and see if that was what made it fail. Since I have no life, maybe I'll do a bisect...

Could be interesting, let me know if you find anything. The workaround I just pushed actually makes the code cleaner, so I don't feel too bad about hacking around the issue in the meantime.

If it helps, I think that the problem is a bug in the way type parameter names are bound. Normally, the argument names of generated functions get bound to their types, while type parameters are bound to their values:

```julia
julia> @generated f(x) = println(x)

julia> f(1) # correct behavior
Int64

julia> @generated g{T}(::Type{Val{T}}) = println(T)

julia> g(Val{1}) # correct behavior
1
```

For some reason, in the old segfaulting code, the type parameters' names were getting erroneously bound to `TypeVar`s, as if the scope of the parameter declaration was leaking into the function body before compilation even occurred. Here's a fake example that demonstrates the behavior (in reality, I couldn't figure out how to reproduce the behavior in a simple case):

```julia
julia> @generated f{T}(::Type{Val{T}}) = println(T)

julia> f(Val{1}) #  I expect this to print `1`
TypeVar
```

In the code I just pushed, there's still weird behavior occurring, but it's more manageable and doesn't cause segfaults. Sometimes, it seems that type parameter names are getting bound to their types, instead of their values, as if they were normal generated function argument bindings instead of type parameter bindings. Here's another fake example for the purpose of demonstration (once again, I couldn't find such a simple reproduction of the bug in reality):

```julia
julia> @generated f{T}(::Type{Val{T}}) = println(T)

julia> f(Val{1}) # I expect this to print `1`
Val{1}
```

The workaround I settled on is [here](https://github.com/JuliaDiff/ForwardDiff.jl/commit/f8433d176d22e41f0b8136eb48112e5240662027#diff-a51666f82528f840f37fcc362fb74659R22); that second definition of `value` only exists to thwart this bug when it arises.

KristofferC:
Nice fixing the segfault!

Just as a fun side note, I'm having a presentation tomorrow for my department about Automatic Differentiation (we do material and computational mechanics) and I will definitely show off how awesome ForwardDiff is :).

KristofferC:
The slides: https://dl.dropboxusercontent.com/u/14340581/kc_on_ad.pdf :smile:

jrevels:
> Just as a fun side note, I'm having a presentation tomorrow for my department about Automatic Differentiation (we do material and computational mechanics) and I will definitely show off how simple and user friendly ForwardDiff is

Awesome! I hope it went well!

jrevels:
So, I did a bunch of work on this over the weekend. Here's what's new:

- got rid of all `@generated` functions, which seemed to fix the random segfaults I was running into
- de-macro'd the API functions, after realizing what a pain it would be to support normal keyword argument features like splatting
- optimized the cache layer as per @KristofferC's suggestions - saw similar speedups to what was reported in #105 
- Hessians are implemented, but my test for them can't pass because Calculus's Hessian code throws an `UndefVar` error (I compare against Calculus in tests). I'll be exploring a fix for this soon.
- wrote more tests

The only thing left now is testing and documentation, besides one last feature: `tensor`/`tensor!`. This function will be a huge pain to write/test/maintain, and probably isn't worth it - forward-mode is probably extremely inefficient in most cases for calculating sparse third-order derivative tensors. **I vote for deprecating `tensor` entirely in `v0.2` (people can still use `v0.1` if they absolutely need it).** I did a GitHub search, and the only package I could find that used `tensor` is [Atoms.jl](https://github.com/libAtoms/Atoms.jl) (cc @gabor1, since you're the most recent committer there).

cortner:
Hi - in Atoms.jl we use `tensor` to compute third derivatives of hamiltonian entries. It would be a real pain to loose that. We only need the full matrix version for it. Will that still be supported? If not, can you point me to alternatives?

jrevels:
Welp, @mlubin and I just realized that ForwardDiff is now implemented generically enough to just implement higher-order **and higher dimensional** derivatives using lower-order methods and `reshape`. I had tried this a while ago, but it didn't seem to work. 

I did some benchmarking and it turns out that writing chunk mode for `hessian` as a composition of `gradient` and `jacobian` is actually much faster than the chunk mode `hessian` implementation I wrote by hand. It also saves a lot of code, so I just pushed the new implementation to this branch. I think the speed-up might be because the new implementation is more cache-friendly than my old implementation.

Now that we've realized this is possible, users can just compose ForwardDiff functions to generate the higher order derivatives that they want. Thus, I'm going to go ahead and say we won't provide `tensor` in ForwardDiff `v0.2`, since making it support the full API would be painful and a lot of code, and I'd rather not add half-implemented functions to the API. 

Instead, users can simply implement `tensor` themselves like:

```julia
function tensor(f, x)
    n = length(x)
    out = ForwardDiff.jacobian(y -> ForwardDiff.hessian(f, y), x)
    return reshape(out, length(x), length(x), length(x))
end
```

I've added this function as a test to make sure that it won't break in the future. 

Now all that's left here is documentation and testing.

KristofferC:
That's really cool.

cortner:
That will be very helpful - thanks. @hjchen1983 : please have a look at this, we will need to do a small rewrite of the perturbation theory when we move to 0.5.

jrevels:
Okay, testing is basically complete. Using johnmyleswhite/Calculus.jl#90, all the tests on this branch pass locally on v0.5. I'll probably need to make a compatibility patch to get them passing on v0.4. 

While testing locally works fine, Travis is throwing errors related to precompilation that I can't reproduce. Anybody know what might be going on there?

jrevels:
Tests now pass locally on v0.4.

jrevels:
Everything here is now done, besides getting tests to pass on Travis - I have no idea what's going on there. 

If anybody is curious about benchmarks, here's the difference between some benchmarks ran on master and this branch on Julia v0.4 (the data is located in the benchmarks folder so you can play with it yourself):

```julia
julia> using JLD, BenchmarkTools

julia> g41 = JLD.load("benchmark/results_julia_v04_forwarddiff_v01.jld", "results");

julia> g42 = JLD.load("benchmark/results_julia_v04_forwarddiff_v02.jld", "results");

julia> compare(a, b, k) = judge(minimum(a[k]), minimum(b[k]))
compare (generic function with 1 method)

julia> compare(g42, g41, "derivative")
6-element BenchmarkTools.BenchmarkGroup:
  tags: []
  "deriv_test_6" => TrialJudgement(-55.51% => improvement)
  "deriv_test_4" => TrialJudgement(-95.39% => improvement)
  "deriv_test_5" => TrialJudgement(-96.40% => improvement)
  "deriv_test_2" => TrialJudgement(-94.97% => improvement)
  "deriv_test_1" => TrialJudgement(-88.74% => improvement)
  "deriv_test_3" => TrialJudgement(-89.70% => improvement)

julia> compare(g42, g41, "gradient")
3-element BenchmarkTools.BenchmarkGroup:
  tags: []
  "rosenbrock" => TrialJudgement(-21.89% => improvement)
  "ackley" => TrialJudgement(+0.11% => invariant)
  "self_weighted_logit" => TrialJudgement(-36.30% => improvement)

julia> compare(g42, g41, "hessian")
3-element BenchmarkTools.BenchmarkGroup:
  tags: []
  "rosenbrock" => TrialJudgement(-81.98% => improvement)
  "ackley" => TrialJudgement(-80.44% => improvement)
  "self_weighted_logit" => TrialJudgement(-78.50% => improvement)

julia> compare(g42, g41, "jacobian")
3-element BenchmarkTools.BenchmarkGroup:
  tags: []
  "trigonometric" => TrialJudgement(-10.58% => improvement)
  "chebyquad" => TrialJudgement(-10.07% => improvement)
  "brown_almost_linear" => TrialJudgement(-9.19% => improvement)
```

KristofferC:
That's what you get for using case insensitive file systems ;) File is called `Partials.j` but you are including `partials.jl`.

jrevels:
lol, thanks! I never would've realized that the local file's case didn't matching the remote file's case:

screen shot 2016-06-15 at 3 46 38 pm

src/ForwardDiff.jl
```diff
@@ -2,90 +2,106 @@ isdefined(Base, :__precompile__) && __precompile__()
...
-                T, C = eltype(P), containtype(P)
-                return ($F){N,T,C}
-            end
-
-            function promote_rule{N,T,C,S<:Real}(::Type{($F){N,T,C}}, ::Type{S})
-                R = promote_type(T, S)
-                return ($F){N,R,switch_eltype(C, R)}
-            end
-        end
+if v"0.4" <= VERSION < v"0.5-"
```
tkelman: please, please, **please** be more specific with VERSION dependent code like this

PR #103, opened 2016-02-09
allow AbstractVector input for gradients

mlubin:
I don't mean to create unnecessary merge conflicts for https://github.com/JuliaDiff/ForwardDiff.jl/pull/102 (so I didn't add tests), but in JuMP we need to pass around subarrays instead of proper Arrays. I don't see why ForwardDiff should insist on the type of the vector inputs so long as it's indexable.

### DIFF
--- a/src/api/gradient.jl
+++ b/src/api/gradient.jl
@@ -4,13 +4,13 @@
 
 # Exposed API methods #
 #---------------------#
-@generated function gradient!{T,A}(output::Vector{T}, f, x::Vector, ::Type{A}=Void;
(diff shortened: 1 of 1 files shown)


mlubin:
Updated to ``::typeof(output)``

jrevels:
If you feel like it, you could loosen the `Vector` restriction on the other API methods as well. Otherwise LGTM, feel free to merge when you're ready.

mlubin:
Thanks! This is all I need for JuMP, and don't want to create any more merge conflicts than necessary.

PR #109, opened 2016-03-08
start, next, done should have type qualifiers

gustafsson:
Partials.jl should not implement a global catch-all `start` for all types.


jrevels:
Oh wow, that's a pretty ridiculous slip-up. Nice catch, thanks!

PR #128, opened 2016-06-25
use in-place get to clean up cache code and potentially avoid an extra key lookup

jrevels:
JuliaCon2016 hackathon has been good to the cache code...


KristofferC:
It's guhna get better

KristofferC:
`do` is pretty slow with the try catch it uses.. maybe it doesn't matter but should probably check

KristofferC:
TODO: Benchmark `ObjectIDDict`

PR #142, opened 2016-08-23
loosen restriction on input array for cache type assertion

jrevels:
This fixes an accidental regression in our support of non-`Array` ForwardDiff inputs ([thanks for pointing this out](https://github.com/JuliaLang/METADATA.jl/pull/6092#issuecomment-241776757), @tkelman).

I'm a tad concerned that this might damage performance for non-`AbstractArray` inputs, since the fallback here uses the result of `ndims` in a type parameter.

It would be nice if there were something like `similar_type` we could call on array-likes, which would allow us to generalize the array types used by the cache types.

### DIFF
--- a/src/cache.jl
+++ b/src/cache.jl
@@ -16,7 +16,8 @@ function JacobianCache{N}(x, chunk::Chunk{N})
     return JacobianCache{N,T,typeof(duals)}(duals, seeds)
 end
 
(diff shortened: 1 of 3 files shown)


tkelman:
thank pkgeval, and keep an eye on the pulse page the day after you tag

PR #150, opened 2016-10-03
Dual specializations for trig-related functions

### LINKED ISSUE #148: Specialized math functions for use with Duals?
I was examining some trig code with Dual values using `@code_llvm` and `@code_native`, and noticed that the generated code included redundant calls to `sin(x)` and `cos(x)`; this derivative could be coded more optimally.

In my case, the function evaluates both `sin` and `cos` of several input variables.  However, I can take advantage of this fact when using Duals, by creating a specialized `sincos` function:

```jl
sincos(x) = (sin(x), cos(x))

function sincos(x::Dual)
    sx = sin(value(x))
    cx = cos(value(x))
    return (
        Dual(sx, cx * partials(x)),
        Dual(cx, -sx * partials(x)),
    )
end
```

I understand ForwardDiff already implements a similar optimization in the case of `exp`, for example.  I'm not really pushing this as a must-have addition (though not opposed if others think it's good).  What I'm really curious to know is if this sort of specialized math function + optimization has other similar use cases and whether it makes sense for other users.

Footnote: I didn't notice a significant runtime improvement from implementing this is my code, so, in my case the idea may be a dud.  (Or I may have too much overhead from other sources.)

dbeach24:
Added Dual specializations for:
  - hypot(x, y)
  - hypot(x, y, z)
  - sincos(x) = (sin(x), cos(x))

Fixes #148

### DIFF
--- a/src/dual.jl
+++ b/src/dual.jl
@@ -318,6 +318,42 @@ end
     return Dual(sqrtv, deriv * partials(n))
 end
 
+@inline Base.hypot{N}(x::Dual{N}, y::Dual{N}) = calc_hypot(x, y)
+@inline Base.hypot(x::Dual, y::Real) = calc_hypot(x, y)
+@inline Base.hypot(x::Real, y::Dual) = calc_hypot(x, y)
+
+@inline function calc_hypot(x, y)
+    vx = value(x)
+    vy = value(y)
+    h = hypot(vx, vy)
+    return Dual(h, (vx/h) * partials(x) + (vy/h) * partials(y))
+end
+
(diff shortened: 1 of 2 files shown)


dbeach24:
Hmmm...  I was testing on Julia 0.5 locally, and that worked fine.
Apparently 0.4.7 doesn't define `hypot(::Int,::Int,::Int)`, or so it would seem.  What's the right thing to do here?  Should I disable those tests for integer cases?

jrevels:
Sorry about being slow to review this, I've been on vacation the past week. This looks good, thanks for contributing!

PR #152, opened 2016-10-13
Add docs and regressions tests for SIMD vectorization of simple dual number operations

### LINKED ISSUE #98: Simple reductions of Vectors of ForwardDiff numbers fail to use SIMD

jrevels:
Closes #98.


test/runtests.jl
```diff
@@ -34,3 +34,10 @@ println("Testing miscellaneous functionality...")
 tic()
 include("MiscTest.jl")
 println("done (took $(toq()) seconds).")
+
+if Base.JLOptions().opt_level >= 3 && VERSION >= v"0.5"
```
jrevels: @KristofferC Do you know what I should add to this conditional to disable these tests for users who won't get SIMD in the LLVM IR?

Also is there any point to me getting this to run on Travis, since Travis just downloads Julia binaries, or are these tests doomed to be local only?

KristofferC: Could try to run the build sysimg.jl script first in .travis.yml file and then start the actual tests?

jrevels: Alright, attempted what you suggested. Let's see if Travis's infrastructure works for this...

KristofferC:
Seems to work? Cool. Maybe the sysimg.jl build should be mentioned for users to get full benefit?

jrevels:
> Seems to work?

Not yet, it's not actually hitting the SIMD tests :( It seems that `Pkg.test` resets `Base.JLOptions().opt_level` back to `2`, which is really annoying (I guess it spawns a subshell to run the tests?). I would just `include("runtests.jl")` manually, but still want to emit coverage information. Can I do that manually by using the `--code-coverage` flag (do I want to use the default `user` setting or the `all` setting)?

> Maybe the sysimg.jl build should be mentioned for users to get full benefit?

That's a good idea.

jrevels:
Guess I should just do [these things](https://github.com/JuliaLang/julia/blob/master/base/pkg/entry.jl#L713), but preserve the optimization level.

KristofferC:
If you want to see what Pkg.test does you can see it here: https://github.com/JuliaLang/julia/blob/3b5c9755b7e2b8f7b4e202ff2d4d3dc07d6decca/base/pkg/entry.jl#L714. Perhaps the -O option from the julia process calling `Pkg.test()` should be forwarded here too?

KristofferC:
To be explicit, it seems that you should use `user` for the coverage flag.

jrevels:
Beat you to it 😛 It'd be nice if `Entry.test!` preserved more of `Base.JLOptions()`...

jrevels:
Well at least now the SIMD tests are running. It seems like [the CPU the job is running on supports SIMD](http://www.cpu-world.com/CPUs/Xeon/Intel-Xeon%20E5-2680%20v2.html), but it's not passing because it doesn't work with inlining off. Maybe we can do two separate runs, then, with the second one being just the SIMD tests and inlining on.

jrevels:
Awesome, SIMD tests now run and pass on Travis!

KristofferC:
Great stuff.

PR #155, opened 2016-11-01
Get rid of the caching layer, and factor out ForwardDiffResults for DiffBase.DiffResult

jrevels:
This PR removes ForwardDiff's caching layer entirely, as well as keyword arguments, in favor of a simpler ~`Options`~ `Config` API which allows users to easily bundle up configuration parameters/work buffers to feed to API methods. This does significantly increase allocations for naively utilized API methods (since it no longer can pull storage from the cache), but enables the user to call all ForwardDiff methods a completely non-allocating manner (as long as the target function is non-allocating). It also gets rid of the overhead due to cache lookup and the maintenance of global state.

Additionally, this PR factors out the `ForwardDiffResult` types and replaces them with support for DiffBase's [DiffResult API](http://www.juliadiff.org/DiffBase.jl/diffresult/).

Besides the new ~`Options`~ `Config` stuff and the switch to using DiffBase, ForwardDiff's API hasn't been altered. I tried to make sure that all breaking changes had proper deprecation warnings, so hopefully this won't outright break people's code.

This PR's doc additions should be sufficient to explain how things work and how to update your code.  If you'd like, you can explore the new docs locally by checking out this branch and opening the HTML in the `docs` folder with your browser.

This PR should not be merged [until DiffBase is registered](https://github.com/JuliaLang/METADATA.jl/pull/6927) (at which point the Travis script here should be updated).

The upcoming [ReverseDiff](https://github.com/JuliaDiff/ReverseDiff.jl) package depends on this PR.


KristofferC:
I can't find any docs on how to allocate work buffers in the `Options` type. Big +1 for making it easier to allocate the cache manually.

KristofferC:
With work buffers perhaps you mean buffers internal to ForwardDiff.jl?

jrevels:
The `Options` constructors are documented [here](https://github.com/JuliaDiff/ForwardDiff.jl/pull/155/files#diff-3e921ddc8c881fae359b7971d2c565b4R87). For example:

```julia
julia> using ForwardDiff

julia> function rosenbrock(x::AbstractVector)
           a = one(eltype(x))
           b = 100 * a
           result = zero(eltype(x))
           for i in 1:length(x)-1
           end
           return result
       end
rosenbrock (generic function with 1 method)

julia> x = rand(1000);

julia> out = similar(x);

# allocates all the work buffers used by ForwardDiff's API methods
julia> opts = ForwardDiff.Options(x);

# first call is warmup
julia> @time ForwardDiff.gradient!(out, rosenbrock, x, opts)
  0.183035 seconds (157.93 k allocations: 6.288 MB, 7.45% gc time)

julia> @time ForwardDiff.gradient!(out, rosenbrock, x, opts);
  0.003557 seconds (4 allocations: 160 bytes)
```

> With work buffers perhaps you mean buffers internal to ForwardDiff.jl?

Yeah, the work buffers (i.e. arrays of `Dual`s) used by the API methods. Was the documentation confusing? I'm open to suggestions if so.

KristofferC:
It was probably me that was stupid :)

jrevels:
> It was probably me that was stupid :)

If a contributor to ForwardDiff had trouble with the docs, I'm sure a user would 😛 I've updated them a bit, hopefully making them more explicit.

jrevels:
I switched the name `Options` to `Config`, which I think makes more since to describe something that might hold work state along with actual "options". It's also singular, which is nicer IMO.

Issue #137, opened 2016-08-09
ForwardDiff v0.2.2 compiles really slowly


rgiordan:
Our [Celeste package](https://github.com/jeff-regier/Celeste.jl) depends extensively on `ForwardDiff.jl` for our [unit tests](https://github.com/jeff-regier/Celeste.jl/blob/master/test/test_derivatives.jl).  When we upgrade to v0.2.2 of `ForwardDiff.jl` we see massive slowdowns in compile time, on the order of 10x longer per function differentiated.  The result is that `ForwardDiff.jl` is too slow to be usable for testing, which for us is a big loss.  Furthermore, we cannot upgrade `Optim.jl` without also upgrading `ForwardDiff.jl` and slamming our test performance.

Note that we're dependent on julia 0.5.   I see that ForwardDiff's own unit tests take about 3x longer in 0.5 than 0.4.

Any thoughts?  Is there any chance that this compile time will come back down soon?

mlubin:
Is this a regression between Julia 0.4 and Julia 0.5 or between ForwardDiff 0.1 and ForwardDiff 0.2 or both?

rgiordan:
Both, I think, though I'm principally reporting a difference between
ForwardDiff 0.1 and 0.2 on Julia 0.5.  We cannot easily run our package on
0.4, so I'm not able to easily test the difference between the two Julia
versions with the same ForwardDiff version.  However, I also see that your
current unit tests run about 3x more slowly on Julia 0.5 than on Julia 0.4.

rgiordan:
@jeff-regier reminded me that on Celeste, the old version of ForwardDiff had about the same performance on Julia 0.4 as 0.5.

jrevels:
I've definitely noticed this as well, especially for Hessians.

jrevels:
Part of it is likely that fact that Julia 0.5's function types give the compiler a lot more info to play with, and I wonder if the rewrite of `NTuple` to `Tuple{Vararg{T,N}}` could have affected compiler performance as well?

KristofferC:
Different functions are different types now so we likely need to compile a lot more now than before when stuff was compiled once for every possible function input. In other words, `Function` used to be concrete, now it is abstract.

tkoolen:
[...]

PR #160, opened 2016-11-29
unrestrict chunk size

jrevels:
A while ago, I refactored ForwardDiff to remove all `@generated` functions from the codebase. A direct effect of this was that a bunch of elementwise functions on `Partials` had to be restricted to a small subset of chunk sizes (from `1` to `10`). Obviously, this restriction can be annoying for users who want to play around with chunk sizes greater than `10`. This PR reintroduces `@generated` functions for the affected `Partials` code, thus enabling the use of any chunk size via the API.

AFAICT after a small bit of profiling, this doesn't affect performance at all. I'll do some more benchmarking later today to confirm that, though.

Note also that I intend to tag ForwardDiff v0.3.0 after this PR gets merged.


mlubin:
How's the hit on compilation time?

KristofferC:
+1. I have found that 10 chunk size is a bit small.

jrevels:
> How's the hit on compilation time?

There's a slight slowdown, but it doesn't seem too significant. Also, I think #137 might be at least somewhat fixed on the `master` branch of ForwardDiff now, so I suppose it's not as much of an issue?

Here's the code I'm using to check this:

```julia

function rosenbrock(x)
   a = one(eltype(x))
   b = 100 * a
   result = zero(eltype(x))
   for i in 1:length(x)-1
   end
   return result
end

out, x = zeros(100, 100), rand(100);

cfg = ForwardDiff.HessianConfig(x);

# measures compilation performance
@time ForwardDiff.hessian!(out, rosenbrock, x, cfg);

# measures runtime performance
@benchmark ForwardDiff.hessian!($out, $rosenbrock, $x, $cfg) evals=1
```

On master (showing the two lines we care about):

```julia
julia> @time ForwardDiff.hessian!(out, rosenbrock, x, cfg);
  2.081374 seconds (2.03 M allocations: 45.544 MB, 1.04% gc time)

julia> @benchmark ForwardDiff.hessian!($out, $rosenbrock, $x, $cfg) evals=1
BenchmarkTools.Trial:
  memory estimate:  87.52 kb
  allocs estimate:  11
  --------------
  minimum time:     5.263 ms (0.00% GC)
  median time:      5.271 ms (0.00% GC)
  mean time:        5.315 ms (0.13% GC)
  maximum time:     8.135 ms (33.59% GC)
  --------------
  samples:          941
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%
```

On this branch:

```julia
julia> @time ForwardDiff.hessian!(out, rosenbrock, x, cfg);
  2.188665 seconds (2.12 M allocations: 48.981 MB, 0.84% gc time)

julia> @benchmark ForwardDiff.hessian!($out, $rosenbrock, $x, $cfg) evals=1
BenchmarkTools.Trial:
  memory estimate:  87.52 kb
  allocs estimate:  11
  --------------
  minimum time:     5.265 ms (0.00% GC)
  median time:      5.274 ms (0.00% GC)
  mean time:        5.321 ms (0.16% GC)
  maximum time:     9.114 ms (0.00% GC)
  --------------
  samples:          940
  evals/sample:     1
  time tolerance:   5.00%
  memory tolerance: 1.00%
```

Using ForwardDiff v0.2.2:

```julia
julia> c = ForwardDiff.Chunk{10}();

julia> @time ForwardDiff.hessian!(out, rosenbrock, x, c);
 17.425611 seconds (2.38 M allocations: 60.944 MB, 0.21% gc time)
```

So it seems compilation times have improved significantly since #137 was filed.

KristofferC:
Should these checks have been removed?

```jl
julia> ForwardDiff.GradientConfig{15}(rand(15))
------ AssertionError ------------------ Stacktrace (most recent call last)

AssertionError: N <= CHUNK_THRESHOLD
```

PR #162, opened 2016-12-02
Fix incorrect Partials supertype relation and Config chunk size assertion

### LINKED ISSUE #161: Order of parameters for Partials.
The order of the parameters for the `Partial` type are `{N,T}` while for `AbstractArray` they are `{T,N}`. This leads to some confusion, for example, trying to print a partial type now gives:

```jl
julia> ForwardDiff.Dual(1, (2,3)).partials
2-element ForwardDiff.Partials{2,Int64}:
Error showing value of type ForwardDiff.Partials{2,Int64}:
------ MethodError --------------------- Stacktrace (most recent call last)

....
 [6] — print_response(::Base.REPL.LineEditREPL, ::Any, ::Void, ::Bool, ::Bool) at REPL.jl:139
 [7] — print_response(::Base.Terminals.TTYTerminal, ::Any, ::Void, ::Bool, ::Bool, ::Void) at REPL.jl:154
 [8] — display(::ForwardDiff.Partials{2,Int64}) at multimedia.jl:143
 [9] — display(::Base.REPL.REPLDisplay{Base.REPL.LineEditREPL}, ::ForwardDiff.Partials{2,Int64}) at REPL.jl:135
 [10] — display(::Base.REPL.REPLDisplay{Base.REPL.LineEditREPL}, ::MIME{Symbol("text/plain")}, ::ForwardDiff.Partials{2,Int64}) at REPL.jl:132
 [11] — #showarray#330(::Bool, ::Function, ::IOContext{Base.Terminals.TTYTerminal}, ::ForwardDiff.Partials{2,Int64}, ::Bool) at show.jl:1617

MethodError: no method matching isless(::Int64, ::Type{Int64})
Closest candidates are:
  isless(::Real, ::AbstractFloat) at operators.jl:41
  isless(::Real, ::ForwardDiff.Dual{N,T<:Real}) at /home/kristoffer/.julia/v0.5/ForwardDiff/src/dual.jl:130
  isless(::Real, ::Real) at operators.jl:75
  ...
```

This is because of `ndims{T,N}(::AbstractArray{T,N}) = N` in Base picking the wrong parameter. Maybe the parameters should be swapped? Implementing this function specifically for Partials, errors out in other places.

### LINKED ISSUE #163: extract_gradient! does not work for outputting to an Array
While this is an internal function, I think this behavior is weird. It should just extract the partials into the passed vector I presume?

```jl
julia> ForwardDiff.extract_gradient!(zeros(3), ForwardDiff.Dual(1.0, (2.0,3.0,4.0)))
------ MethodError --------------------- Stacktrace (most recent call last)

 [1] — extract_gradient!(::Array{Float64,1}, ::ForwardDiff.Dual{3,Float64}) at gradient.jl:40
 [2] — copy!(::Array{Float64,1}, ::ForwardDiff.Partials{3,Float64}) at multidimensional.jl:614

MethodError: no method matching getindex(::Tuple{Float64,Float64,Float64}, ::CartesianIndex{1})
```

jrevels:
fixes #161
fixes #163

### DIFF
--- a/src/config.jl
+++ b/src/config.jl
@@ -12,10 +12,7 @@ for Config in (:GradientConfig, :JacobianConfig)
             seeds::NTuple{N,Partials{N,T}}
             duals::D
             # disable default outer constructor
-            function $Config(seeds, duals)
-                @assert N <= CHUNK_THRESHOLD
-                new(seeds, duals)
-            end
+            $Config(seeds, duals) = new(seeds, duals)
         end
 
         # This is type-unstable, which is why our docs advise users to manually enter a chunk size
--- a/src/partials.jl
+++ b/src/partials.jl
@@ -1,4 +1,4 @@
-immutable Partials{N,T} <: AbstractArray{N,T}
+immutable Partials{N,T} <: AbstractVector{T}
     values::NTuple{N,T}
 end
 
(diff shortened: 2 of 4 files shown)


KristofferC:
Thanks.

PR #164, opened 2016-12-02
make deprecation messages more accurate

KristofferC:
Is this clearer?


jrevels:
Thanks for catching this!

PR #166, opened 2016-12-04
enable one-sided nested differentiation for binary Dual functions

jrevels:
This makes it possible to do mixed-mode nested AD using ReverseDiff's pre-compiled tapes, which can be faster and more memory efficient than using `ForwardDiff.hessian` or `ReverseDiff.hessian`.

This change intends to be equivalent to adding the following definitions to binary Dual functions:

```julia
f{N,M,A<:Real,B<:Dual}(a::Dual{N,A}, b::Dual{M,B}) = invoke(f, (Real, Dual), a, b)
f{N,M,A<:Dual,B<:Real}(a::Dual{N,A}, b::Dual{M,B}) = invoke(f, (Dual, Real), a, b)
```

...but instead of the above, it does a bit of gymnastics to avoid `invoke`.

### DIFF
--- a/src/dual.jl
+++ b/src/dual.jl
@@ -1,3 +1,5 @@
+const ExternalReal = Union{subtypes(Real)...}
+
 ########
 # Dual #
 ########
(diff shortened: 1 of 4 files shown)


jrevels:
Man, these SIMD tests are actually really useful for catching regressions.

src/dual.jl
```diff
@@ -211,45 +231,41 @@ end
...
-                expv = ($f)(v1, v2)
-                powval = v2 * ($f)(v1, v2 - 1)
-                logval = expv * log(v1)
-                new_partials = _mul_partials(partials(n1), partials(n2), powval, logval)
-                return Dual(expv, new_partials)
-            end
+            v1, v2 = value(n1), value(n2)
+            expv = ($f)(v1, v2)
+            powval = v2 * ($f)(v1, v2 - 1)
+            logval = isconstant(n2) ? one(expv) : expv * log(v1)
```
jrevels: This `isconstant(n2)` branch is necessary for the following to work:

```
julia> a, b = Dual(-2.0,1.0), Dual(2.0,0.0)
(Dual(-2.0,1.0),Dual(2.0,0.0))

julia> a ^ b
Dual(4.0,-4.0)
```

This case usually turns up when upstream type promotion has converted some scalar `x::Real` to `Dual{N,typeof(x)}` (with all zero partial components).

My slight refactor here is because this PR somehow causes the compiler to give up on inferring the original branch for nested `Dual`s (AFAICT it's fully inferrable, Julia is just being wonky). Unfortunately,  in its refactored position, this branch now partially prevents SIMD vectorization of this method with nested `Dual` arguments, thereby causing the current test failure.

Does anybody know a good workaround here? If worst comes to worst, I'll accept the minor SIMD regression to enable this PR's functionality, but I'm hoping I can have my cake and eat it too.

@mlubin @KristofferC  curious if you guys have any insights

PR #170, opened 2016-12-07
don't accidentally widen input type in sqrt calculation

### LINKED ISSUE #169: Unexpected promotion of dual numbers with sqrt
```jl
julia> a = ForwardDiff.Dual(Float32(1), Float32(2))
Dual(1.0,2.0)

julia> typeof(a)
ForwardDiff.Dual{1,Float32}

julia> typeof(sqrt(a))
ForwardDiff.Dual{1,Float64}
```

The issue is the multiplication with 0.5 here: https://github.com/JuliaDiff/ForwardDiff.jl/blob/b467b89cbc2119d449a1bf14c106d94280f2f548/src/dual.jl#L342

Could either change it to `  / 2` but that might be a performance hit. Could also promote the `0.5` to the number type of the input argument but that might not be possible for some input types. Could also just convert the result to the same number type as the valtype of the input dual number.

jrevels:
fixes #169 

This doesn't seem to introduce any significant performance difference. On master (`d` is a `Dual{3,Float64}`):

```julia
julia> @benchmark sqrt($d)
BenchmarkTools.Trial:
  memory estimate:  0.00 bytes
  allocs estimate:  0
  --------------
  minimum time:     9.078 ns (0.00% GC)
  median time:      9.113 ns (0.00% GC)
  mean time:        9.119 ns (0.00% GC)
  maximum time:     17.372 ns (0.00% GC)
```

with this PR:

```julia
julia> @benchmark sqrt($d)
BenchmarkTools.Trial:
  memory estimate:  0.00 bytes
  allocs estimate:  0
  --------------
  minimum time:     9.087 ns (0.00% GC)
  median time:      9.098 ns (0.00% GC)
  mean time:        9.165 ns (0.00% GC)
  maximum time:     62.783 ns (0.00% GC)
```

### DIFF
--- a/src/dual.jl
+++ b/src/dual.jl
@@ -339,7 +339,7 @@ end
 
 @inline function Base.sqrt{N}(n::Dual{N})
     sqrtv = sqrt(value(n))
-    deriv = 0.5 / sqrtv
+    deriv = inv(sqrtv + sqrtv)
     return Dual(sqrtv, deriv * partials(n))
 end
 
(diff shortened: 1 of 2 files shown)


KristofferC:
Clever solution :)

oscardssmith:
Not sure if this is any different, but it seems like `1/(2*sqrtv)` might be slightly faster while still non widening

Issue #176, opened 2016-12-14
need to define rtoldefault for Dual


tpapp:
Eg
```
import Base: rtoldefault

rtoldefault{N, T <: Real}(::Type{ForwardDiff.Dual{N,T}}) = rtoldefault(T)
```
does the trick. Otherwise it falls back to `0`, and `isapprox` fails.

tpapp:
PR #177.

jrevels:
fixed by #177

PR #177, opened 2016-12-14
Added a method for Base.rtoldefault, fixing #176.

tpapp:
Small fix for #176, with tests. Assumes that only `value` is compared by predicated (looks that way from the code), hope this is correct.

### DIFF
--- a/src/dual.jl
+++ b/src/dual.jl
@@ -173,6 +173,8 @@ Base.isinteger(n::Dual) = isinteger(value(n))
 Base.iseven(n::Dual) = iseven(value(n))
 Base.isodd(n::Dual) = isodd(value(n))
 
+Base.rtoldefault{N, T <: Real}(::Type{Dual{N,T}}) = Base.rtoldefault(T)
+
 ########################
 # Promotion/Conversion #
 ########################
--- a/src/dual.jl
+++ b/src/dual.jl
@@ -100,6 +100,8 @@ Base.copy(n::Dual) = n
 Base.eps(n::Dual) = eps(value(n))
 Base.eps{D<:Dual}(::Type{D}) = eps(valtype(D))
 
+Base.rtoldefault{N, T <: Real}(::Type{Dual{N,T}}) = Base.rtoldefault(T)
+
 Base.floor{T<:Real}(::Type{T}, n::Dual) = floor(T, value(n))
 Base.floor(n::Dual) = floor(value(n))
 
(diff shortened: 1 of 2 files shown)


test/DualTest.jl
```diff
@@ -434,4 +434,10 @@ for N in (0,3), M in (0,4), T in (Int, Float32)
     end
 end
 
+let p = Partials((0.0,0.0)),
+    d = Dual(1.0, p)
+    @test Base.rtoldefault(typeof(d)) ≡ Base.rtoldefault(typeof(value(d)))
+    @test Dual(1.0-eps(Float64), p) ≈ d
+end
```
jrevels: Instead of testing this here, can you place it [here](https://github.com/tpapp/ForwardDiff.jl/blob/f90357d9d6a30789930be0832c109153b86d6d83/test/DualTest.jl#L135) instead, and test it using `FDNUM` and `NESTED_FDNUM` instances instead of making a new `Dual`? 

That way, it will be tested against different number types, number of partials, nested duals, etc. AFAICT, your implementation currently works in all those cases, so the tests should catch any regressions with those cases if somebody changes the code in the future.

src/dual.jl
```diff
@@ -173,6 +173,8 @@ Base.isinteger(n::Dual) = isinteger(value(n))
 Base.iseven(n::Dual) = iseven(value(n))
 Base.isodd(n::Dual) = isodd(value(n))
 
+Base.rtoldefault{N, T <: Real}(::Type{Dual{N,T}}) = Base.rtoldefault(T)
```
jrevels: Maybe move this to [here](https://github.com/tpapp/ForwardDiff.jl/blob/f90357d9d6a30789930be0832c109153b86d6d83/src/dual.jl#L102) (so that it's not under the "predicates" section, since it's not a predicate).

tpapp:
Made the changes you requested, thanks for your help.

test/DualTest.jl
```diff
@@ -131,6 +131,9 @@ for N in (0,3), M in (0,4), T in (Int, Float32)
         @test round(FDNUM) === round(PRIMAL)
         @test round(FDNUM2) === round(PRIMAL2)
         @test round(NESTED_FDNUM) === round(PRIMAL)
+
+        @test Base.rtoldefault(typeof(FDNUM)) ≡ Base.rtoldefault(typeof(PRIMAL))
```
jrevels: Can you also add:

```julia
@test Base.rtoldefault(typeof(NESTED_FDNUM)) ≡ Base.rtoldefault(typeof(PRIMAL))
```

Otherwise, looks good!

tpapp: Added it (sorry for the delay, I was on vacation).

jrevels:
Looks good - thanks again!

PR #179, opened 2016-12-15
[RFC/WIP] don't propagate NaN/Inf if Partials are zero

### LINKED ISSUE #178: log(Dual(0.0,0.0)) is Dual(-Inf,NaN)
I have been tracking an issue of `NaN` propagation in my code, and narrowed it down to the following MWE:
```julia
julia>  log(Dual(0.0,0.0))
Dual(-Inf,NaN)
```
Would `Dual(-Inf,0.0)` make sense? After all, the value is not sensitive to that variable.

The context is a Hidden Markov Chain problem for which I am fitting a model using derivative-based Bayesian methods. It makes sense to keep track of the log likelihood because of underflow/overflow.

A **quick workaround** (that would overwrite the method) would be appreciated so that I could proceed with work (I haven't had time to dig into `ForwardDiff` internals yet), but given some advice on what to do I would also submit a PR.

jrevels:
This aims to fix #178, and any related issues with other functions. 

I am not totally sure this does the right thing for nested Dual numbers; perhaps the perturbation check should be recursive?

There is a slight performance penalty here which is more or less significant depending on the function. I'm also interested to see what it does to SIMD performance.

EDIT:Removed a snippet of an alternate idea that I've now decided against.


jrevels:
The approach I ended up going with here is to make `Partials` "`NaN`/`Inf`-safe". 

Unfortunately, I'm seeing about performance regressions of ~10% on this branch vs. master when benchmarking some DiffBase functions, and that's probably going to be near the lower bound on such regressions for this change - I'm only benchmarking cases where it only needs to do the `isnan`/`isinf`/zero checks, i.e. not even checking the whole `Partials`.

Maybe it would be better to create a module-level flag for enabling this change so that users can opt-in only if they need it, preserving performance for users who don't need it. cc @mlubin @KristofferC for your opinions.

src/partials.jl
```diff
@@ -75,18 +75,29 @@ Base.convert{N,T}(::Type{Partials{N,T}}, partials::Partials{N,T}) = partials
...
 @inline @compat(Base.:*)(x::Real, partials::Partials) = partials*x
-@inline @compat(Base.:/)(partials::Partials, x::Real) = Partials(div_tuple_by_scalar(partials.values, x))
 
-@inline function _mul_partials{N}(a::Partials{N}, b::Partials{N}, afactor, bfactor)
-    return Partials(mul_tuples(a.values, b.values, afactor, bfactor))
+# NaN/Inf-safe methods #
+#----------------------#
+
+@inline function @compat(Base.:*)(partials::Partials, x::Real)
+    x = ifelse((isnan(x) || isinf(x)) && iszero(partials), one(x), x)
```
mlubin: ``!isfinite`` could be better than ``isnan || isinf``

KristofferC:
10% seems quite a hefty penalty on my opinion. 

One way could be to define most operations for an AbstractDual and then have two concrete types where the NaN behaviour is dispatched on. I don't really like this way though. Feels a bit too intrusive.

jrevels:
> I don't really like this way though.

"This way" referring to this PR (and the proposed "users switch this on/off via a module-level constant") or "this way" referring to your `NaN`-safe type idea? 

I do think the performance hit is too great here to make this the default behavior. The most convenient API for this, IMO, would be adding something like a `nansafe` flag to the config types so that users could easily enable this behavior when they need it. Having this be a runtime switchable thing would probably mean implementing something close to what you mentioned (though I might just go with a Bool type parameter to `Partials` to flip it on/off). That could be done in a separate PR, though, and we could just provide the module-level option for now.

jrevels:
> Any performance benefit?

From switching to `!isfinite(x)`? Yeah, thanks! It seems to improve performance by around ~3%-5% on the DiffBase functions I'm looking at versus `isnan(x) || isinf(x)`.

mlubin:
The module-level option is problematic for packages that use ForwardDiff under the hood. What is JuMP supposed to do?

KristofferC:
I thought that introducing a second concrete Dual number type is perhaps a quite large change to solve this problem (at least compared to the solution you have now). A 5% performance penalty might be acceptable to skip the headache of two concrete types? I haven't really run into this problem myself so I feel I can't really weigh options against eachother in an informed way.

jrevels:
> The module-level option is problematic for packages that use ForwardDiff under the hood. What is JuMP supposed to do?

Really good point. It's definitely bothersome that whatever module-level option the user sets would apply to all downstream module usage, when they might instead want to only apply it selectively (e.g. have `NaN`-safe mode enabled for their JuMP model but disabled for usage in Optim). For now, however, it'd still an improvement over not having this option at all.

In a future PR, we could add `AbstractConfig` constructors like `GradientConfig(x, Chunk{N}, NaNSafe{true})` (OT: I really wish something like https://github.com/JuliaLang/julia/pull/16580 would land in v0.6, but that's really unlikely...fingers crossed for v1.0). The generated dual numbers would then contain partials of type `Partials{N,T,true}`, which would dispatch to the `NaN`-safe versions of methods. Downstream modules could then support passing that option through their APIs (or simply setting it to a reasonable default for themselves internally).

.travis.yml
```diff
@@ -10,10 +10,10 @@ script:
     - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
     - if (julia -e 'VERSION < v"0.5" && exit(1)'); then
         julia -e 'include(joinpath(JULIA_HOME, Base.DATAROOTDIR, "julia", "build_sysimg.jl")); build_sysimg(force=true)';
-        julia -e 'Pkg.clone(pwd()); Pkg.build("ForwardDiff"); Pkg.test("ForwardDiff"; coverage=true)';
+        julia -e 'Pkg.clone(pwd()); Pkg.build("ForwardDiff"); Pkg.checkout("DiffBase"); Pkg.test("ForwardDiff"; coverage=true)';
```
tkelman: Do you still need to do this? These kind of things should be very temporary, otherwise you're not testing against the same versions of things your users will have installed.

jrevels: DiffBase serves as a repository of test functions. If I didn't do this, I'd have to re-tag DiffBase every time I wanted to add a new shared test function, which would get really annoying.

I share your concern though. I've been meaning to factor the shared test suite out of DiffBase and into an unregistered package (DiffTestSuite or something). Then DiffTestSuite could be a test-only dependency that I `Pkg.checkout` every time, and I could go back to CI-testing DiffBase's functionality with the proper versions. That's what I probably should've done this in the first place...I'll set it up tomorrow.

tkelman: `Pkg.add("ForwardDiff"); Pkg.test("ForwardDiff")` should pass. If that requires tagging more often, tag more often. Pkg.test shouldn't depend on unregistered packages.

tkelman: Same dependency versioning issue applies - if it's needed to pass tests and doesn't exist in every version of DiffBase, then at least `test/REQUIRE` needs to have a corresponding minimum version requirement.

jrevels: Fair enough - I'll fix this to use tagged DiffBase versions and start tagging DiffBase more often, then.

tkelman: I don't mind if you want to run additional tests on CI-only, or have things temporarily checked out to branches on CI if that's necessary while things are being developed -- as long as you keep in mind how that configuration will differ from what users will have installed, and that Pkg.test should work for them on releases too.

jrevels: Yeah, no worries! My future workflow will be to keep master branch's CI running against tagged dependency versions for the reasons you mentioned. I can always toggle a PR's CI config to use whatever DiffBase branch I want during development, and then before merging, I can tag the new DiffBase version (if necessary) and switch the config back to using only tagged versions.

andreasnoack:
This is what I expect to see after flipping the `NANSAFE_MODE_ENABLED` flag
```julia
julia> Inf*ForwardDiff.Dual{Float64}(1.0,0.0)
Dual{Float64}(Inf,0.0)
```
but should I expect to see
```julia
julia> Inf*ForwardDiff.Dual{Float64}(1.0,0.0,1.0)
Dual{Float64}(Inf,NaN,Inf)
```
?

jrevels:
Hmm. Yeah, it would probably be consistent for us to check/propagate all partials individually when `NANSAFE_MODE_ENABLED` is set to `true`. Might cause a bigger performance hit for this `NANSAFE` mode but maybe not, and it'd probably be worthwhile either way.

PR #189, opened 2017-02-10
Fix most depwarn on 0.6

yuyichao:
There are 3 `@test_approx_eq_eps` that are not fixed. Naively replacing them causes the tests to fail. Not sure which corner case they are hitting yet.


jrevels:
Thanks!

> Naively replacing them causes the tests to fail. Not sure which corner case they are hitting yet.

According to @andreasnoack, `@test_approx_eq_eps x y eps` is actually performing a different (less strict by a constant?) comparison than `@test isapprox(x, y, atol=eps)`. I'll open a different PR to fix the tolerance for those cases.

PR #191, opened 2017-02-11
Fix 0.6 abstract type declaration depwarn

yuyichao:
Local tests pass on 0.4 and 0.5. I'm assuming @ararslan [will have `a branch that should fix it` on 0.6](https://github.com/JuliaStats/StatsBase.jl/pull/235#issuecomment-279183388).


yuyichao:
Ping. I think this can be merged without #192?

PR #193, opened 2017-02-12
add prettier printing for duals

KristofferC:
With inspiration taking from the complex pretty printing in base:

```julia
julia> ForwardDiff.Dual(3, 1, 2)
3 + 1ɛ₁ + 2ɛ₂

julia> ForwardDiff.Dual(3, 1)
3 + 1ɛ

julia> ForwardDiff.Dual(3, -1)
3 - 1ɛ

julia> ForwardDiff.Dual(3, 1, 2, 3, 4, 5, 6, 7, 8, 9, 1, 2)
3 + 1ɛ₁ + 2ɛ₂ + 3ɛ₃ + 4ɛ₄ + 5ɛ₅ + 6ɛ₆ + 7ɛ₇ + 8ɛ₈ + 9ɛ₉ + 1ɛ₁₀ + 2ɛ₁₁
```

Will add tests after obligatory bikeshedding :)

### DIFF
--- a/src/dual.jl
+++ b/src/dual.jl
@@ -403,10 +403,28 @@ end
 # Pretty Printing #
 ###################
 
+const _subscripts = ["₀", "₁", "₂", "₃", "₄", "₅", "₆", "₇", "₈", "₉"]
+
 function Base.show{N}(io::IO, n::Dual{N})
-    print(io, "Dual(", value(n))
-    for i in 1:N
-        print(io, ",", partials(n, i))
+    parts, val = partials(n), value(n)
+    compact = get(io, :compact, false)
+    show(io, val)
+    for (i, p) in enumerate(parts)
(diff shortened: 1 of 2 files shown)


KristofferC:
Perhaps better to implement `show` for the partials instead.

jrevels:
I actually had something similar to this at one point, except it would parenthesize epsilon components for nested dual numbers. I got rid of it in favor of the existing behavior, which I now find much more legible for actual work. I really don't want to go back to drowning in epsilon soup...

I appreciate your work here, so sorry to be a downer, but it's probably going to get a 👎 from me unless there are clearer examples of where this pretty printing is more legible than the current pretty printing.

KristofferC:
Sounds reasonable. I'm just procrastinating here anyway ;)

PR #199, opened 2017-02-23
deprecation fix for indexing traits

jrevels:
this is #198, but rebased against master


jrevels:
Travis v0.6 failure should be eventually fixed by #200

PR #200, opened 2017-02-23
use SpecialFunctions package due to function deprecations in Julia v0.6

jrevels:
Supercedes #192 - we should actually overload SpecialFunctions functions instead of Base functions on v0.6.

For example, on v0.5, this PR will overload both `Base.erf` and `SpecialFuntions.erf` with `Dual` definitions. If `VERSION >= v"0.6.0-dev.2767"`, only `SpecialFuntions.erf` is overloaded with a `Dual` definition.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -3,3 +3,4 @@ DiffBase 0.0.3
 Compat 0.17.0
 Calculus 0.2.0
 NaNMath 0.2.2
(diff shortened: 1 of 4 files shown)


jrevels:
Travis v0.6 is failing due to a SIMD regression, which is unrelated to this PR

Issue #202, opened 2017-03-08
fma not defined


tpapp:
MWE:
```julia
julia>  using ForwardDiff

julia>  f(x) = fma(x, 3.0, 4.0)
WARNING: Method definition f(Any) in module Main at REPL[2]:1 overwritten at REPL[6]:1.
f (generic function with 1 method)

julia>  ForwardDiff.derivative(f, 1.0)
ERROR: fma not defined for ForwardDiff.Dual{1,Float64}
 in derivative(::#f, ::Float64) at /home/tamas/.julia/v0.5/ForwardDiff/src/derivative.jl:5
```
A PR will follow soon — I guess I just [add this manually](http://www.juliadiff.org/ForwardDiff.jl/contributing.html#manually-adding-functions-to-forwarddiff)?

tpapp:
Now that the PR is merged, I am closing the issue.

I wonder if you could do a minor release in the near future, so that I could `REQUIRE` a version that contains the fix.

PR #203, opened 2017-03-08
Added fma, with tests.

tpapp:
Tests required a third value set (PRIMAL etc) in the testing
framework, which was also added.

Fixes issue #202.

### DIFF
--- a/src/dual.jl
+++ b/src/dual.jl
@@ -410,6 +410,38 @@ end
 @ambiguous @inline Base.atan2(y::Real, x::Dual) = calc_atan2(y, x)
 @ambiguous @inline Base.atan2(y::Dual, x::Real) = calc_atan2(y, x)
 
+@inline function Base.fma(x::Dual, y::Dual, z::Dual)
+    vx, vy = value(x), value(y)
+    result = fma(vx, vy, value(z))
+    return Dual(result,
+                _mul_partials(partials(x), partials(y), vx, vy) + partials(z))
+end
+
+@inline function Base.fma(x::Dual, y::Dual, z::Real)
+    vx, vy = value(x), value(y)
+    result = fma(vx, vy, z)
+    return Dual(result, _mul_partials(partials(x), partials(y), vx, vy))
+end
+
+@inline function Base.fma(x::Dual, y::Real, z::Dual)
+    vx = value(x)
+    result = fma(vx, y, value(z))
+    return Dual(result, partials(x) * y + partials(z))
+end
+
+@inline Base.fma(x::Real, y::Dual, z::Dual) = fma(y, x, z)
+
+@inline function Base.fma(x::Dual, y::Real, z::Real)
+    vx = value(x)
+    return Dual(fma(vx, y, value(z)), partials(x) * y)
+end
+
+@inline Base.fma(x::Real, y::Dual, z::Real) = fma(y, x, z)
+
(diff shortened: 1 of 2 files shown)


src/dual.jl
```diff
@@ -410,6 +410,38 @@ end
 @ambiguous @inline Base.atan2(y::Real, x::Dual) = calc_atan2(y, x)
 @ambiguous @inline Base.atan2(y::Dual, x::Real) = calc_atan2(y, x)
 
+@inline function Base.fma(x::Dual, y::Dual, z::Dual)
+    vx, vy = value(x), value(y)
+    result = fma(vx, vy, value(z))
+    return Dual(result,
```
KristofferC: Wouldn't it be better to define a `fma` function for `Partials` as well, so that `fma` instructions will be used?

tpapp: Did you mean for the 
```julia
Base.fma(x::Dual, y::Real, z::Dual)
```
case? For the others I don't see how it would help.

KristofferC: I was thinking something like (where `R` is the value part and `E` is the partials part):

```
Partials(x * y + z)
=
Partials[(Rx * [Ey1, Ey2, ...]) +
         (Ry * [Ex1, Ex2, ...])
             + [Ez1, Ez2, ...]
        ]
= 
[Rx * Ey1 + Ry * Ex1 + Ez1,
 Rx * Ey2 + Ry * Ex2 + Ez2,
 ...]
= 
[fma(Rx, Ey1, fma(Ry, Ex1, Ez1)),
 fma(Rx, Ey2, fma(Ry, Ex2, Ez2)),
 ...]
```


An implementation would be:

```
@generated function Base.fma{N}(x::Dual{N}, y::Dual{N}, z::Dual{N})
    ex = Expr(:tuple, [:(fma(value(x), partials(y)[$i], fma(value(y), partials(x)[$i], partials(z)[$i]))) for i in 1:N]...)
    return quote
        v = fma(value(x), value(y), value(z))
        Dual(v, $ex)
    end
end
```

so that:

```
 x, y, z = Dual(rand(9)...) , Dual(rand(9)...), Dual(rand(9)...);
julia> x*y + z - fma(x,y,z)
Dual(-5.551115123125783e-17,0.0,0.0,0.0,0.0,0.0,0.0,-2.220446049250313e-16,0.0)
```

This gives a bunch of fused multiply adds in the generated code:

```
	vfmadd213sd	8(%rcx), %xmm2, %xmm8
	vfmadd231sd	8(%rdx), %xmm1, %xmm8
```

but I'm not sure how big of a performance it is in general.

KristofferC: Some benchmarking with and without O3 shows that the `fma` version is just barely (5-10%) faster than a function `f(x,y,z) = x*y + z`.

tpapp: My understanding is that the point of `fma` is not speed but precision. Try
```julia
for i in 1:100
    x, y, z = rand(3)
    if fma(x,y,z) != x*y+z
        println("*")
    end
end
```

KristofferC: I thought it was both. Anyway, regarding accuracy, I guess the function I posted is better since it does more fusing then...?

tpapp: Possibly. Could you perhaps merge my PR as is, then you could add your function?

KristofferC: I let @jrevels do the merging.

jrevels:
Thanks!

PR #185, opened 2017-01-25
test against tagged DiffBase versions instead of DiffBase master

jrevels:
ref @tkelman's [comment here](https://github.com/JuliaDiff/ForwardDiff.jl/pull/179#discussion_r97707014)

PR #205, opened 2017-03-17
Cherry-pick #204 to a new release-0.3 branch that supports Julia 0.4

tkelman:
Ref https://github.com/JuliaDiff/ForwardDiff.jl/pull/185#issuecomment-287239945, would like to change the target branch to that once it's created. DiffBase 0.0.5 still supports Julia 0.4 and caused ForwardDiff 0.3.4, the last tag here that supported 0.4, to now fail its tests - https://github.com/JuliaCI/pkg.julialang.org/blob/69282b38ea4ea43a5b300862eac40d89bbf03b7a/logs/ForwardDiff_0.4.log

In local testing, this fixes it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ language: julia
 julia:
     - 0.4
     - 0.5
-    - nightly
 notifications:
     email: false
 sudo: false
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,5 +1,5 @@
-julia 0.4
-DiffBase 0.0.3
+julia 0.4 0.6
+DiffBase 0.0.1
 Compat 0.8.6
-Calculus 0.2.0
-NaNMath 0.2.2
+Calculus 0.1.15
+NaNMath 0.2.1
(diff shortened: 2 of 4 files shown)


jrevels:
If this is only to make PkgEval happy, I'm fine with doing the backport this time (since we already have everything laid out), but I guess in the future I'd rather just bump DiffBase's Julia lower bound version requirement to match ForwardDiff/ReverseDiff than take on the development overhead of supporting a Julia version which has already gone the way of the dinosaur (in my mind, anyway).

tkelman:
Next time do that before tagging a version that causes downstream test failures. If this wasn't a test addition pointing out an obvious bug with a simple fix, I'd agree.

tkelman:
Julia 0.4 is still used in quite a few places. If it isn't broken, try not to make things any worse there.

tkelman:
Good to go here, and tag as 0.3.5?

tkelman:
ping

jrevels:
Thanks!

PR #206, opened 2017-03-18
fix wrong partials multiplied in FMA

KristofferC:
The methods gave the wrong partials but this was not noticed in the tests because `==` only compares the `value` part:

```jl
julia> x, y, z = Dual(1,1,2,3), Dual(2,4,5,6), Dual(3,7,8,9);

julia> x*y + z - fma(x, y, z)
Dual(0,-3,-3,-3)
```

 Fix the bug and introduce a `dualapprox` function to test approximate results of both the value and partial part:

```jl
julia> x, y, z = Dual(1,1,2,3), Dual(2,4,5,6), Dual(3,7,8,9);

julia> x*y + z - fma(x, y, z)
Dual(0,0,0,0)
```

cc @tpapp

### DIFF
--- a/test/DualTest.jl
+++ b/test/DualTest.jl
@@ -387,20 +387,20 @@ for N in (0,3), M in (0,4), T in (Int, Float32)
 
     @test partials(NaNMath.pow(Dual(-2.0, 1.0), Dual(2.0, 0.0)), 1) == -4.0
 
-    @test fma(FDNUM, FDNUM2, FDNUM3) == Dual(fma(PRIMAL, PRIMAL2, PRIMAL3),
+    test_approx_diffnums(fma(FDNUM, FDNUM2, FDNUM3), Dual(fma(PRIMAL, PRIMAL2, PRIMAL3),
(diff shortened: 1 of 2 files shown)


mlubin:
I'm guessing the commit of ``partials.jl.mem`` was not intentional?

KristofferC:
Jeez, it's up for 15 seconds and someone notices.. :P

test/DualTest.jl
```diff
@@ -15,6 +15,8 @@ samerng() = MersenneTwister(1)
 # exponent by one
 intrand(T) = T == Int ? rand(2:10) : rand(T)
 
+dualapprox(A, B) = value(A) ≈ value(B) && partials(A) ≈ partials(B)
```
jrevels: The (outdated-ly named) `test_approx_diffnums` defined below is what is used for this in the rest of the tests. Can you reuse/rename/rewrite that function instead?

KristofferC: Sure, didn't notice it.

PR #207, opened 2017-03-18
FMA all the way down

KristofferC:
This makes the operations between partials also use FMA when possible, which is likely desirable if the user uses FMA in the first place.

Commit made on top of #206.


KristofferC:
Derivation at https://github.com/JuliaDiff/ForwardDiff.jl/pull/203#discussion_r104918938

jrevels:
Rebase against master now that #206 is merged? Thanks for these fixes!

jrevels:
Superceded by #214

Issue #220, opened 2017-04-23
crash on windows nightly


mlubin:
https://ci.appveyor.com/project/JuliaOpt/jump-jl/build/1.0.14/job/bks4v0cawpga2qur#L161

Is this a known issue?

mlubin:
(It's not obvious that ForwardDiff is even to blame)

jrevels:
Stacktrace points to [here](https://github.com/JuliaDiff/ForwardDiff.jl/blob/v0.4.2/src/dual.jl#L195). Looks like a precompilation/type inference bug? I have no idea why that would be Windows-specific, though. Did CI on other OSes pass?

We should enable AppVeyor for ForwardDiff...

mlubin:
Here's the corresponding PR: https://github.com/JuliaOpt/JuMP.jl/pull/1008
All OS/0.5/0.6 combinations pass except windows 0.6.

mlubin:
Note appveyor is running julia-debug

KristofferC:
Doesn't seem relevant anymore.

PR #221, opened 2017-04-24
enable AppVeyor

jrevels:
prompted by #220

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,7 +1,5 @@
 environment:
   matrix:
-  - JULIAVERSION: "julialang/bin/winnt/x86/0.5/julia-0.5-latest-win32.exe"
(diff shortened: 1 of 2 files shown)


jrevels:
@mlubin So it looks like everything is passing here, but I'm not using `julia-debug` so that could be the difference between here and #220. Should I be using that along with/instead of the default? I'm a total noob when it comes to AppVeyor/Windows.

mlubin:
It's worth trying ``julia-debug`` just to see if you can reproduce the issue I had. @tkelman might have some advice on which should be the default.

jrevels:
Oh, duh - ForwardDiff master is drastically different from ForwardDiff 0.4.2, so that's probably a factor as well.

jrevels:
I'm going to merge this for now, since it includes a bug fix for 32 bit systems.

PR #222, opened 2017-04-27
add eltype for AbstractConfig

jrevels:
This allows users to easily extract the types of the `Dual` instances that will be injected into their code:

```julia
julia> cfg = ForwardDiff.GradientConfig(prod, rand(20));

julia> eltype(cfg)
ForwardDiff.Dual{ForwardDiff.Tag{Base.#prod,0xb046287d533b082d},Float64,10}
```

cc @tkoolen


tkoolen:
Thanks!

PR #224, opened 2017-05-08
Accessing lower order results for StaticArrays methods

aaowens:
I wanted to use the StaticArrays methods for some numerical optimization routines, and I didn't see any way to access the lower order results using the existing methods. This PR adds methods `hessian_grad`, `jacobian_vals`, `gradient_val` to do so.

Example:
```
julia> f(x) = exp(x[1])*x[2]^x[3]

julia> sx = @SVector rand(3)
3-element SVector{3,Float64}:
 0.331614
 0.683379
 0.307809

julia> function test1(sx)
       hess = ForwardDiff.hessian(f, sx)
       grad = ForwardDiff.gradient(f, sx)
       (hess, grad)
       end
test1 (generic function with 1 method)

julia> function test2(sx)
       (hess, grad) = ForwardDiff.hessian_grad(f, sx)
       end
test2 (generic function with 1 method)

julia> @btime test1($sx)
  557.204 ns (0 allocations: 0 bytes)
([1.23916 0.558143 -0.471753; 0.558143 -0.56534 1.60079; -0.471753 1.60079 0.179599], [1.23916, 0.558143, -0.471753])

julia> @btime test2($sx)
  385.565 ns (0 allocations: 0 bytes)
([1.23916 0.558143 -0.471753; 0.558143 -0.56534 1.60079; -0.471753 1.60079 0.179599], [1.23916, 0.558143, -0.471753])
```

I'm very new at this, so apologies if this PR is improperly constructed.

### DIFF
--- a/src/gradient.jl
+++ b/src/gradient.jl
@@ -40,12 +40,8 @@ end
         return SArray{S}($result)
     end
 end
-@generated function extract_value(y::Real, ::SArray{S,X,D,N}) where {S,X,D,N}
-    result = Expr(:tuple, [:(value(y, $i)) for i in 1:N]...)
-    return quote
-        $(Expr(:meta, :inline))
-        return SArray{S}($result)
-    end
+function extract_value(y::Real, ::SArray{S,X,D,N}) where {S,X,D,N}
+    y.value
 end
 
 function extract_gradient!(out::DiffResult, y::Real)
--- a/src/gradient.jl
+++ b/src/gradient.jl
@@ -40,8 +40,12 @@ end
         return SArray{S}($result)
     end
 end
-function extract_value(y::Real, ::SArray{S,X,D,N}) where {S,X,D,N}
-   a.value
+@generated function extract_value(y::Real, ::SArray{S,X,D,N}) where {S,X,D,N}
+    result = Expr(:tuple, [:(value(y, $i)) for i in 1:N]...)
+    return quote
+        $(Expr(:meta, :inline))
+        return SArray{S}($result)
+    end
 end
 
 function extract_gradient!(out::DiffResult, y::Real)
--- a/src/gradient.jl
+++ b/src/gradient.jl
@@ -25,6 +25,7 @@ function gradient!(out, f::F, x, cfg::AllowedGradientConfig{F,H} = GradientConfi
 end
 
 @inline gradient(f::F, x::SArray) where {F} = vector_mode_gradient(f, x)
+@inline gradient_val(f::F, x::SArray) where {F} = vector_mode_gradient_val(f, x)
 
 @inline gradient!(out, f::F, x::SArray) where {F} = vector_mode_gradient!(out, f, x)
 
@@ -38,6 +39,9 @@ end
         $(Expr(:meta, :inline))
         return SArray{S}($result)
     end
+end
+function extract_value(y::Real, ::SArray{S,X,D,N}) where {S,X,D,N}
+   a.value
 end
 
 function extract_gradient!(out::DiffResult, y::Real)
@@ -87,6 +91,11 @@ end
 
 @inline function vector_mode_gradient(f::F, x::SArray) where F
     return extract_gradient(vector_mode_dual_eval(f, x), x)
+end
+
+@inline function vector_mode_gradient_val(f::F, x::SArray) where F
+    a = vector_mode_dual_eval(f, x)
+    return (extract_gradient(a, x), extract_value(a, x))
 end
 
 @inline function vector_mode_gradient!(out, f::F, x::SArray) where F
--- a/src/hessian.jl
+++ b/src/hessian.jl
@@ -31,7 +31,7 @@ function hessian!(out::DiffResult, f::F, x, cfg::AllowedHessianConfig{F,H} = Hes
 end
 
 hessian(f::F, x::SArray) where {F} = jacobian(y -> gradient(f, y), x)
-hessian_vals(f::F, x::SArray) where {F} = jacobian_vals(y -> gradient(f, y), x)
+hessian_grad(f::F, x::SArray) where {F} = jacobian_vals(y -> gradient(f, y), x)
 
 hessian!(out, f::F, x::SArray) where {F} = jacobian!(out, y -> gradient(f, y), x)
 
--- a/src/hessian.jl
+++ b/src/hessian.jl
@@ -31,6 +31,7 @@ function hessian!(out::DiffResult, f::F, x, cfg::AllowedHessianConfig{F,H} = Hes
 end
 
 hessian(f::F, x::SArray) where {F} = jacobian(y -> gradient(f, y), x)
+hessian_vals(f::F, x::SArray) where {F} = jacobian_vals(y -> gradient(f, y), x)
 
 hessian!(out, f::F, x::SArray) where {F} = jacobian!(out, y -> gradient(f, y), x)
 
--- a/src/jacobian.jl
+++ b/src/jacobian.jl
@@ -145,7 +145,7 @@ end
 
 @inline function vector_mode_jacobian_vals(f::F, x::SArray) where F
     a = vector_mode_dual_eval(f, x)
-    return (extract_jacobian(a, x), a.val)
+    return (extract_jacobian(a, x), extract_value(a,x))
 end
 
 @inline function vector_mode_jacobian!(out, f::F, x::SArray{S,V,D,N}) where {F,S,V,D,N}
--- a/src/jacobian.jl
+++ b/src/jacobian.jl
@@ -132,6 +132,11 @@ end
     return extract_jacobian(vector_mode_dual_eval(f, x), x)
 end
 
+@inline function vector_mode_jacobian_vals(f::F, x::SArray) where F
+    a = vector_mode_dual_eval(f, x)
+    return (extract_jacobian(a, x), a.val)
+end
+
 @inline function vector_mode_jacobian!(out, f::F, x::SArray{S,V,D,N}) where {F,S,V,D,N}
     ydual = vector_mode_dual_eval(f, x)
     extract_jacobian!(out, ydual, N)
--- a/src/jacobian.jl
+++ b/src/jacobian.jl
@@ -67,7 +67,7 @@ end
     result = Expr(:tuple, [:(value(ydual[$i])) for i in 1:M]...)
     return quote
         $(Expr(:meta, :inline))
-        return SArray{SX}($result)
+        return SArray{SY}($result)
     end
 end
 
--- a/src/jacobian.jl
+++ b/src/jacobian.jl
@@ -62,6 +62,15 @@ end
     end
 end
 
+@generated function extract_value(ydual::SArray{SY,VY,DY,M},
+                                     x::SArray{SX,VX,DX,N}) where {SY,VY,DY,M,SX,VX,DX,N}
+    result = Expr(:tuple, [:(value(ydual[$i])) for i in 1:M]...)
+    return quote
+        $(Expr(:meta, :inline))
+        return SArray{Tuple{M}}($result)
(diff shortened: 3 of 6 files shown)


jrevels:
Hi, thanks for the PR! 

ForwardDiff uses [DiffBase's common DiffResult API](http://www.juliadiff.org/DiffBase.jl/diffresult/) for extracting lower-order results from higher-order derivative operations. This already works with StaticArrays:

```julia
julia> using ForwardDiff, BenchmarkTools, StaticArrays

julia> f(x) = exp(x[1])*x[2]^x[3]

julia> sx = @SVector rand(3);

julia> out = DiffBase.HessianResult(sx);

julia> cfg = ForwardDiff.HessianConfig(f, out, sx);

# using your test1 function
julia> @btime test1($sx);
  318.674 ns (0 allocations: 0 bytes)

julia> test2(out, sx, cfg) = ForwardDiff.hessian!(out, f, sx, cfg)
test2 (generic function with 1 method)

julia> @btime test2($out, $sx, $cfg);
  404.150 ns (3 allocations: 96 bytes)
```

[The StaticArrays fallback this code is hitting](https://github.com/JuliaDiff/ForwardDiff.jl/blob/master/src/hessian.jl#L37) is pretty slow, even compared to the naive `test1`. Instead of going through that fallback (and similar ones for the other API methods), we should just add specialized methods for StaticArrays. 

That way, we can achieve the same performance you've demonstrated here without needing new API functions.

Note that this may require adding new definitions for `HessianResult(::SArray)` etc. to DiffBase - for our purposes here, it'd likely be cheaper just to swap `DiffResult` fields than to fill a mutable array.

aaowens:
Sorry, I should have clarified that I knew about the DiffBase approach. As you said, it is pretty slow. I believe this is because it is allocating even for StaticArrays. On my computer, a more complete example is

```
julia> using ForwardDiff, StaticArrays, BenchmarkTools

julia> f(x) = exp(x[1])*x[2]^x[3]

julia> sx = @SVector rand(3);

julia> out = DiffBase.HessianResult(sx);

julia> cfg = ForwardDiff.HessianConfig(f, out, sx);

julia> function test1(sx)
         hess = ForwardDiff.hessian(f, sx)
         grad = ForwardDiff.gradient(f, sx)
         (hess, grad)
       end
test1 (generic function with 1 method)

julia> function test2(sx)
         (hess, grad) = ForwardDiff.hessian_grad(f, sx)
       end
test2 (generic function with 1 method)

julia> function test3(out, sx, cfg)
         ForwardDiff.hessian!(out, f, sx, cfg)
         (DiffBase.hessian(out), DiffBase.gradient(out))
       end
test3 (generic function with 1 method)

julia> @btime test1($sx)
  543.452 ns (0 allocations: 0 bytes)
([0.182774 1.4217 -0.396623; 1.4217 -1.39324 -1.48429; -0.396623 -1.48429 0.860677], [0.182774, 1.4217, -0.396623])

julia> @btime test2($sx)
  376.254 ns (0 allocations: 0 bytes)
([0.182774 1.4217 -0.396623; 1.4217 -1.39324 -1.48429; -0.396623 -1.48429 0.860677], [0.182774, 1.4217, -0.396623])

julia> @btime test3($out, $sx, $cfg)
  553.856 ns (4 allocations: 128 bytes)
([0.182774 1.4217 -0.396623; 1.4217 -1.39324 -1.48429; -0.396623 -1.48429 0.860677], [0.182774, 1.4217, -0.396623])
```

If you think it's possible to obtain the same performance through DiffBase, that would be great.

jrevels:
Apologies for the late response, I've been on vacation.

> If you think it's possible to obtain the same performance through DiffBase, that would be great.

I think it should be, if we add specialized definitions for `HessianResult(::SArray)` etc. to the `DiffResult` API. Currently, these methods allocate mutable arrays, then let ForwardDiff/ReverseDiff load values into them later. Instead, if an `SArray` is provided,  we could just initialize an instance of the type, and then let ForwardDiff/ReverseDiff switch the fields to point at the resulting `SArray`. This pointer switch should be negligible compared to the cost of the computation:

```julia
julia> using BenchmarkTools

julia> mutable struct Foo
           data::Array{Float64,2}
       end

julia> x, y = rand(10, 10), rand(10, 10);

julia> fx, fy = Foo(x), Foo(y);

julia> @benchmark begin
           tmp = $fx.data
           $fx.data = $fy.data
           $fy.data = tmp
       end
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     8.465 ns (0.00% GC)
  median time:      8.645 ns (0.00% GC)
  mean time:        8.657 ns (0.00% GC)
  maximum time:     16.924 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     999
```

aaowens:
I think you are right that the pointer switch is pretty negligible; however, allocating a `DiffResult` is not free. I made an example for the scalar case. `test1` uses the DiffResult, and I think it uses exactly the approach you propose here. `test2` constructs a `Dual` directly and uses that to differentiate without any allocation.

```

julia> function test1(x, N)
         (fx, jx) = zero(x), zero(x)
         res = DiffBase.DiffResult(fx, jx)
         for i = 1:N
           ForwardDiff.derivative!(res, exp, x + 1/i)
           fx = DiffBase.value(res)
           jx = DiffBase.derivative(res)
         end
         (fx, jx)
       end
test1 (generic function with 1 method)

julia> function test2(x, N)
         (fx, jx) = zero(x), zero(x)
         for i = 1:N
           res = exp(ForwardDiff.Dual(x + 1/i, 1.))
           (fx, jx) = res.value, res.partials[1]
         end
         (fx, jx)
       end
test2 (generic function with 1 method)

julia> x, N = 2., 1
(2.0, 1)

julia> @btime test1($x, $N)
  33.942 ns (1 allocation: 32 bytes)
(20.085536923187668, 20.085536923187668)

julia> @btime test2($x, $N)
  23.982 ns (0 allocations: 0 bytes)
(20.085536923187668, 20.085536923187668)

julia> N = 4
4

julia> @btime test1($x, $N)
  79.815 ns (1 allocation: 32 bytes)
(9.487735836358526, 9.487735836358526)

julia> @btime test2($x, $N)
  66.329 ns (0 allocations: 0 bytes)
(9.487735836358526, 9.487735836358526)

julia> N = 100
100

julia> @btime test1($x, $N)
  1.596 μs (1 allocation: 32 bytes)
(7.463317347319193, 7.463317347319193)

julia> @btime test2($x, $N)
  1.447 μs (0 allocations: 0 bytes)
(7.463317347319193, 7.463317347319193)
```

Whether this matters depends on the problem. For simple root-finding with an initial guess close to the solution, `N=4` seems reasonable, so there is a ~20% performance loss from using `DiffResult`.

jrevels:
Just to reiterate: I'm claiming we should add new constructor methods/types to DiffBase, not use the old ones (e.g. a `HessianResult(::SArray)` might return some new type, like a `StaticDiffResult`).

To reply to your benchmark using the existing interface: The allocation you see here is expected, since the code constructs a mutable DiffResult. In practice, of course, one would appropriately preallocate any buffers required for performance critical code paths instead of using allocating methods. For example, this benchmark is a more reasonable comparison, IMO:

```julia
julia> function test1!(res, x, N)
           for i in 1:N
               ForwardDiff.derivative!(res, exp, x + 1/i)
           end
           return res
       end
test1! (generic function with 1 method)

julia> x, N = 2.0, 4
(2.0, 4)

julia> res = DiffBase.DiffResult(x, x)
DiffBase.DiffResult{1,Float64,Tuple{Float64}}(2.0, (2.0,))

julia> @btime test1!($res, $x, $N)
  58.026 ns (0 allocations: 0 bytes)
DiffBase.DiffResult{1,Float64,Tuple{Float64}}(9.487735836358526, (9.487735836358526,))

julia> @btime test2($x, $N)
  59.679 ns (0 allocations: 0 bytes)
(9.487735836358526, 9.487735836358526)
```

The point here is that the allocation is O(1) w.r.t. calls to the differentiation routine `df!` (for example, here, `df!` is `test1!` for some `N`). If `df!` usage is such that the single allocation is costly in comparison to the computation, it implies a run time on the order of nanoseconds, in which case `df!` isn't performance critical anyway. Of course, if you are calling `df!` many times, those nanoseconds might add up to something significant. By that point, however, the cost of the allocation becomes negligible.

For large N, the main timing difference is likely solely from the extra pointer maneuvering:

```julia
julia> N = 1000
1000

julia> @btime test1!($res, $x, $N);
  14.270 μs (0 allocations: 0 bytes)

julia> @btime test2($x, $N);
  12.927 μs (0 allocations: 0 bytes)
```

If we add new methods to DiffBase, we could make the new types statically allocated and avoid the pointer switch altogether, and still follow the non-mutating parts of the `DiffResult` API. 

It'll be easier if we have the DiffBase code before doing any refactoring here. I'll start working on make a PR to DiffBase, and then after that's merged, we can revisit DiffResult usage in ForwardDiff.

aaowens:
Cool, that sounds very promising. I hadn't known there was a non-mutating part of the `DiffBase` API.

aaowens:
The `ImmutableDiffResult` looks like it will work. I tried it here:

```
julia> using DiffBase, StaticArrays, ForwardDiff, BenchmarkTools

julia> v = rand()
0.25802151701921927

julia> sd = SVector{3}(rand(3));

julia> sr = DiffBase.DiffResult(v, sd)
ImmutableDiffResult(0.25802151701921927, ([0.907497, 0.625286, 0.771794],))

julia> function ForwardDiff.extract_gradient!(out::DiffBase.ImmutableDiffResult, dual::ForwardDiff.Dual)
         out = DiffBase.DiffResult(ForwardDiff.value(dual), SVector(ForwardDiff.partials(dual).values))
         return out
       end

julia> f(x) = sum(sin.(x))

julia> sr2 = ForwardDiff.gradient!(sr, f, sd)
ImmutableDiffResult(2.0707163748962127, ([0.61572, 0.810796, 0.71666],))

julia> typeof(sr)
DiffBase.ImmutableDiffResult{1,Float64,Tuple{SVector{3,Float64}}}

julia> typeof(sr2)
DiffBase.ImmutableDiffResult{1,Float64,Tuple{SVector{3,Float64}}}
```

There is probably a cleaner way to extract the gradient without the overhead of the way I did it, going from Partials to a tuple to a StaticArray.

The performance looks good too. With -O3 it is almost as fast as my direct approach. 

```
julia> function test1(sr, sd)
         ForwardDiff.gradient!(sr, f, sd)
       end
test1 (generic function with 1 method)

julia> function test2(sd)
         sr = DiffBase.DiffResult(1., @SArray zeros(3))
         ForwardDiff.gradient!(sr, f, sd)
       end
test2 (generic function with 1 method)

julia> function test3(sd)
         ForwardDiff.gradient_val(f, sd)
       end
test3 (generic function with 1 method)

julia> @btime test1($sr, $sd)
  69.700 ns (0 allocations: 0 bytes)
ImmutableDiffResult(2.0707163748962127, ([0.61572, 0.810796, 0.71666],))

julia> @btime test2($sd)
  62.203 ns (0 allocations: 0 bytes)
ImmutableDiffResult(2.0707163748962127, ([0.61572, 0.810796, 0.71666],))

julia> @btime test3($sd)
  58.797 ns (0 allocations: 0 bytes)
([0.61572, 0.810796, 0.71666], 2.0707163748962127)
```

aaowens:
I have `ImmutableDiffResult` methods mostly working at https://github.com/aaowens/ForwardDiff.jl/tree/Immutable , but no tests yet. As it is now `hessian!` works, but you need `DiffBase.jacobian(out)` to access the results, for some reason `hessian` doesn't work.

aaowens:
I'm closing this since I put the `ImmutableDiffResult` methods in #236 .

PR #225, opened 2017-05-08
Fix year in README

alhirzel:


jrevels:
Note to self: never update a README from within a time machine! 😛 

Thanks!

PR #232, opened 2017-05-25
add specialized muladd

KristofferC:
The code is exactly the same as for `fma`. I tried to create both fma and muladd in a loop using `@eval` but it was tricky to interpolate things into the generated function expression etc. After a while I just did the copy paste...

The difference between `fma` and `muladd` is that `fma` forces fused multiply add for accuracy, while `muladd` is a performance optimization that the compiler can chose not to do if the cpu does not support the fma operation.


jrevels:
Thanks!

PR #236, opened 2017-06-01
Add ImmutableDiffResult methods

aaowens:
This is an attempt to add methods to `gradient!` and other `!` methods which return ImmutableDiffResults. It passes the `Number` and StaticArrays tests once they are modified to return the result instead of trying to mutate it. 

The `ImmutableDiffResult` input is just for dispatch purposes and isn't actually used anywhere. It might be desirable to add some type signatures to make sure the returned type is the same as the input type.


src/derivative.jl
```diff
@@ -37,6 +37,22 @@ end
...
+end
+
+@inline function derivative!(::ImmutableDiffResult, f!::F, y, x::R, cfg::AllowedDerivativeConfig{F,H} = DerivativeConfig(f!, y, x)) where {F,R<:Real,H}
+    ydual = cfg.duals
+    seed!(ydual, y)
+    f!(ydual, Dual{Tag{F,H}}(x, one(x)))
+    map!(value, y, ydual)
+    out = DiffResult(map(value, ydual), extract_derivative(ydual))
+    return out
+end
```
jrevels: Instead of adding new methods here, I think you can just update `extract_value!` and `extract_derivative!` callsites to alias the output, right?

src/gradient.jl
```diff
@@ -28,6 +28,11 @@ end
 
 @inline gradient!(out, f::F, x::SArray) where {F} = vector_mode_gradient!(out, f, x)
 
+@inline function gradient!(out::ImmutableDiffResult, f, x::SArray)
+    dual = vector_mode_dual_eval(f, x)
+    out = DiffResult(value(dual), extract_gradient(dual, x))
+end
```
jrevels: I think the same is true here (and similar methods) - these methods shouldn't be necessary, since the `DiffBase` API should do all this work for us as long as we switch over to the new output-aliasing convention.

aaowens: Do you mean, for example:
```
function extract_gradient!(out::DiffResult, dual::Dual)
    out = DiffBase.value!(out, value(dual))
    out = DiffBase.gradient!(out, partials(dual))
    return out
end
```
That would be simpler for sure. I worried about performance by constructing the ImmutableDiffResult in pieces, but it's probably negligible.

jrevels: Yup. Theoretically, I don't think performance will suffer as long as everything inlines properly. Definitely worth benchmarking though.

aaowens: I tried it in a different branch. The performance seems fine. The `!` methods returning a `DiffResult`  have performance within 10% of the non-`!` methods in the (limited) benchmarking I did.

I saw no way to use the existing methods for hessians since they seem to rely on mutating for the value, so I tried this:
```
function hessian!(result::ImmutableDiffResult, f::F, x::SArray) where {F} 
    d1 = dualize(f, x)
    d2 = dualize(f, d1)
    fd2 = f(d2)
    val = value(value(fd2))
    grad = extract_gradient(value(fd2), x)
    hess = extract_jacobian(partials(fd2), x)
    result = DiffBase.hessian!(result, hess)
    result = DiffBase.gradient!(result, grad)
    result = DiffBase.value!(result, val)
end
```
Do you think this method is a workable solution? It has 1 allocation in my tests, not sure why. The performance is fine. For the `prod` example in the tests, it is actually faster than `hessian`.

aaowens:
This works fairly well know, except for the one allocation on `hessian!` 

```
julia> using ForwardDiff, StaticArrays, BenchmarkTools

julia> f(x) = sum(sin.(x))

julia> sx = SVector(1., 1., 1.)
3-element SVector{3,Float64}:
 1.0
 1.0
 1.0

julia> gout = DiffBase.GradientResult(sx)
ImmutableDiffResult(1.0, ([1.0, 1.0, 1.0],))

julia> hout = DiffBase.HessianResult(sx)
ImmutableDiffResult(1.0, ([1.0, 1.0, 1.0], [0.0 0.0 0.0; 0.0 0.0 0.0; 0.0 0.0 0.0]))

julia> @btime ForwardDiff.gradient(f, $sx)
  81.182 ns (0 allocations: 0 bytes)
3-element SVector{3,Float64}:
 0.540302
 0.540302
 0.540302

julia> @btime ForwardDiff.gradient!($gout, f, $sx)
  92.716 ns (0 allocations: 0 bytes)
ImmutableDiffResult(2.5244129544236893, ([0.540302, 0.540302, 0.540302],))

julia> @btime ForwardDiff.hessian(f, $sx)
  237.317 ns (0 allocations: 0 bytes)
3×3 StaticArrays.SArray{Tuple{3,3},Float64,2,9}:
 -0.841471   0.0        0.0     
  0.0       -0.841471   0.0     
  0.0        0.0       -0.841471

julia> @btime ForwardDiff.hessian!($hout, f, $sx)
  288.354 ns (1 allocation: 160 bytes)
ImmutableDiffResult(2.5244129544236893, ([0.540302, 0.540302, 0.540302], [-0.841471 0.0 0.0; 0.0 -0.841471 0.0; 0.0 0.0 -0.841471]))

julia> f2(x) = exp.(x) + f(x)
f2 (generic function with 1 method)

julia> jout = DiffBase.JacobianResult(sx)
ImmutableDiffResult([1.0, 1.0, 1.0], ([0.0 0.0 0.0; 0.0 0.0 0.0; 0.0 0.0 0.0],))

julia> @btime ForwardDiff.jacobian(f2, $sx)
  107.196 ns (0 allocations: 0 bytes)
3×3 StaticArrays.SArray{Tuple{3,3},Float64,2,9}:
 3.25858   0.540302  0.540302
 0.540302  3.25858   0.540302
 0.540302  0.540302  3.25858 

julia> @btime ForwardDiff.jacobian!($jout, f2, $sx)
  108.995 ns (0 allocations: 0 bytes)
ImmutableDiffResult([5.24269, 5.24269, 5.24269], ([3.25858 0.540302 0.540302; 0.540302 3.25858 0.540302; 0.540302 0.540302 3.25858],))

julia> x = 1.
1.0

julia> @btime ForwardDiff.derivative(exp, $x)
  5.402 ns (0 allocations: 0 bytes)
2.718281828459045

julia> out = ForwardDiff.DiffResult(x, x)
ImmutableDiffResult(1.0, (1.0,))

julia> @btime ForwardDiff.derivative!($out, exp, $x)
  5.766 ns (0 allocations: 0 bytes)
ImmutableDiffResult(2.718281828459045, (2.718281828459045,))

```

src/hessian.jl
```diff
@@ -58,5 +58,17 @@ hessian(f::F, x::SArray, cfg::AllowedHessianConfig{F,H}) where {F,H} = hessian(f
...
+    d2 = dualize(f, d1)
+    fd2 = f(d2)
+    val = value(value(fd2))
+    grad = extract_gradient(value(fd2), x)
+    hess = extract_jacobian(partials(fd2), x)
+    result = DiffBase.hessian!(result, hess)
+    result = DiffBase.gradient!(result, grad)
+    result = DiffBase.value!(result, val)
+end
+hessian!(result::ImmutableDiffResult, f::F, x::AbstractArray, cfg::AllowedHessianConfig{F,H} = HessianConfig(f, result, x)) where {F,H} = hessian!(result, f, x)
```
jrevels: Hmmm isn't this method self-recursive? We should probably drop it.

This is looking great by the way, really looking forward to it landing on master! Thanks again for your contributions.

src/jacobian.jl
```diff
@@ -166,11 +176,20 @@ end
...
 end
 
+@inline function vector_mode_jacobian!(out::ImmutableDiffResult, f::F, x::SArray{S,V,D,N}) where {F,S,V,D,N}
+    ydual = vector_mode_dual_eval(f, x)
+    jout = extract_jacobian(ydual, x)
+    vout = extract_value(ydual, DiffBase.value(out))
+    out = DiffBase.jacobian!(out, jout)
+    out = DiffBase.value!(out, vout)
+    return out
+end
```
jrevels: Isn't this covered by the method defined directly above (`vector_mode_jacobian!(result, f::F, x::SArray{S,V,D,N})`)? What's the reasoning for a new method here?

aaowens: I ran into two issues here.
1. `extract_jacobian!(result, ydual, N)` as written doesn't work because you cannot `setindex` on a reshaped `SArray`.
2. I think `extract_value!(result, ydual)` did work, but was slow.

Since the `@generated` function already existed for `extract_jacobian`, and it was straightforward to copy it for values, I used that approach here. I pass the value in `extract_value(ydual, DiffBase.value(out))` to get the shape right. In one of the tests the dimensions of `value(out)` are not the same as the native dimensions of `f(x)`.

src/hessian.jl
aaowens: Not quite sure what you mean. I think `hessian!` needs an equivalent to `hessian(f::F, x::SArray, cfg::AllowedHessianConfig{F,H}) where {F,H} = hessian(f, x)` . This does that.

Is the method below better? I didn't mean to define `cfg` as an optional argument.
```
hessian!(result::ImmutableDiffResult, f::F, x::AbstractArray, cfg::AllowedHessianConfig{F,H}) where {F,H} = hessian!(result, f, x)
```

jrevels: Yup, that optional argument was the source of my confusion. Also `x::AbstractArray` - I believe that should be `x::SArray`. Otherwise, we'd be throwing away a `cfg` that would need to get reconstructed anyway (e.g. if this method was called with `x::Array`).

aaowens: Ok

aaowens:
I don't have anything more to add right now. The tests pass locally, but rely on the master version of DiffBase.

The main outstanding issue I see is the allocation and slowness of `hessian!`. It might have something do with calling `dualize` twice. I think it's possible to use just one call with a `dualize2` function, but I don't know how to write it.

jrevels:
This was superseded by #245, which includes the work here but rebased against master + some additional stuff needed to get tests passing. 

@aaowens I gave you authorship of [the squashed commit](https://github.com/JuliaDiff/ForwardDiff.jl/commit/98af046aad0d3875ec8d8498f8e84d972ebaf96c
) that made it onto master from #245, since the meat of it was your work. Thanks again for the contribution!

aaowens:
Great! These efficient stack allocated methods are useful for my work.

PR #244, opened 2017-07-24
add missing Partials{0} definitions

jrevels:
Prompted by the following code, which was failing when `+(::Partials{0}, ::Partials{N})` fell back to the `AbstractVector` method:

```julia
ForwardDiff.hessian(x -> sum(hypot.([1,2,3], x)), [1,2,3])
```

cc @ranjanan


ranjanan:
Thanks a lot Jarrett :-)

PR #245, opened 2017-07-28
bump DiffBase upper bound to v0.4.0

jrevels:


jrevels:
Getting tests passing here requires https://github.com/JuliaDiff/DiffBase.jl/pull/13

PR #262, opened 2017-09-18
switch DiffBase dependency over to DiffRules/DiffTests/DiffResults, remove RealInterface dependency

jrevels:
This requires DiffRules, DiffTests, and DiffResults to be registered/tagged


fredrikekre:
Oh, it will be fixed by #269. Nice!

PR #279, opened 2017-10-25
tweak tag generation

simonbyrne:
This does 3 related things
1. Re-enables void tags by providing `nothing` as function argument
2. Allows opt out of tag checking via `Val{false}()` argument.
3. Simplifies the Hessian to use the same tag (which is useful for my own purposes of computing Riemannian metrics, but also makes the code a bit cleaner).

### DIFF
--- a/src/derivative.jl
+++ b/src/derivative.jl
@@ -15,14 +15,16 @@ This method assumes that `isa(f(x), Union{Real,AbstractArray})`.
 end
 
 """
-    ForwardDiff.derivative(f!, y::AbstractArray, x::Real, cfg::DerivativeConfig = DerivativeConfig(f!, y, x))
+    ForwardDiff.derivative(f!, y::AbstractArray, x::Real, cfg::DerivativeConfig = DerivativeConfig(f!, y, x), check=Val{true}())
 
 Return `df!/dx` evaluated at `x`, assuming `f!` is called as `f!(y, x)` where the result is
 stored in `y`.
+
+Set `check` to `Val{false}()` to disable tag checking. This can lead to perturbation confusion, so should be used with care.
 """
 @inline function derivative(f!, y::AbstractArray, x::Real,
-                            cfg::DerivativeConfig{T} = DerivativeConfig(f!, y, x)) where {T}
-    checktag(T, f!, x)
+                            cfg::DerivativeConfig{T} = DerivativeConfig(f!, y, x), ::Val{CHK}=Val{true}()) where {T, CHK}
+    CHK && checktag(T, f, x)
     ydual = cfg.duals
     seed!(ydual, y)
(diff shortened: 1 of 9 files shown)


simonbyrne:
@jrevels I take it you're in favour of this? If so, I'll update the docs.

jrevels:
> I take it you're in favour of this? If so, I'll update the docs.

Yup! Good stuff, thanks!

simonbyrne:
okay, I've updated the docs

src/derivative.jl
```diff
@@ -15,14 +15,16 @@ This method assumes that `isa(f(x), Union{Real,AbstractArray})`.
 end
 
 """
-    ForwardDiff.derivative(f!, y::AbstractArray, x::Real, cfg::DerivativeConfig = DerivativeConfig(f!, y, x))
+    ForwardDiff.derivative(f!, y::AbstractArray, x::Real, cfg::DerivativeConfig = DerivativeConfig(f!, y, x), check=Val{true}())
```
jrevels: Instead of doing this, can't we just disable checking by dispatching on the `Void` tag? That would be cleaner and less of an API change.

ChrisRackauckas: That's what I was referring to here: https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/200#issuecomment-340879224 in that now a user cannot turn off the tag checking in a call to another package if you can't send that extra argument.

simonbyrne: Short answer: no, and we should generally discourage use of the `Void` tag, especially in packages.

The problem is that it reopens the door to perturbation confusion. For example, [its use in NLsolve.jl](https://github.com/JuliaNLSolvers/NLsolve.jl/blob/535b7361c700b39b6f1f1148f805c6036491ef55/src/autodiff.jl#L7) means that if the function passed to it _also_ uses `Void` tags then there is potential for confusion.

A better pattern is for the library to generate its own tag, and use that, similar to what I did for Hessian:
https://github.com/JuliaDiff/ForwardDiff.jl/pull/279/files#diff-7957397746d48056ebb4bb3af5bab442R14

ChrisRackauckas: The problem with the previous change though was that generating your own tag no longer works, which is what I was doing before.

simonbyrne:
After a chat, I think we've fixed @ChrisRackauckas's issues. This should be good to merge and tag.

ChrisRackauckas:
Indeed its fixed

jrevels:
Can we add some tests for the `check = Val{false}()` case first that demonstrates the functionality we need?

simonbyrne:
Whoops, this slipped off my mind. I've now added some tests.

jrevels:
Thanks!

Issue #305, opened 2018-02-18
FieldVectors


zsunberg:
Hi all,

We have a big codebase that uses `FieldVector`s from static arrays. It would be extremely powerful for us (and I'm sure many others in the future) to be able to do the following:

```julia
using StaticArrays
using ForwardDiff

struct Point{R<:Real} <: FieldVector{2,R}
    x::R
    y::R
end

StaticArrays.similar_type(p::Type{P}, ::Type{R}, size::Size{(2,)}) where {P<:Point, R<:Real} = Point{R}

f(p) = p.x
ForwardDiff.gradient(f, Point(2.0,3.0))
```

Currently this fails with `ERROR: LoadError: type MArray has no field x` because it's trying to evaluate `f` with an MArray of Duals.

This seems pretty easy to fix - maybe just writing a new version of `dualize` and adding a little more code where you have specialized things for `SArray` to also handle `FieldVector`. Do you think it is that simple? Should we take a crack at it, or do you foresee massive difficulties?

cc: @kylejbrown17 @maximebouton

jrevels:
> This seems pretty easy to fix - maybe just writing a new version of dualize and adding a little more code where you have specialized things for SArray to also handle FieldVector. Do you think it is that simple? Should we take a crack at it, or do you foresee massive difficulties?

Seems like a good idea! 

If `FieldVector`s have similar indexing behavior to `SVector`s, you might even be able to just tweak the current `SArray`-specialized methods in order to achieve this (instead of adding new methods just for `FieldVector`).

KristofferC:
This works now

```
julia> ForwardDiff.gradient(f, Point(2.0,3.0))
2-element Point{Float64} with indices SOneTo(2):
 1.0
 0.0
```

PR #307, opened 2018-02-28
Added Functionality for FieldVectors (from StaticArrays)

kylejbrown17:
Hi Team,

This is my attempt to extend all `SArray` functionality to `FieldVectors`. This resolves issue #305. The basic approach mostly involves replacing `SArray` with `Union{FieldVector, SArray}` in method definitions.  I have included tests to mirror those used for testing `SArray` codepaths for Gradients, Hessians and Jacobians.

### DIFF
--- a/src/apiutils.jl
+++ b/src/apiutils.jl
@@ -29,6 +29,18 @@ end
 
 @inline static_dual_eval(::Type{T}, f, x::SArray) where {T} = f(dualize(T, x))
 
+########### FieldVectors ###########
+@generated function dualize(::Type{T}, x::FieldVector{N,V}) where {T,V,N}
+    dx = Expr(:tuple, [:(Dual{T}(x[$i], chunk, Val{$i}())) for i in 1:N]...)
+    return quote
+        chunk = Chunk{N}()
+        $(Expr(:meta, :inline))
+        return similar_type($x, Dual{T,V,N})($(dx))
+    end
+end
+@inline static_dual_eval(::Type{T}, f, x::FieldVector) where {T} = f(dualize(T, x))
+####################################
+
 function vector_mode_dual_eval(f, x, cfg::Union{JacobianConfig,GradientConfig})
     xdual = cfg.duals
     seed!(xdual, x, cfg.seeds)
--- a/src/apiutils.jl
+++ b/src/apiutils.jl
@@ -17,29 +17,19 @@ end
 ###################################
 # vector mode function evaluation #
 ###################################
-
-@generated function dualize(::Type{T}, x::SArray{S,V,D,N}) where {T,S,V,D,N}
+@generated function dualize(::Type{T}, x::Union{FieldVector, SArray}) where {T}
+    N = length(x)
+    V = eltype(x)
     dx = Expr(:tuple, [:(Dual{T}(x[$i], chunk, Val{$i}())) for i in 1:N]...)
     return quote
-        chunk = Chunk{N}()
+        chunk = Chunk{$N}()
         $(Expr(:meta, :inline))
-        return SArray{S}($(dx))
+        # return SArray{S}($(dx))
+        similar_type($x,Dual{T,$V,$N})($(dx))
     end
 end
 
-@inline static_dual_eval(::Type{T}, f, x::SArray) where {T} = f(dualize(T, x))
-
-########### FieldVectors ###########
-@generated function dualize(::Type{T}, x::FieldVector{N,V}) where {T,V,N}
-    dx = Expr(:tuple, [:(Dual{T}(x[$i], chunk, Val{$i}())) for i in 1:N]...)
-    return quote
-        chunk = Chunk{N}()
-        $(Expr(:meta, :inline))
-        return similar_type($x, Dual{T,V,N})($(dx))
-    end
-end
-@inline static_dual_eval(::Type{T}, f, x::FieldVector) where {T} = f(dualize(T, x))
-####################################
+@inline static_dual_eval(::Type{T}, f, x::Union{FieldVector, SArray}) where {T} = f(dualize(T, x))
 
 function vector_mode_dual_eval(f, x, cfg::Union{JacobianConfig,GradientConfig})
     xdual = cfg.duals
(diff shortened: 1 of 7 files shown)


zsunberg:
@kylejbrown17 thanks! I added some more tests, and it looks like there is a stack overflow when calculating the jacobian. We should track this down and fix it before merging. Can you give me push access to your fork of this repo so that I can push to it? Thanks!

zsunberg:
Oh, ok, I see the problem... Not sure if it is by design. The issue is in StaticArrays.jl. Will file an issue there. I fixed the tests, so this should be ready to merge.

src/apiutils.jl
```diff
@@ -17,17 +17,19 @@ end
 ###################################
 # vector mode function evaluation #
 ###################################
-
-@generated function dualize(::Type{T}, x::SArray{S,V,D,N}) where {T,S,V,D,N}
+@generated function dualize(::Type{T}, x::Union{FieldVector, SArray}) where {T}
+    N = length(x)
+    V = eltype(x)
```
jrevels: The type constraint on `x` allows this (and the other similar changes in this PR) to technically work, since we can probably assume that no upstream code will re-overload `length` or `eltype` for `Union{FieldVector, SArray}`. 

However, it's a bit sketchy...it makes it easier for bugs to accidentally be introduced in the future via seemingly innocent changes (e.g. loosening the type constraint on `x`).

Would it be hard to just pull these out of the type parameter directly instead?

kylejbrown17: It seems like calling `length` and `eltype` makes this more straightforward to extend to additional types. @zsunberg and I decided to retrieve `N` and `V` via `length` and `eltype` because it was already quite complicated to try to extract them directly from the type parameters.

I've tried a few different variations on this:
```julia
@generated function dualize2(::Type{T}, x::Union{FieldVector{N,V}, SArray{S,V,D,N}}) where {T,S,V,D,N}
```
Everything I've tried fails when called on an instance of `FieldVector`, although most of the approaches seem to work on `SVector`. I'm not 100% sure what you mean by "loosen the type constraint", but it seems like that would cause an even bigger headache if we want to extract the parameters this way.
 
If we tried to extract type parameters directly within the function body, we'd need a separate case for each type (because `V` and `N` are in different places for each type):
```julia
N,V = supertype(typeof(x)).parameters              # <: FieldVector
S,V,D,N = typeof(x).parameters             # SArray
```

Either of the above ways requires more customization than simply calling `length` and `eltype`. I think this way is more stable. It should work for vector types for which `length` and `eltype` are defined.

Another option is to have two different function definitions in every case where we need `V` and `N`.

jrevels: > I think this way is more stable. It should work for vector types for which length and eltype are defined.

I think you might misunderstand my concern, which is my bad - I didn't explain the issue directly enough. If this were normal code, then of course I'd be fine with `length` and `eltype` getting used here. 

However, this code is being called in an `@generated` thunk, and in general, calling overloadable methods inside of `@generated` function thunks is something that should be done with care, because downstream code can cause the generator to be "invalidated" in ways the compiler can't easily predict. Such code can cause world-age errors or silently incorrect method lookups, and are very hard for end-users to fix/debug/understand because they can depend on precompilation settings, code load order, etc. Thus, we want to be very careful about potentially introducing these kinds of bugs.

In this case, since `FieldVector` is an abstract type, so I could subtype it and overload `length` and `eltype` on that subtype. For the sake of example, let's call this subtype `MyFieldVectorSubtype`.  If a user loaded ForwardDiff, called `dualize` on some other `<:FieldVector`, then loaded the code that defines `FieldVectorSubtype`, then called `dualize` on an instance of `MyFieldVectorSubtype`, the compiler could possibly throw a world-age error (most likely), silently call the wrong methods, run into some other compilation error, etc.

While this may seem like a niche class of bugs, they can crop up in practice quite regularly; for one ForwardDiff-related instance of such a bug, see https://github.com/JuliaArrays/StaticArrays.jl/issues/316.

By the way, I really appreciate this PR, and am excited for this feature!

zsunberg: Ah ok, yeah, we didn't understand the specific concerns with @generated

What about
```julia
@generated function dualize(::Type{T}, x::Union{FieldVector{N,V}, SArray{<:Any,V,<:Any,N}}) where {T,V,N}
```
?

zsunberg: (we didn't know about the <:Any syntax in the type parameters)

zsunberg: If that doesn't work, I think two dualize functions would be a little simpler than _length and _eltype

src/apiutils.jl
```diff
@@ -17,17 +17,19 @@ end
...
     dx = Expr(:tuple, [:(Dual{T}(x[$i], chunk, Val{$i}())) for i in 1:N]...)
     return quote
-        chunk = Chunk{N}()
+        chunk = Chunk{$N}()
         $(Expr(:meta, :inline))
-        return SArray{S}($(dx))
+        # return SArray{S}($(dx))
```
zsunberg: We should delete this commented out line.

src/apiutils.jl
kylejbrown17: @jrevels, thanks for clarifying! This makes more sense. We'll incorporate the changes you suggested and get back to you.

src/apiutils.jl
kylejbrown17: Ah, sorry about that. Not sure why I left that sitting in there.

oschub:
I was interested in using `FieldVectors` in combination with `ForwardDiff` and asked in the Julia Slack channel for help since this PR has already been open for a long time. Mateusz Baran suggested the following: 

_ForwardDiff.dualize calls StaticArrays.similar_type at generation time, not call time, which means new methods of that function may or may not be picked (like the one you've added). This modification solves it:_
```
@generated function ForwardDiff.dualize(::Type{T}, x::StaticArray) where T
    N = length(x)
    dx = Expr(:tuple, [:(ForwardDiff.Dual{T}(x[$i], chunk, Val{$i}())) for i in 1:N]...)
    return quote
        $(Expr(:meta, :inline))
        chunk = ForwardDiff.Chunk{$N}()
        V = StaticArrays.similar_type($x, $(ForwardDiff.Dual{T,eltype(x),N}))
        return V($(dx))
    end
end
```
_I think this should be fine. I'd say my version is safer than the one currently present in ForwardDiff but both the current one and mine are a bit less safe then what was desired in that PR. I really doubt anyone would overload `eltype` or `length` for a subtype of `FieldVector` but that's actually where it's a bit unsafe._

To be honest these generated functions are a bit over my head but I wanted to share the suggestion from Mateusz Baran here to perhaps revive the discussion in this PR and to prevent the suggestion from getting lost due to Slack's message history limit.

PR #315, opened 2018-04-04
specialize on input function on a few more wrapper functions

KristofferC:
Benchmark

```jl
using ForwardDiff
using BenchmarkTools

function run0(x, f, n)
    out = similar(x);
    cfg = ForwardDiff.GradientConfig(f, x, ForwardDiff.Chunk{5}());
    for i = 1:n
        ForwardDiff.gradient!(out, f, x, cfg);
    end
    return out
end

V(x) = (dot(x,x)-1)^2;

x5 = rand(5);

@btime run0(x5, V, 100);
```

Before:
```
  42.694 μs (103 allocations: 6.92 KiB)
```

After:
```
  10.979 μs (3 allocations: 688 bytes)
```

Ref: https://discourse.julialang.org/t/forwarddiff-and-gradientconfig-memory-usage/10145/2.
Arguably, StaticArrays should have been used for this size but perhaps this is still worth it.
Same "issue" might exist for derivatives / jacobians etc.

### DIFF
--- a/src/apiutils.jl
+++ b/src/apiutils.jl
@@ -29,13 +29,13 @@ end
 
 @inline static_dual_eval(::Type{T}, f, x::SArray) where {T} = f(dualize(T, x))
 
-function vector_mode_dual_eval(f, x, cfg::Union{JacobianConfig,GradientConfig})
+function vector_mode_dual_eval(f::F, x, cfg::Union{JacobianConfig,GradientConfig}) where {F}
     xdual = cfg.duals
     seed!(xdual, x, cfg.seeds)
     return f(xdual)
 end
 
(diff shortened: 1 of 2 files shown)


simonbyrne:
Wait, specialization is affected by the existence of type parameters in method definitions?

ChrisRackauckas:
>Wait, specialization is affected by the existence of type parameters in method definitions?

Only for functions IIRC. I really don't like this exception...

simonbyrne:
Is it documented anywhere?

ChrisRackauckas:
https://github.com/JuliaLang/julia/issues/19137

Issue #322, opened 2018-05-22
Type promotion of BigFloat or Bool and Dual is ambiguous 


Non-Contradiction:
This is what I get on julia 0.6.2
```
julia> using ForwardDiff

julia> promote_type(Bool, ForwardDiff.Dual{nothing,Float64,5})
ERROR: MethodError: promote_rule(::Type{Bool}, ::Type{ForwardDiff.Dual{nothing,Float64,5}}) is ambiguous. Candidates:
  promote_rule(::Type{R}, ::Type{ForwardDiff.Dual{T,V,N}}) where {R<:Bool, T, V<:Real, N} in ForwardDiff at /Users/lichangcheng/.julia/v0.6/ForwardDiff/src/dual.jl:303
  promote_rule(::Type{R}, ::Type{ForwardDiff.Dual{T,V,N}}) where {R<:Real, T, V<:Real, N} in ForwardDiff at /Users/lichangcheng/.julia/v0.6/ForwardDiff/src/dual.jl:303
  promote_rule(::Type{Bool}, ::Type{T}) where T<:Number in Base at bool.jl:10
Possible fix, define
  promote_rule(::Type{Bool}, ::Type{ForwardDiff.Dual{T,V<:Real,N}})
Stacktrace:
 [1] promote_type(::Type{Bool}, ::Type{ForwardDiff.Dual{nothing,Float64,5}}) at ./promotion.jl:161

julia> promote_type(BigFloat, ForwardDiff.Dual{nothing,Float64,5})
ERROR: MethodError: promote_rule(::Type{BigFloat}, ::Type{ForwardDiff.Dual{nothing,Float64,5}}) is ambiguous. Candidates:
  promote_rule(::Type{R}, ::Type{ForwardDiff.Dual{T,V,N}}) where {R<:BigFloat, T, V<:Real, N} in ForwardDiff at /Users/lichangcheng/.julia/v0.6/ForwardDiff/src/dual.jl:303
  promote_rule(::Type{R}, ::Type{ForwardDiff.Dual{T,V,N}}) where {R<:Real, T, V<:Real, N} in ForwardDiff at /Users/lichangcheng/.julia/v0.6/ForwardDiff/src/dual.jl:303
  promote_rule(::Type{BigFloat}, ::Type{#s268} where #s268<:Real) in Base.MPFR at mpfr.jl:260
Possible fix, define
[...]

PR #323, opened 2018-05-22
Type promotion for Dual with BigFloat and Bool

Non-Contradiction:
Related to #322.


jrevels:
Sorry for the late response, thanks for this!

PR #327, opened 2018-06-28
Make sure that SpecialFunctions are visible in ForwardDiff

andreasnoack:
This was causing the failure in https://github.com/JuliaStats/StatsFuns.jl/issues/49 so a bugfix release after this (or some other fix) has been merged would be appreciated.


jrevels:
This should be fixed in DiffRules, not here (see https://github.com/JuliaDiff/DiffRules.jl/issues/17)

PR #329, opened 2018-07-03
fixed for Julia v0.7

fredrikekre:


fredrikekre:
Great; all green! Some things to note:
1. Some package, e.g. StaticArrays that require Statistics are broken on beta, so I disabled testing on beta for now
2. There is no need to build sysimg to make simd work on 0.7 (🎉) so I removed that on 0.7.

Would be great to get this merged and tagged, to make ForwardDiff pass Keno's new pkgeval!

KristofferC:
Perhaps a few more `using Compat.LinearAlgebra` in the other diff modules to silence some of the deprecation warnings when testing?

fredrikekre:
Those are https://github.com/JuliaDiff/DiffTests.jl/pull/3

.travis.yml
```diff
@@ -1,18 +1,19 @@
...
     - 0.6
+    # - 0.7
     - nightly
-matrix:
-  allow_failures:
-  - julia: nightly
+# matrix:
+#   allow_failures:
+#   - julia: nightly
 notifications:
```
jrevels: It might be better to do:

```
julia:
    - 0.7
    - nightly
matrix:
  allow_failures:
  - julia: nightly
```

I think the next release of ForwardDiff should be v0.7 only. We can backport future fixes to previous minor releases if needed. Trying to maintain a single codebase to support two separate Julia versions isn't worth it IMO.

jrevels:
As mentioned in the other comment, I think we just drop v0.6 support on ForwardDiff master with this commit. Obviously the existing ForwardDiff release can continue to be supported via backports if needed.

fredrikekre:
Okay, I will update the PR and eliminate 0.6 stuff.

.travis.yml
fredrikekre: See comments here: https://github.com/JuliaDiff/DiffTests.jl/pull/3#discussion_r200322905
Arguably `nightly` is more important to test at this stage compared to `- 0.7` which is `0.7-beta`

jrevels: If a contributor makes a PR which passes `0.7` but not `nightly`, I'd merge it regardless, since I don't pass/fail that contributor's work based on unreleased/changing upstream behavior.

Downstream package CI really should not depend on unfixed upstream dependencies. If an upstream tool wants downstream tools to update, upstream can post new tags for the `0.7` target, and downstream can update accordingly.

Note that the allowing failures in the build matrix still runs the tests, so we can still keep our eye on it.

fredrikekre: Guess we have to wait for rc-1 then.

fredrikekre:
Now passes on `nightly` and `0.7-beta2`.

fredrikekre:
bump

jrevels:
Thanks again!

PR #337, opened 2018-07-25
Add constructor calling convert for (::Type{<:Dual})(::Real), fix #336

schmrlng:
From an organizational perspective this extra method could also live at the top with the other constructors; I'm not sure what's preferred.

PR #342, opened 2018-08-08
fix constructor ambiguity introduced by #337

jrevels:
Also deprecates that functionality with a helpful message, because we shouldn't be propagating this weird constructor --> convert pun anyway.

@Keno this should fix your problem with StatsFuns.jl

### DIFF
--- a/src/dual.jl
+++ b/src/dual.jl
@@ -45,8 +45,17 @@ end
 @inline Dual{T}(value::Real, partials::Tuple{}) where {T} = Dual{T}(value, Partials{0,typeof(value)}(partials))
 @inline Dual{T}(value::Real, partials::Real...) where {T} = Dual{T}(value, partials)
 @inline Dual{T}(value::V, ::Chunk{N}, p::Val{i}) where {T,V<:Real,N,i} = Dual{T}(value, single_seed(Partials{N,V}, p))
-
 @inline Dual(args...) = Dual{Nothing}(args...)
+
+function Dual{T,V,N}(x::Real) where {T,V,N}
+    Base.depwarn("Dual{$T,$V,$N}(x::Real) is deprecated, use `convert(Dual{$T,$V,$N}, x)` instead.", :Dual)
+    return convert(Dual{T,V,N}, x)
+end
+
+function Dual{T,V}(x::Real) where {T,V}
+    Base.depwarn("Dual{$T,$V}(x::Real) is deprecated, use `convert(Dual{$T,$V}, x)` instead.", :Dual)
+    return convert(Dual{T,V}, x)
+end
 
 ##############################
 # Utility/Accessor Functions #
@@ -315,7 +324,6 @@ end
 Base.convert(::Type{Dual{T,V,N}}, d::Dual{T}) where {T,V<:Real,N} = Dual{T}(convert(V, value(d)), convert(Partials{N,V}, partials(d)))
 Base.convert(::Type{Dual{T,V,N}}, x::Real) where {T,V<:Real,N} = Dual{T}(convert(V, x), zero(Partials{N,V}))
 Base.convert(::Type{D}, d::D) where {D<:Dual} = d
-(::Type{D})(x::Real) where {D<:Dual} = convert(D, x)
 
 Base.float(d::Dual{T,V,N}) where {T,V,N} = convert(Dual{T,promote_type(V, Float16),N}, d)
 Base.AbstractFloat(d::Dual{T,V,N}) where {T,V,N} = convert(Dual{T,promote_type(V, Float16),N}, d)
--- a/test/DualTest.jl
+++ b/test/DualTest.jl
@@ -54,6 +54,9 @@ for N in (0,3), M in (0,4), V in (Int, Float32)
     ################
 
     @test Dual{TestTag()}(PRIMAL, PARTIALS...) === FDNUM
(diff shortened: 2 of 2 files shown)


tkoolen:
Why not just define these constructors? This breaks e.g.

```julia
julia> using ForwardDiff

julia> A = convert.(Ref(ForwardDiff.Dual), rand(3, 3))
3×3 Array{ForwardDiff.Dual{Nothing,Float64,0},2}:
 Dual{Nothing}(0.859139)  Dual{Nothing}(0.56059)   Dual{Nothing}(0.248962)
 Dual{Nothing}(0.722703)  Dual{Nothing}(0.971363)  Dual{Nothing}(0.437886)
 Dual{Nothing}(0.235157)  Dual{Nothing}(0.101501)  Dual{Nothing}(0.054518)

julia> A \ rand(3)
ERROR: Dual{Nothing,Float64,0}(x::Real) is deprecated, use `convert(Dual{Nothing,Float64,0}, x)` instead.
Stacktrace:
 [1] depwarn(::String, ::Symbol) at ./deprecated.jl:65
 [2] Type at /home/twan/.julia/packages/ForwardDiff/OXtu9/src/dual.jl:51 [inlined]
 [3] oneunit(::Type{ForwardDiff.Dual{Nothing,Float64,0}}) at ./number.jl:321
 [4] lutype(::Type) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.0/LinearAlgebra/src/lu.jl:159
 [5] #lu#103(::Bool, ::Function, ::Array{ForwardDiff.Dual{Nothing,Float64,0},2}) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.0/LinearAlgebra/src/lu.jl:242
 [6] lu at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.0/LinearAlgebra/src/lu.jl:242 [inlined]
 [7] \(::Array{ForwardDiff.Dual{Nothing,Float64,0},2}, ::Array{Float64,1}) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.0/LinearAlgebra/src/generic.jl:870
 [8] top-level scope at none:0
```

(Julia 1.0, `depwarn=error`)

schmrlng:
I was just about to lobby for these methods (with no warning) as well. `convert`/constructor punning appears to be sanctioned (for the limited case of `Number`s) in [base](https://github.com/JuliaLang/julia/blob/f8d82ccba81223c9d0d43e4077acb2a208bedc7a/base/number.jl#L7) and is applied in arguably fundamental methods like [`oneunit`](https://github.com/JuliaLang/julia/blob/f8d82ccba81223c9d0d43e4077acb2a208bedc7a/base/number.jl#L320).

jrevels:
That's just bad style in `Base`. `oneunit` should be definitely be changed to use `convert` now that `convert`/constructor punning is no longer assumed behavior.

JeffBezanson:
`oneunit` is a bit different since unit quantities are not created by conversion. `Dual` is a subtype of Number, and Base assumes that convert and constructors are equivalent for numbers. In fact we define
```
convert(::Type{T}, x::Number) where {T<:Number} = T(x)
```
so that number types can only implement constructors (where before they only implemented `convert`).

jrevels:
> Base assumes that convert and constructors are equivalent for numbers

Why are numbers special in this regard? Is this a good assumption that we should document as part of the actual `<:Number` interface, or is it simply an assumption that's evolved to exist over time?

My point for deprecating these methods is that I'd rather people use the "blessed" constructors ForwardDiff provides for these numbers, since I want to require that people are explicit about the perturbation coefficients.

If they simply want to represent an existing `Dual` as a `Dual` with a slightly different type, well, that's what `convert` is for (IIUC).

JeffBezanson:
> Why are numbers special in this regard?

Because they're immutable, atomic sorts of things. For some kinds of types (including numbers) there just doesn't seem to be a meaningful distinction between constructing and converting. What could `Int(2.0)` possibly do that `convert(Int, 2.0)` wouldn't do, or vice-versa?

jrevels:
> convert(Dual, ::Real) but not Dual(::Real)

ForwardDiff does allow both of these. That's not what's being debated here. 

The debate is whether or not we should allow people to easily "construct" improperly instantiated dual numbers, or otherwise, about whether we should add redundant constructors that are more likely to be used incorrectly than the existing ones.

> What could Int(2.0) possibly do that convert(Int, 2.0) wouldn't do, or vice-versa?

Sure, but that doesn't seem to have much to do with dual numbers, which have quite a different implementation from types like `Int` where the constructor/convert pun really does make sense.

>  For some kinds of types (including numbers) there just doesn't seem to be a meaningful distinction between constructing and converting. 

I agree that for some types, it is true that the constructor/convert pun is totally reasonable. However, this doesn't seem like the kind of property that should be considered to be strictly determined from a subtype relation, since it is heavily dependent on the specific implementation/purpose of the type. In other words, some types of `<:Number` naturally behave this way, while others (like dual numbers) don't seem to IMO.

Anyway, if this was a pure dual numbers implementation, I'd be more aggressive about this. However, the point of ForwardDiff's dual numbers is to do AD on Julia code, so they need to adopt whatever conventions are in Base. If that means following the (undocumented? maybe I just missed this?) Base decision that the convert/construct pun holds for all number types, then we'll follow along, even if the Base decision doesn't seem sensible to me in an abstract sense.

PR #357, opened 2018-09-14
CSE for Base functions

MikeInnes:
ForwardDiff previously always qualified calls like `Base.exp`, but since DiffRules does not qualify base functions in its code, this meant that CSE would not recognise both calls as being identical.


jrevels:
Thanks for catching and fixing this!

PR #368, opened 2018-10-12
remove generated functions on tuples

KristofferC:
Performance seems ok but I havent done big extensive testing:

```
julia> p = rand(ForwardDiff.Partials{8, Float32});

julia> @code_llvm p+p

  %3 = bitcast { [8 x float] } addrspace(11)* %1 to <8 x float> addrspace(11)*
  %4 = load <8 x float>, <8 x float> addrspace(11)* %3, align 4
  %5 = bitcast { [8 x float] } addrspace(11)* %2 to <8 x float> addrspace(11)*
  %6 = load <8 x float>, <8 x float> addrspace(11)* %5, align 4
  %7 = fadd <8 x float> %4, %6
  %8 = bitcast { [8 x float] }* %0 to <8 x float>*
  store <8 x float> %7, <8 x float>* %8, align 4
  ret void
}


julia> @code_llvm ForwardDiff._mul_partials(p, p, 2.f0, 4.f0);

  %5 = bitcast { [8 x float] } addrspace(11)* %1 to <8 x float> addrspace(11)*
  %6 = load <8 x float>, <8 x float> addrspace(11)* %5, align 4
  %7 = insertelement <8 x float> undef, float %3, i32 0
  %8 = shufflevector <8 x float> %7, <8 x float> undef, <8 x i32> zeroinitializer
  %9 = fmul <8 x float> %8, %6

  %10 = bitcast { [8 x float] } addrspace(11)* %2 to <8 x float> addrspace(11)*
  %11 = load <8 x float>, <8 x float> addrspace(11)* %10, align 4
  %12 = insertelement <8 x float> undef, float %4, i32 0
  %13 = shufflevector <8 x float> %12, <8 x float> undef, <8 x i32> zeroinitializer
  %14 = fmul <8 x float> %13, %11

  %15 = fadd <8 x float> %9, %14

  %16 = bitcast { [8 x float] }* %0 to <8 x float>*
  store <8 x float> %15, <8 x float>* %16, align 4
  ret void
}
```

Rosenbrock benchmark

```jl
using ForwardDiff: GradientConfig, Chunk, gradient!

function rosenbrock(x)
    a = one(eltype(x))
    b = 100 * a
    result = zero(eltype(x))
    for i in 1:length(x)-1
    end
    return result
end

# input vector
x = rand(10000);

# output buffer
out = similar(x);

cfg = ForwardDiff.GradientConfig(rosenbrock, x);
```

Before:

```jl
julia> @btime ForwardDiff.gradient!(out, rosenbrock, x, cfg);
  147.659 ms (0 allocations: 0 bytes)
```

After:

```jl
julia> @btime ForwardDiff.gradient!(out, rosenbrock, x, cfg);
  147.660 ms (0 allocations: 0 bytes)
```

### DIFF
--- a/src/partials.jl
+++ b/src/partials.jl
@@ -6,10 +6,7 @@ end
 # Utility/Accessor Functions #
 ##############################
 
-@generated function single_seed(::Type{Partials{N,V}}, ::Val{i}) where {N,V,i}
-    ex = Expr(:tuple, [ifelse(i === j, :(one(V)), :(zero(V))) for j in 1:N]...)
-    return :(Partials($(ex)))
-end
+single_seed(::Type{Partials{N,V}}, ::Val{i}) where {N,V,i} = Partials(ntuple(j -> j == i ? one(V) : zero(V), Val(N)))
 
 @inline valtype(::Partials{N,V}) where {N,V} = V
 @inline valtype(::Type{Partials{N,V}}) where {N,V} = V
@@ -34,18 +31,20 @@ Base.mightalias(x::AbstractArray, y::Partials) = false
 # Generic Functions #
 #####################
 
-@inline Base.iszero(partials::Partials) = iszero_tuple(partials.values)
+@inline Base.iszero(partials::Partials) = all(iszero, partials.values)
 
 @inline Base.zero(partials::Partials) = zero(typeof(partials))
-@inline Base.zero(::Type{Partials{N,V}}) where {N,V} = Partials{N,V}(zero_tuple(NTuple{N,V}))
+@inline Base.zero(::Type{Partials{N,V}}) where {N,V} = Partials{N,V}(ntuple(_ -> zero(V), Val(N)))
 
 @inline Base.one(partials::Partials) = one(typeof(partials))
-@inline Base.one(::Type{Partials{N,V}}) where {N,V} = Partials{N,V}(one_tuple(NTuple{N,V}))
+@inline Base.one(::Type{Partials{N,V}}) where {N,V} = Partials{N,V}(ntuple(_ -> one(V), Val(N)))
 
 @inline Random.rand(partials::Partials) = rand(typeof(partials))
-@inline Random.rand(::Type{Partials{N,V}}) where {N,V} = Partials{N,V}(rand_tuple(NTuple{N,V}))
+@inline Random.rand(::Type{Partials{N,V}}) where {N,V} = Partials{N,V}(ntuple(_ -> rand(V), Val(N)))
+
 @inline Random.rand(rng::AbstractRNG, partials::Partials) = rand(rng, typeof(partials))
-@inline Random.rand(rng::AbstractRNG, ::Type{Partials{N,V}}) where {N,V} = Partials{N,V}(rand_tuple(rng, NTuple{N,V}))
+@inline Random.rand(rng::AbstractRNG, ::Type{Partials{N,V}}) where {N,V} = Partials{N,V}(ntuple(_ -> rand(rng, V), Val(N)))
+
 
 Base.isequal(a::Partials{N}, b::Partials{N}) where {N} = isequal(a.values, b.values)
 Base.:(==)(a::Partials{N}, b::Partials{N}) where {N} = a.values == b.values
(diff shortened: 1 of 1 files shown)


jrevels:
Can we benchmark this change against low dimensional jacobian with StaticArray input/output? IIRC there are cases like this in the tests if you want to compare those.

cc @andyferris According to @KristofferC, StaticArrays abuses generated functions in the same way the `Partials` code here did before this PR. Just wanted to give you a heads up in case you wanted to port these changes so that everybody using StaticArrays can benefit . Obviously it would also be ideal to eventually replace `Partials` with `SVector` entirely, but that's more work in the short term 😛

KristofferC:
Well, I didn't say StaticArrays *abused*  generated functions, just that changing to StaticArrays won't lead to a reduction in generated function use. For example, plus of two `StaticVector` finally dispatches to https://github.com/JuliaArrays/StaticArrays.jl/blob/f445984f316a92ea98c4220693310b6860a4198c/src/mapreduce.jl#L20-L32.

KristofferC:
```jl
julia> using StaticArrays, BenchmarkTools

julia> x = rand(3, 3);

julia> sx = StaticArrays.SArray{Tuple{3,3}}(x);

julia> _diff(A) = diff(A; dims=1);
```

Master

```jl
julia> @btime ForwardDiff.jacobian(_diff, x)
  4.514 μs (6 allocations: 2.61 KiB)
```

PR

```jl
julia> @btime ForwardDiff.jacobian(_diff, x)
  4.628 μs (6 allocations: 2.61 KiB)
```

KristofferC:
The background is basically the following comment in #autodiff slack by @keno   

> Also all those generated functions (e.g. add_tuples) don’t actually help these days and in fact can pessimize inference

andyferris:
Thanks!

@Keno It would be helpful to me if I could understand this better. Is there an inference-friendly way of constructing a length-N tuple without using any generated functions or explicit O(N) code? (I mightn't count recursion, the other option I'm aware of, as "friendly" since AFAICT that approach seems to be O(N^2) effort for a compiler). For example, what can go wrong with `add_tuples`?

jrevels:
> If the goal is to remove usage of generated functions, as far as I can tell, this still relies on the generated function ntuple.

>  Is there an inference-friendly way of constructing a length-N tuple without using any generated functions or explicit O(N) code? 

I'm not sure it's been discussed on GitHub, but we've talked about making `ntuple` a fast builtin in the future (i.e. add a manual tfunc for it in inference rather than having to infer through all the recursive tuple machinery). 

OTOH, tuple/_apply-related inference/optimization/codegen has been rapidly improving in the post-1.0 world, so maybe making it a builtin won't be necessary. This whole situation may also improve if we get something like https://github.com/JuliaLang/julia/issues/29368, since it would alleviate a lot of the current pain of dealing with covariant structures in inference.

KristofferC:
Using

```
for c in (4,8,12,16,20,24,28,32,36)
   cfg = GradientConfig(rosenbrock, x, Chunk{c}());
   @btime gradient!(out, rosenbrock, x, $cfg);
end
```

I get on master

```

196.285 ms (0 allocations: 0 bytes)
128.814 ms (0 allocations: 0 bytes)
122.965 ms (0 allocations: 0 bytes)
112.407 ms (0 allocations: 0 bytes)
100.913 ms (0 allocations: 0 bytes)
88.715 ms (0 allocations: 0 bytes)
87.217 ms (0 allocations: 0 bytes)
84.240 ms (0 allocations: 0 bytes)
81.291 ms (0 allocations: 0 bytes)
```

and with this PR

```
195.447 ms (0 allocations: 0 bytes)
129.538 ms (0 allocations: 0 bytes)
124.539 ms (0 allocations: 0 bytes)
111.933 ms (0 allocations: 0 bytes)
100.593 ms (0 allocations: 0 bytes)
90.064 ms (0 allocations: 0 bytes)
87.663 ms (0 allocations: 0 bytes)
82.538 ms (0 allocations: 0 bytes)
78.922 ms (0 allocations: 0 bytes)
```

It was surprising to me how large chunk size is beneficial. But the function is quite similar so likely doesn't require much stack space which might be relevant?

KristofferC:
To measure compile time, here is the result with `@btime` changed to `@time`

Master:

```
 0.673772 seconds (1.13 M allocations: 56.933 MiB, 1.59% gc time)
  0.327096 seconds (456.96 k allocations: 23.293 MiB, 1.43% gc time)
  0.342328 seconds (511.11 k allocations: 25.692 MiB, 1.48% gc time)
  0.358762 seconds (565.44 k allocations: 28.132 MiB, 1.46% gc time)
  0.360280 seconds (620.05 k allocations: 30.575 MiB, 2.89% gc time)
  0.366458 seconds (675.06 k allocations: 33.041 MiB, 2.58% gc time)
  0.370327 seconds (730.51 k allocations: 35.499 MiB, 1.37% gc time)
  0.364240 seconds (788.03 k allocations: 38.006 MiB, 2.47% gc time)
  0.428677 seconds (852.49 k allocations: 40.798 MiB, 2.26% gc time)
```

PR:

```
  0.450079 seconds (859.98 k allocations: 45.502 MiB, 2.32% gc time)
  0.433871 seconds (1.10 M allocations: 57.474 MiB, 3.62% gc time)
  0.489985 seconds (1.37 M allocations: 70.898 MiB, 3.79% gc time)
  0.576018 seconds (1.65 M allocations: 84.386 MiB, 13.10% gc time)
  0.553180 seconds (1.92 M allocations: 97.866 MiB, 5.21% gc time)
  0.674659 seconds (2.19 M allocations: 111.465 MiB, 14.04% gc time)
  0.639073 seconds (2.47 M allocations: 124.858 MiB, 5.78% gc time)
  0.699460 seconds (2.75 M allocations: 138.962 MiB, 6.08% gc time)
  0.830345 seconds (3.07 M allocations: 153.975 MiB, 10.98% gc time)
```

So compilation time does unfortunately seem worse.

KristofferC:
I don't think we will get away from this being worse for compile-time so I will close this.

PR #369, opened 2018-10-15
Remove V <: Real type restriction

### LINKED ISSUE #216: Remove need for valtype of Dual to be <: Real
It would be useful to remove the need for the eltype of `Dual.value` to be `<:Real`.
Cf. the discussion at https://github.com/JuliaIntervals/IntervalArithmetic.jl/issues/2

Keno:
Instead, use an extensible function that the constructor uses to
check whether a type is valid to be used as `Dual`'s scalar type.

Fixes #216

### DIFF
--- a/src/dual.jl
+++ b/src/dual.jl
@@ -2,9 +2,22 @@
 # Dual #
 ########
 
-struct Dual{T,V<:Real,N} <: Real
+"""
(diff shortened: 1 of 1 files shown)


Keno:
@jrevels Good to go?

PR #374, opened 2018-10-30
Partials do not alias

KristofferC:
Benchmark

```jl
function rosenbrock(x)
   a = one(eltype(x))
   b = 100 * a
   result = zero(eltype(x))
   for i in 1:length(x)-1
   end
   return result
end

using BenchmarkTools
import ForwardDiff
const x = rand(10)
const y = similar(x)
const gradient_cache = ForwardDiff.GradientConfig(rosenbrock, x)
@btime ForwardDiff.gradient!(y, rosenbrock, x, gradient_cache)
```

Before

```jl
julia> @btime ForwardDiff.gradient!(y, rosenbrock, x, gradient_cache)
  247.428 ns (1 allocation: 96 bytes)
```

After

```jl
julia> @btime ForwardDiff.gradient!(y, rosenbrock, x, gradient_cache)
  178.401 ns (0 allocations: 0 bytes)
```

### DIFF
--- a/src/partials.jl
+++ b/src/partials.jl
@@ -2,6 +2,8 @@ struct Partials{N,V} <: AbstractVector{V}
     values::NTuple{N,V}
 end
 
+Base.mightalias(x::AbstractArray, y::Partials) = false
+
 ##############################
 # Utility/Accessor Functions #
 ##############################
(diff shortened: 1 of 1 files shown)


jrevels:
Huh, TIL about `mightalias`.

PR #376, opened 2018-11-05
bump DEFAULT_CHUNK_THRESHOLD to 12

KristofferC:
We now have that lovely SIMD by default so should probably think about that a bit more w.r.t the `Chunk` we choose.

Doing a benchmark such as

```jl
using ForwardDiff
using BenchmarkTools

import ForwardDiff.Dual
f(x) = (exp(x) * sin(x) + x^2 / tan(x))^3 / x^2
d_10 = Dual(rand(11)...)
d_12 = Dual(rand(13)...)

@btime f($d_10)
@btime f($d_12)
```

We find

```jl
julia> @btime f($d_10);
  77.918 ns (0 allocations: 0 bytes)

julia> @btime f($d_12);
  69.597 ns (0 allocations: 0 bytes)
```

so it is, in fact, faster to compute with chunk sizes of 12 instead of 10 (likely because they then all nicely fit into AVX registers and no need to clean up the scalars)

To see the effect on our favorite toy problem:

```jl
function rosenbrock(x)
   a = one(eltype(x))
   b = 100 * a
   result = zero(eltype(x))
   for i in 1:length(x)-1
   end
   return result
end

using BenchmarkTools
import ForwardDiff
const x = rand(60)
const y = similar(x)
const gradient_cache = ForwardDiff.GradientConfig(rosenbrock, x)
```

Before:

```jl
julia> @btime ForwardDiff.gradient!(y, rosenbrock, x, gradient_cache)
  5.589 μs (0 allocations: 0 bytes)
```

After:
```jl
julia> @btime ForwardDiff.gradient!(y, rosenbrock, x, gradient_cache)
  4.761 μs (0 allocations: 0 bytes)
```

### DIFF
--- a/src/prelude.jl
+++ b/src/prelude.jl
@@ -6,7 +6,7 @@ const UNARY_PREDICATES = Symbol[:isinf, :isnan, :isfinite, :iseven, :isodd, :isr
 
 const BINARY_PREDICATES = Symbol[:isequal, :isless, :<, :>, :(==), :(!=), :(<=), :(>=)]
 
-const DEFAULT_CHUNK_THRESHOLD = 10
+const DEFAULT_CHUNK_THRESHOLD = 12
 
 struct Chunk{N} end
 
--- a/test/JacobianTest.jl
+++ b/test/JacobianTest.jl
@@ -101,7 +101,7 @@ cfgx = ForwardDiff.JacobianConfig(sin, x)
 for f in DiffTests.ARRAY_TO_ARRAY_FUNCS
     v = f(X)
     j = ForwardDiff.jacobian(f, X)
-    @test isapprox(j, Calculus.jacobian(x -> vec(f(x)), X, :forward), atol=FINITEDIFF_ERROR)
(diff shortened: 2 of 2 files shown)


KristofferC:
Test failures due to some arbitrary chosen `FINITEDIFF_ERROR`. Yay...

jrevels:
> Test failures due to some arbitrary chosen FINITEDIFF_ERROR. Yay...

Ugh, that's annoying. I really hate the finite differences approximation death cycle (want automated testing for AD -> need machine precision derivatives -> need AD -> want automated testing for AD...). More simple analytic unit tests would also be nice...

Bump the tolerance for those cases I guess? There are also methods for calculating finite difference approximation error bounds for a given function/input, it would be nice to incorporate those in the future (which ideally should be provided by the same package providing the finite diff implementation).

KristofferC:
Bumped the tolerance a bit.

jrevels:
LGTM, thanks!

PR #378, opened 2019-01-10
Add the other rng functions

ChrisRackauckas:


jrevels:
Actually, just thought of something; the `rand(::Dual)` method probably should be overloaded, these specific method definitions just happen to be wrong. IIUC, they should just be:

```julia
Random.rand(d::Dual) = rand(value(d))
Random.rand(rng::AbstractRNG, d::Dual) = rand(rng, value(d))
```

The reason is that the value wrapped in `Dual` might be a dimension length, e.g. calling `rand(Dual(10))` should just call `rand(10)`. 

Of course, there are probably many constructors in Base that have `Integer` restrictions we'd have to overload (or loosen) to comprehensively support this kind of thing, but actually going after that is a whole other ball game...

ChrisRackauckas:
Added back.

MikeInnes:
That's all correct but just FYI, the `rand(d::Dual)` method is redundant with `rand(rng, d)`, because there's a fallback method that uses the default RNG. In all cases you can get away with just defining the `rand(rng::AbstractRNG, ...)` signature.

jrevels:
LGTM, just requires a fix to [this failing test](https://travis-ci.org/JuliaDiff/ForwardDiff.jl/jobs/479516959#L588) (that was written incorrectly in the first place to test the old behavior)

ChrisRackauckas:
I changed the rng tests to match the behavior assumptions from our discussion. Essentially, when it's a dual integer it should create an array of floats just like if it was an integer, essentially a pass through of a control variable that was accidentally dual. If it's a float type, then it should act like the other float types and create a dual of the similar type but with zero derivative.

test/DualTest.jl
```diff
@@ -170,21 +170,28 @@ for N in (0,3), M in (0,4), V in (Int, Float32)
...
+        @test randexp(samerng(), typeof(FDNUM)) === Dual{TestTag()}(randexp(samerng(), V), zero(Partials{N,V}))
+        @test randexp(samerng(), typeof(NESTED_FDNUM)) === Dual{TestTag()}(Dual{TestTag()}(randexp(samerng(), V), zero(Partials{M,V})),
+        zero(Partials{N,Dual{TestTag(),V,M}}))
+    end
 
     # Predicates #
     #------------#
 
     @test ForwardDiff.isconstant(zero(FDNUM))
-    @test ForwardDiff.isconstant(rand(FDNUM))
```
ChrisRackauckas: For Integers this already tests that it's returning a real Array{Float64}, so this is unnecessary. For floating point numbers we've explicitly written in the test that it should have zero partials, so these tests were redundant on floats but failed on integers (due to no method for Array{Float64}), hence deleted.

jrevels:
Thanks!

PR #379, opened 2019-01-12
Add `seed_duals` convenience and document how to seed Dual numbers

ChrisRackauckas:
Using the Dual implementation in ForwardDiff can be quite useful. This documents how to do it, and adds a seeding function `seed_duals` which seeds a unique partial for every component of a vector, much like is done internally (in fact, should it be used internally?).

### DIFF
--- a/src/dual.jl
+++ b/src/dual.jl
@@ -73,6 +73,16 @@ end
 @inline Dual{T,V,N}(x::Number) where {T,V,N} = convert(Dual{T,V,N}, x)
 @inline Dual{T,V}(x) where {T,V} = convert(Dual{T,V}, x)
 
+############################
+# Convenience Constructors #
+############################
+
+function seed_duals(x::AbstractArray{V},::Type{T},
+                    ::Chunk{N} = Chunk(x)) where {V,T,N}
+  seeds = construct_seeds(Partials{N,V})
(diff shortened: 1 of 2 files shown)


KristofferC:
FWIW, this increases the API surface of ForwardDiff considerably. Dual numbers are not even mentioned in the manual (except in the implementation details).

pkofod:
> FWIW, this increases the API surface of ForwardDiff considerably. Dual numbers are not even mentioned in the manual (except in the implementation details).

This *is* in the advanced section of the docs though.

src/dual.jl
```diff
@@ -73,6 +73,16 @@ end
...
 @inline Dual{T,V}(x) where {T,V} = convert(Dual{T,V}, x)
 
+############################
+# Convenience Constructors #
+############################
+
+function seed_duals(x::AbstractArray{V},::Type{T},
+                    ::Chunk{N} = Chunk(x)) where {V,T,N}
+  seeds = construct_seeds(Partials{N,V})
+  duals = [Dual{T}(x[i],seeds[i]) for i in 1:length(x)]
```
hyrodium, line 83:
```suggestion
    seeds = construct_seeds(Partials{N,V})
    duals = [Dual{T}(x[i],seeds[i]) for i in eachindex(x)]
```

* Fix indents
* Use `eachindex(x)` instead of `1:length(x)`

PR #381, opened 2019-01-25
Relax type signatures on multiplication with partials

staticfloat:
We need this for TPU work so that things like `XRTArrays` can be multiplied in.

### DIFF
--- a/src/partials.jl
+++ b/src/partials.jl
@@ -81,7 +81,6 @@ Base.convert(::Type{Partials{N,V}}, partials::Partials{N,V}) where {N,V} = parti
 @inline Base.:+(a::Partials{N}, b::Partials{N}) where {N} = Partials(add_tuples(a.values, b.values))
 @inline Base.:-(a::Partials{N}, b::Partials{N}) where {N} = Partials(sub_tuples(a.values, b.values))
 @inline Base.:-(partials::Partials) = Partials(minus_tuple(partials.values))
-@inline Base.:*(x::Real, partials::Partials) = partials*x
 @inline Base.:*(x, partials::Partials) = partials*x
 
 @inline function _div_partials(a::Partials, b::Partials, aval, bval)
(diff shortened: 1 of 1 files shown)


src/partials.jl
```diff
@@ -82,6 +82,7 @@ Base.convert(::Type{Partials{N,V}}, partials::Partials{N,V}) where {N,V} = parti
 @inline Base.:-(a::Partials{N}, b::Partials{N}) where {N} = Partials(sub_tuples(a.values, b.values))
 @inline Base.:-(partials::Partials) = Partials(minus_tuple(partials.values))
 @inline Base.:*(x::Real, partials::Partials) = partials*x
```
KristofferC: Can be deleted now?

src/partials.jl
```diff
@@ -110,6 +111,9 @@ else
     @inline function Base.:*(partials::Partials, x::Real)
```
KristofferC: Same here?

src/partials.jl
staticfloat: I think you're right.  @Keno can you confirm that you weren't leaving these specializations on `::Real` for some obscure reason?

jrevels:
We need some versions of these methods with some type restrictions to avoid inevitable ambiguity warnings:

```
  ...testing Partials{3,Int64}
Error During Test at /home/travis/build/JuliaDiff/ForwardDiff.jl/test/PartialsTest.jl:111
  Test threw exception
  Expression: X * PARTIALS == PARTIALS * X
  MethodError: *(::Float64, ::ForwardDiff.Partials{3,Int64}) is ambiguous. Candidates:
    *(x, partials::ForwardDiff.Partials) in ForwardDiff at /home/travis/build/JuliaDiff/ForwardDiff.jl/src/partials.jl:84
    *(A::Number, B::AbstractArray) in Base at arraymath.jl:52
  Possible fix, define
    *(::Number, ::ForwardDiff.Partials)
```

Would be nice to just replace `Partials` with StaticVectors entirely sometime soon since we've already taken on StaticArrays as a dependency

PR #392, opened 2019-02-22
Integer overload to fix Duals in ranges

ChrisRackauckas:
MWE fixed:

```julia
using ForwardDiff: Dual
t = Dual.((0.0,10.0),0)
(t[1]:1/60:t[2])
```

### DIFF
--- a/src/dual.jl
+++ b/src/dual.jl
@@ -86,7 +86,7 @@ end
     if S ≺ T
         d
     else
-        throw(DualMismatchError(T,S))        
+        throw(DualMismatchError(T,S))
     end
 end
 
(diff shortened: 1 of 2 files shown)


MikeInnes:
The `Integer` conversion most likely isn't already enabled because it would drop gradients. Seems like it'd be more correct to do return a range of duals in this case.

KristofferC:
This seems to work now:

```

julia> t = Dual.((0.0,10.0),0)
(Dual{Nothing}(0.0,0.0), Dual{Nothing}(10.0,0.0))

julia> (t[1]:1/60:t[2])
Dual{Nothing}(0.0,0.0):Dual{Nothing}(0.016666666666666666,0.0):Dual{Nothing}(10.0,0.0)
```

PR #406, opened 2019-07-12
WIP: GPU-compatibility of Jacobians

ChrisRackauckas:
MWE:

```julia
using ForwardDiff
A = cu(-rand(3,3))
u = cu([1.0;0.0;0.0])
du = similar(u)
f(du,u) = mul!(du,A,u)
ForwardDiff.jacobian(f,du,u)
```

### DIFF
--- a/src/apiutils.jl
+++ b/src/apiutils.jl
@@ -55,36 +55,31 @@ end
 
 function seed!(duals::AbstractArray{Dual{T,V,N}}, x,
                seed::Partials{N,V} = zero(Partials{N,V})) where {T,V,N}
-    for i in eachindex(duals)
-        duals[i] = Dual{T,V,N}(x[i], seed)
-    end
+    duals .= Dual{T,V,N}.(x, (seed,))
     return duals
 end
 
 function seed!(duals::AbstractArray{Dual{T,V,N}}, x,
                seeds::NTuple{N,Partials{N,V}}) where {T,V,N}
-    for i in 1:N
-        duals[i] = Dual{T,V,N}(x[i], seeds[i])
-    end
+    i = 1:N
+    @views duals[i] .= Dual{T,V,N}.(x[i], seeds)
     return duals
 end
 
 function seed!(duals::AbstractArray{Dual{T,V,N}}, x, index,
                seed::Partials{N,V} = zero(Partials{N,V})) where {T,V,N}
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
@vchuravy any idea why this would fallback to the Base broadcasting?

```julia
using ForwardDiff, CuArrays
x = CuArray([ForwardDiff.Dual(1.0f0,(0.0f0,0.0f0)) for i in 1:10])
CuArrays.allowscalar(false)
ForwardDiff.partials.(Float32,x,1)
```

KristofferC:
https://github.com/JuliaDiff/ForwardDiff.jl/pull/351

The reason that PR got reverted needs to be checked here.

ChrisRackauckas:
I see. As a continuation of https://github.com/JuliaDiff/ForwardDiff.jl/pull/353 @maleadt . The routes I'm thinking of are:

1) Requires.jl, add overloads for CuArrays. Would you accept that?
2) We may just add overloads for this in Requires of DiffEqBase. DiffEqBase has a faster `@..` broadcast which gets rid of the timing issues of changing to broadcast (Julia v1.0 broadcast is quite slow :( ), and I'm not sure we could have no performance loss without using this.
3) We may just make sure the color-autodiff form of this into SparseDiffTools.jl.

Of course, not having it in ForwardDiff.jl is a hassle, and having the piracy version in DiffEqBase is a little awkward, but my main priority is to get it working for DiffEq and then find out where to put the code later.

src/apiutils.jl
```diff
@@ -55,36 +55,31 @@ end
...
     return duals
 end
 
 function seed!(duals::AbstractArray{Dual{T,V,N}}, x,
                seeds::NTuple{N,Partials{N,V}}) where {T,V,N}
-    for i in 1:N
-        duals[i] = Dual{T,V,N}(x[i], seeds[i])
-    end
+    i = 1:N
+    @views duals[i] .= Dual{T,V,N}.(x[i], seeds)
```
KristofferC, line 65: Perhaps these `@views` could be reduced in "strength" by putting them where they are needed, e.g. only on the `x[i]`? And could the usage of `@.` be consistent among the changes (it's not used here, but used in other places).

charleskawczynski:
I think we can close this now that #472 has merged.

PR #408, opened 2019-08-27
also test on 1.2

KristofferC:


KristofferC:
`  Expression: dual_isapprox(FDNUM ^ PRIMAL, exp(PRIMAL * log(FDNUM)))`

seems to fail a bit randomly on 1.3 and x86 1.2.

PR #414, opened 2019-10-14
Fix x^y for the case when x is zero.

andreasnoack:
Currently, the derivative in the exponent is wrong then the base is zero, i.e. it gives
```julia
julia> (x -> ForwardDiff.derivative(y -> x^y,  0.5))(0.0)
NaN
```
even though the limit is well defined. With this PR, it becomes
```julia
julia> (x -> ForwardDiff.derivative(y -> x^y,  0.5))(0.0)
0.0
```

### DIFF
--- a/src/dual.jl
+++ b/src/dual.jl
@@ -442,7 +442,13 @@ for f in (:(Base.:^), :(NaNMath.pow))
                 vx, vy = value(x), value(y)
                 expv = ($f)(vx, vy)
                 powval = vy * ($f)(vx, vy - 1)
-                logval = isconstant(y) ? one(expv) : expv * log(vx)
+                if isconstant(y)
+                    logval = one(expv)
+                elseif iszero(vx) && vy > 0
+                    logval = zero(vx)
+                else
+                    logval = expv * log(vx)
+                end
                 new_partials = _mul_partials(partials(x), partials(y), powval, logval)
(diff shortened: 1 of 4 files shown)


andreasnoack:
Hm. I can't reproduce the Windows error. Tried both Julia 1.0 and 1.2 and tests pass.

KristofferC:
Yeah, has happened for quite a long itme.

andreasnoack:
Bump

src/dual.jl
```diff
@@ -302,7 +302,7 @@ end
 # Predicates #
 #------------#
 
-isconstant(d::Dual) = iszero(partials(d))
+isconstant(d::Dual) = Base.iszero(partials(d))
```
ararslan:
```suggestion
isconstant(d::Dual) = iszero(partials(d))
```
The qualification from Base shouldn't actually be necessary since you're changing the `iszero` definition here to extend the one in Base. Or am I missing something?

andreasnoack: Indeed. It no longer needs the qualification. I've changed it.

andreasnoack:
I'd appreciate a release with this fix.

KristofferC:
Yeah but someone needs to add a Project file.

andreasnoack:
I see. I can do that.

PR #415, opened 2019-10-28
From REQUIRE to Project.toml

andreasnoack:


andreasnoack:
I also bumped the version here so we can use this commit for the release

Project.toml
```diff
@@ -0,0 +1,33 @@
+name = "ForwardDiff"
+uuid = "f6369f11-7733-5829-9624-2563aa707210"
+version = "0.11.0"
```
fredrikekre:
```suggestion
version = "0.10.1"
```

KristofferC: `0.10.4`, or?

fredrikekre: Yea.

KristofferC:
I'll do some maintenance of the way the docs are built and then tag a new version.

Issue #324, opened 2018-06-11
StackOverflowError with complex `sqrt`


FMeirinhos:
When differentiating over a complex sqrt

`x0 = 5`
`ForwardDiff.derivative(x -> sqrt((1 + 1im) * x), x0)`

a StackOverflowError is encountered.

I have found similar issues #111 #196 #239, but I don't see how I can fix this problem.

tkoolen:
This is basically https://github.com/JuliaLang/julia/issues/26552 I believe.

EdoAlvarezR:
I'm running into the same issue when trying to have ForwardDiff calculating the derivative over a complex sqrt operation. Here I was able to reproduce the error in a simplified way:

```julia
> x = ForwardDiff.Dual{Nothing}(1,0.0,0.0,0.0)*im
> sqrt(x)
StackOverflowError:

Stacktrace:
 [1] Type at ./complex.jl:12 [inlined] (repeats 2 times)
 [2] float at ./complex.jl:967 [inlined]
 [3] sqrt(::Complex{ForwardDiff.Dual{Nothing,Float64,3}}) at ./complex.jl:461 (repeats 80000 times)
```

Has anyone found a solution for this?

chriscoey:
I too am running into this

chriscoey:
I too am running into this

adam-coogan:
This is a problem for me as well. Any ideas about how to deal with this?

jfmoulin:
I run into same error with a just  updated Turing/ForwardDiff.

andreasnoack:
See the discussion in https://github.com/JuliaDiff/ForwardDiff.jl/issues/157

baggepinnen:
I am trying this out in julia v1.6 and it's sometimes working okay now. `sqrt(::Complex)` contains a call to `ssqs` which has a branch that sometimes fails on `Dual`: the problematic function is `Base.exponent` which is not defined for `Dual`
```julia
 function Base.ssqs(x::T, y::T) where T <: ForwardDiff.Dual
           k::Int = 0
           ρ = x*x + y*y
           @show x
           if !isfinite(ρ) && (isinf(x) || isinf(y))
               ρ = convert(T, Inf)
           elseif isinf(ρ) || (ρ==0 && (x!=0 || y!=0)) || ρ<nextfloat(zero(T))/(2*eps(T)^2)
               m::T = max(abs(x), abs(y))
               @show m
[...]

PR #419, opened 2019-11-07
Remove bogus AbstractFloat conversion.

tpapp:
Fixes
```julia

julia> AbstractFloat(Dual(1.0, 2.0)) isa AbstractFloat
false
```

### DIFF
--- a/src/dual.jl
+++ b/src/dual.jl
@@ -86,7 +86,7 @@ end
     if S ≺ T
         d
     else
-        throw(DualMismatchError(T,S))        
+        throw(DualMismatchError(T,S))
     end
 end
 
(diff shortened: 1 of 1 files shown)


tpapp:
FWIW, I think the way forward for the host of stack overflow issues (https://github.com/JuliaMath/SpecialFunctions.jl/issues/186, #324, etc) is to  remove the `float(::Dual)` method as well.

`Dual` is not `<: AbstractFloat`, so that method violates its contract.

KristofferC:
FWIW, was added in https://github.com/JuliaDiff/ForwardDiff.jl/pull/213/commits/2e8d016d534b9c482667fd9514fd8c00a4e89aa0.

tpapp:
Friendly bump: I am happy to work on this if anything else is required.

sethaxen:
Another friendly bump. What needs to be done to get this PR merged?

nalimilan:
> FWIW, I think the way forward for the host of stack overflow issues ([JuliaMath/SpecialFunctions.jl#186](https://github.com/JuliaMath/SpecialFunctions.jl/issues/186), #324, etc) is to remove the `float(::Dual)` method as well.
> 
> `Dual` is not `<: AbstractFloat`, so that method violates its contract.

Note that `float(::Complex)` also returns a `Complex`, so the contract of that function isn't to return an `AbstractFloat`. I can't comment on what's the best choice for `Dual`, though.

andreasnoack:
See https://github.com/JuliaMath/SpecialFunctions.jl/issues/186#issuecomment-550958244. The issue with `loggamma` was fixed a while ago but it keeps showing up every time SpecialFunctions make a new breaking release because older versions of DiffRules didn't have appropriate compat bounds.

oxinabox (approved):

PR #425, opened 2019-11-27
Allow SpecialFunctions@0.9

fredrikekre:
SpecialFunctions keep releasing "breaking" releases so...


ararslan:
FWIW, 0.9 was tagged instead of 0.8.x to comply with the "no Julia version dropping in a patch release" rule, since it's now 1.3-only.

fredrikekre:
Yea, its suboptimal when you only give yourself room to play with 2 numbers instead of 3 :)

PR #430, opened 2019-12-16
Bump for DiffRules 1.0 compat

ChrisRackauckas:
Should this just be to 1.0?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ForwardDiff"
 uuid = "f6369f11-7733-5829-9624-2563aa707210"
-version = "0.11.0"
+version = "0.10.8"
 
 [deps]
 CommonSubexpressions = "bbf7d656-a473-5ed7-a52c-81e309532950"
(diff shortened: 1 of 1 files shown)


Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "ForwardDiff"
 uuid = "f6369f11-7733-5829-9624-2563aa707210"
-version = "0.10.7"
+version = "0.11.0"
```
fredrikekre:
```suggestion
version = "0.10.8"
```
?

ChrisRackauckas: but it had a compatibility change

fredrikekre: Yes? How is that breaking?

ChrisRackauckas: Doesn't General require a 0.x.0 bump with compatibility changes?

fredrikekre: No.

ChrisRackauckas: Alrighty then.

PR #433, opened 2019-12-24
That is pure, not `@pure`

oxinabox:
From: https://github.com/YingboMa/ForwardDiff2.jl/pull/11#discussion_r361219444

> Are you sure we need this @pure.
This is a illegal usage as it calls generic functions.
And its not a particularly a safe usage illegal usage at that.
In that if someone wants to add support for V1 or V2, being their fancy new numeric type (say ArbFloat).
by overloading promote_type(::V, ::Dual) then bad things can happen.

### DIFF
--- a/src/dual.jl
+++ b/src/dual.jl
@@ -323,8 +323,8 @@ end
 # Promotion/Conversion #
 ########################
 
-Base.@pure function Base.promote_rule(::Type{Dual{T1,V1,N1}},
-                                      ::Type{Dual{T2,V2,N2}}) where {T1,V1,N1,T2,V2,N2}
+function Base.promote_rule(::Type{Dual{T1,V1,N1}},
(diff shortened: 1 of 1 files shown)


ararslan (approved):

oxinabox:
This breaks this test;
https://github.com/JuliaDiff/ForwardDiff.jl/blob/4ec6ce198d15bf783bf4efbf57d976441b76d64e/test/SIMDTest.jl#L68

we don't get `r"(.*fmul \<4 x double\>){2}"`
we do get `r"(.*fmul \<8 x double\>)"`
but I am not sure that that is doing the particular SIMD we are after.

KristofferC:
That might be broken in master as well?

oxinabox:
> That might be broken in master as well?

Correct

KristofferC:
https://github.com/JuliaDiff/ForwardDiff.jl/pull/509

PR #437, opened 2020-01-20
Restrict `getindex` for partials to a single Int

timholy:
Fixes a display bug:

```julia
julia> p = ForwardDiff.Partials((1,2,3))
3-element ForwardDiff.Partials{3,Int64}:
Error showing value of type ForwardDiff.Partials{3,Int64}:
ERROR: MethodError: no method matching 
getindex(::Tuple{Int64,Int64,Int64}, ::Int64, ::Int64)
```


timholy:
https://discourse.julialang.org/t/forwarddiff-how-to-get-the-derivative-part-of-a-dual-number/33560

PR #438, opened 2020-01-20
Move SIMD tests to a separate job and allow them to fail

andreasnoack:
to avoid overall test failure when they fail. This is needed because they
seem to be brittle.

I'll open an issue about making the tests more reliable but we risk merging something breaking (correctness and not just performance) as long as we fail overall when the SIMD tests fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: julia
+
 julia:
     - 1.0
     - 1.2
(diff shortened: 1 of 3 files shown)


KristofferC:
:+1, but doesn't work on AV. Not sure how much it is worth testing on Windows though since all of this is pure Julia code.

andreasnoack:
Ha. The test failure on Windows is actually a proper error in the test from https://github.com/JuliaDiff/ForwardDiff.jl/pull/437 on 32 bit that I didn't notice because of all the SIMD errors.

PR #449, opened 2020-03-05
Document multidimensional arrays

thisrod:


ChrisRackauckas:
Can you fix the line lengths? I think that's all that's needed and this is an easy merge.

thisrod:
@ChrisRackauckas done

Issue #265, opened 2017-09-23
Changes in v0.6 release


mlubin:
It's not described in NEWS, have there been any API changes? Is it safe to update?

PR #451, opened 2020-04-06
NaN-Safe Branch

ChrisRackauckas:
This is a branch with NaN-safe mode enabled to make it easier for folks to get this turned on. We can just keep rebasing as necessary. It's good for helping people make MWEs who don't necessarily know a bunch of dev stuff.

### DIFF
--- a/src/prelude.jl
+++ b/src/prelude.jl
@@ -1,4 +1,4 @@
-const NANSAFE_MODE_ENABLED = false
+const NANSAFE_MODE_ENABLED = true
 
 const AMBIGUOUS_TYPES = (AbstractFloat, Irrational, Integer, Rational, Real, RoundingMode)
 


asinghvi17:
Would it make sense to read an ENV variable at precompile time instead?

KristofferC:
Another possibility is just to exploit the Julia `#265` fix and have something like

```
nan_safe() = false
nan_safe(v::Bool) = @eval nan_safe() = $v

@inline function Base.:*(partials::Partials, x::Real)
    x = nansafe() ? x : ifelse(!isfinite(x) && iszero(partials), one(x), x)
    return Partials(scale_tuple(partials.values, x))
end
```

and you can just toggle `nan_safe` e.g. in the REPL.

andreasnoack:
I'm in favor of making this the default while working on an easier way to opt out. I've never understood why it's okay to have a wrong default here and it's really frustrating to hit these cases.

PR #453, opened 2020-05-06
Add prevfloat and nextfloat for Dual

johnmyleswhite:
Add the prevfloat and nextfloat methods for Dual numbers

### DIFF
--- a/src/dual.jl
+++ b/src/dual.jl
@@ -257,6 +257,14 @@ Base.copy(d::Dual) = d
 Base.eps(d::Dual) = eps(value(d))
 Base.eps(::Type{D}) where {D<:Dual} = eps(valtype(D))
 
+function Base.nextfloat(d::ForwardDiff.Dual{T,V,N}) where {T,V,N}
+    ForwardDiff.Dual{T}(nextfloat(d.value), d.partials)
+end
+
+function Base.prevfloat(d::ForwardDiff.Dual{T,V,N}) where {T,V,N}
+    ForwardDiff.Dual{T}(prevfloat(d.value), d.partials)
+end
+
 Base.rtoldefault(::Type{D}) where {D<:Dual} = Base.rtoldefault(valtype(D))
 
 Base.floor(::Type{R}, d::Dual) where {R<:Real} = floor(R, value(d))


ChrisRackauckas:
I think it should retain dualness a la https://github.com/SciML/DiffEqBase.jl/blob/master/src/init.jl#L61-L63

johnmyleswhite:
Happy to copy those lines over here instead of what I have if that's preferred.

src/dual.jl
```diff
@@ -257,6 +257,14 @@ Base.copy(d::Dual) = d
 Base.eps(d::Dual) = eps(value(d))
 Base.eps(::Type{D}) where {D<:Dual} = eps(valtype(D))
 
+function Base.nextfloat(d::ForwardDiff.Dual{T,V,N}) where {T,V,N}
+    ForwardDiff.Dual{T}(nextfloat(d.value), d.partials)
+end
+
+function Base.prevfloat(d::ForwardDiff.Dual{T,V,N}) where {T,V,N}
+    ForwardDiff.Dual{T}(prevfloat(d.value), d.partials)
```
YingboMa, line 265:
```suggestion
    ForwardDiff.Dual{T}(prevfloat(float(d.value)), d.partials)
```

src/dual.jl
```diff
@@ -257,6 +257,14 @@ Base.copy(d::Dual) = d
 Base.eps(d::Dual) = eps(value(d))
 Base.eps(::Type{D}) where {D<:Dual} = eps(valtype(D))
 
+function Base.nextfloat(d::ForwardDiff.Dual{T,V,N}) where {T,V,N}
+    ForwardDiff.Dual{T}(nextfloat(d.value), d.partials)
```
YingboMa, line 261:
```suggestion
    ForwardDiff.Dual{T}(nextfloat(float(d.value)), d.partials)
```

src/dual.jl
devmotion: On the other hand, e.g., `prevfloat(::Int)` is not defined, so maybe it's fine without `float(...)` and it should just error if `prevfloat(d.value)` is not defined?

YingboMa: A code pattern can be
```julia
foo(x) = goo(float(x))
goo(x) = ...
```
Oh yeah. Yes, that's fine.

ChrisRackauckas:
I think this is good enough to merge and has been sitting for comments long enough that I'm going to go for it. I agree with @devmotion here that since nextfloat is only defined for floats I don't think it needs to promote, but there can be follow up discussion on that. Even if we decide on that, it won't change the standard behavior on floats, so it would be a non-breaking change which makes merging as is fine.

Issue #364, opened 2018-10-09
support for complex valued function


mistguy:
IIUC, ForwardDiff.jl does not support complex valued function such as `x->exp(im*x)`. Is there any plan to support this? I believe this is possible since [AutoGrad.jl](https://github.com/denizyuret/AutoGrad.jl) seems to support this.

fredrikekre:
Ref #157

mistguy:
I believe #157 is talking about complex arguments, not complex valued functions. Complex values functions should be much simpler since I am talking real arguments.

devmotion:
Support for complex-valued functions with real arguments was added in https://github.com/JuliaDiff/ForwardDiff.jl/pull/583:
```julia
julia> using ForwardDiff

julia> ForwardDiff.derivative(x->exp(im*x), 1.0)
-0.8414709848078965 + 0.5403023058681398im
```

petar-andrejic:
> Support for complex-valued functions with real arguments was added in #583:
> 
> ```julia
> julia> using ForwardDiff
> 
> julia> ForwardDiff.derivative(x->exp(im*x), 1.0)
> -0.8414709848078965 + 0.5403023058681398im
> ```

ForwardDiff.gradient etc don't work for complex valued functions of multiple real arguments

petar-andrejic:
e.g.

```
julia> ForwardDiff.gradient(x->exp(im * x[1]+x[2]), [3.0,2.0])
ERROR: DimensionMismatch: gradient(f, x) expects that f(x) is a real number. Perhaps you meant jacobian(f, x)?
```

Issue #157, opened 2016-11-15
Support for complex numbers


AlexanderKoshkarov:
It would be nice if the package would support the complex numbers. Probably it is difficult to check when the derivative exist, and what path to take on the complex plane to take derivative. However if you will make default derivative path (for example along real axis) it should be good enough (I think) for [DifferentialEquations.jl](https://github.com/JuliaDiffEq/DifferentialEquations.jl) where derivatives are used to calculate Jacobian in implicit ODE integrators. (@ChrisRackauckas is using `derivative!` and `jacobian!` in DifferentialEquations.jl)
If it would be possible to indicate the curve on complex plane where to take derivative, it also would be awesome. Thanks.
```
julia> using ForwardDiff; import ForwardDiff

julia> f(x::Vector) = sum(sin, x) + prod(tan, x) * sum(sqrt, x);

julia> x = rand(5); xi = rand(Complex128,5);

julia> g = x -> ForwardDiff.gradient(f, x);

julia> g(x)
5-element Array{Float64,1}:
 1.48017
 3.33484
 1.36112
 1.55229
 1.376  

julia> g = x -> ForwardDiff.gradient(f, xi);

julia> g(xi)
ERROR: TypeError: Dual: in T, expected T<:Real, got Type{Complex{Float64}}
 in Type at /home/kosh/.julia/v0.5/ForwardDiff/src/cache.jl:52 [inlined]
 in construct_jacobian_caches(::Array{Complex{Float64},1}, ::ForwardDiff.Chunk{5}) at /home/kosh/.julia/v0.5/ForwardDiff/src/cache.jl:157
 in get!(::ForwardDiff.##23#24{Array{Complex{Float64},1},ForwardDiff.Chunk{5}}, ::Dict{Tuple{Int64,Int64,DataType,Bool},Any}, ::Tuple{Int64,Int64,DataType,Bool}) at ./dict.jl:663
 in multithread_jacobian_cachefetch!(::Array{Complex{Float64},1}, ::ForwardDiff.Chunk{5}, ::Bool, ::Bool) at /home/kosh/.julia/v0.5/ForwardDiff/src/cache.jl:65
 in jacobian_cachefetch! at /home/kosh/.julia/v0.5/ForwardDiff/src/cache.jl:74 [inlined]
[...]

PR #455, opened 2020-05-24
complex number support

GiggleLiu:
Related Issues

#157 
#324 
#364 

This is a quick fix for some complex valued functions. Any thoughts?

NOTE: I copied some functions from Base, because they fixed the type to foating point numbers.

### DIFF
--- a/src/complex.jl
+++ b/src/complex.jl
@@ -1,0 +1,172 @@
+Base.prevfloat(x::Dual{T,V}) where {T,V<:AbstractFloat} = prevfloat(x.value)
+Base.nextfloat(x::Dual{T,V}) where {T,V<:AbstractFloat} = nextfloat(x.value)
+
+function Base.log(z::Complex{T}) where {A, FT<:AbstractFloat, T<:Dual{A,FT}}
+    T1::T  = 1.25
+    T2::T  = 3
+    ln2::T = log(convert(T,2))  #0.6931471805599453
+    x, y = reim(z)
+    ρ, k = Base.ssqs(x,y)
+    ax = abs(x)
+    ay = abs(y)
+    if ax < ay
+        θ, β = ax, ay
+    else
+        θ, β = ay, ax
+    end
+    if k==0 && (0.5 < β*β) && (β <= T1 || ρ < T2)
+        ρρ = log1p((β-1)*(β+1)+θ*θ)/2
+    else
+        ρρ = log(ρ)/2 + k*ln2
+    end
+    Complex(ρρ, angle(z))
+end
+function Base.tanh(z::Complex{T}) where {A, FT<:AbstractFloat, T<:Dual{A,FT}}
+    Ω = prevfloat(typemax(T))
+    ξ, η = reim(z)
+    if isnan(ξ) && η==0 return Complex(ξ, η) end
+    if 4*abs(ξ) > asinh(Ω) #Overflow?
+        Complex(copysign(one(T),ξ),
+                copysign(zero(T),η*(isfinite(η) ? sin(2*abs(η)) : one(η))))
+    else
+        t = tan(η)
+        β = 1+t*t #sec(η)^2
+        s = sinh(ξ)
+        ρ = sqrt(1 + s*s) #cosh(ξ)
+        if isinf(t)
+            Complex(ρ/s,1/t)
+        else
+            Complex(β*ρ*s,t)/(1+β*s*s)
+        end
+    end
+end
+
+_convert(T, x::Dual) = convert(T, x.value)
+function Base._cpow(z::Union{Dual{A,T}, Complex{<:Dual{A,T}}}, p::Union{Dual{B,T}, Complex{<:Dual{B,T}}}) where {T,A,B}
+    if isreal(p)
+        pᵣ = real(p)
+        if isinteger(pᵣ) && abs(pᵣ) < typemax(Int32)
+            # |p| < typemax(Int32) serves two purposes: it prevents overflow
+            # when converting p to Int, and it also turns out to be roughly
+            # the crossover point for exp(p*log(z)) or similar to be faster.
+            if iszero(pᵣ) # fix signs of imaginary part for z^0
+                zer = flipsign(copysign(zero(T),pᵣ), imag(z))
+                return Complex(one(T), zer)
+            end
+            ip = _convert(Int, pᵣ)
+            if isreal(z)
+                zᵣ = real(z)
+                if ip < 0
+                    iszero(z) && return Complex(T(NaN),T(NaN))
+                    re = Base.power_by_squaring(inv(zᵣ), -ip)
+                    im = -imag(z)
+                else
+                    re = Base.power_by_squaring(zᵣ, ip)
+                    im = imag(z)
+                end
+                # slightly tricky to get the correct sign of zero imag. part
+                return Complex(re, ifelse(iseven(ip) & signbit(zᵣ), -im, im))
+            else
+                return ip < 0 ? Base.power_by_squaring(inv(z), -ip) : Base.power_by_squaring(z, ip)
+            end
+        elseif isreal(z)
+            # (note: if both z and p are complex with ±0.0 imaginary parts,
+            #  the sign of the ±0.0 imaginary part of the result is ambiguous)
+            if iszero(real(z))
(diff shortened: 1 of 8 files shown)


src/complex.jl
```diff
@@ -0,0 +1,172 @@
+Base.prevfloat(x::Dual{T,V}) where {T,V<:AbstractFloat} = prevfloat(x.value)
+Base.nextfloat(x::Dual{T,V}) where {T,V<:AbstractFloat} = nextfloat(x.value)
+
+function Base.log(z::Complex{T}) where {A, FT<:AbstractFloat, T<:Dual{A,FT}}
+    T1::T  = 1.25
```
YingboMa, line 5:
```suggestion
    T1  = convert(T, 5//4)
```

src/complex.jl
```diff
@@ -0,0 +1,172 @@
+Base.prevfloat(x::Dual{T,V}) where {T,V<:AbstractFloat} = prevfloat(x.value)
+Base.nextfloat(x::Dual{T,V}) where {T,V<:AbstractFloat} = nextfloat(x.value)
+
+function Base.log(z::Complex{T}) where {A, FT<:AbstractFloat, T<:Dual{A,FT}}
```
YingboMa, line 4:
```suggestion
function Base.log(z::Complex{T}) where {A, T<:Dual{A,<:AbstractFloat}}
```

src/complex.jl
```diff
@@ -0,0 +1,172 @@
...
+    T1::T  = 1.25
+    T2::T  = 3
```
YingboMa, line 6:
```suggestion
    T2  = convert(T, 3)
```

src/complex.jl
```diff
@@ -0,0 +1,172 @@
...
+    ln2::T = log(convert(T,2))  #0.6931471805599453
```
YingboMa, line 7: I don't think you need a type annotation here.

src/complex.jl
YingboMa: I am not sure about this, because the compiler cannot optimize the rational number away.

test/ComplexTest.jl
```diff
@@ -0,0 +1,53 @@
+using ForwardDiff: Dual
```
YingboMa, line 1: ForwardDiff wraps all test files into a module.

src/complex.jl
```diff
@@ -0,0 +1,172 @@
...
+        θ, β = ay, ax
+    end
+    if k==0 && (0.5 < β*β) && (β <= T1 || ρ < T2)
+        ρρ = log1p((β-1)*(β+1)+θ*θ)/2
+    else
+        ρρ = log(ρ)/2 + k*ln2
+    end
+    Complex(ρρ, angle(z))
+end
+function Base.tanh(z::Complex{T}) where {A, FT<:AbstractFloat, T<:Dual{A,FT}}
```
YingboMa, line 24:
```suggestion
function Base.tanh(z::Complex{T}) where {A, T<:Dual{A,<:AbstractFloat}}
```

src/complex.jl
GiggleLiu: This function is copied from Base. One of the goal of this PR is to stay consistent with the implementation in Base.

https://github.com/JuliaLang/julia/blob/575701248adc951038b9727e646230737e5b3c91/base/complex.jl#L564

test/ComplexTest.jl
GiggleLiu: Yes, interesting design. May I know the reason to put tests in a module?

src/complex.jl
GiggleLiu: same as above.

test/ComplexTest.jl
GiggleLiu: I have updated the tests. Good now?

YingboMa: It won't introduce variables to `Main`, in this way. So tests don't influence each other. This is the motivation for https://github.com/YingboMa/SafeTestsets.jl, too.

GiggleLiu:
@YingboMa 
Although we get only 3 votes in #456 , but I need to differentiate complex number in one of my recent projects.
Do you think we should take the chance to polish and get this PR merged?

src/complex.jl
```diff
@@ -0,0 +1,172 @@
+Base.prevfloat(x::Dual{T,V}) where {T,V<:AbstractFloat} = prevfloat(x.value)
```
ChrisRackauckas, line 1: see https://github.com/JuliaDiff/ForwardDiff.jl/pull/453

GiggleLiu: Thanks. Is that PR going to be merged soon? If so, feel free to merge that PR first, I can do the compatibility check later.

ChrisRackauckas:
Is this not all covered by https://github.com/JuliaLang/julia/pull/36030 ?

src/complex.jl
ChrisRackauckas: done

ChrisRackauckas:
Looking at the two PRs, it seems like @YingboMa upstreamed it all to Base where it should go (correct me if I'm wrong and anything was left out). So complex AD is thus fixed on v1.6. However, the tests are still definitely useful from here. So I'd suggest the following:

- Move these overloads to an internal module. Document in the README `using ForwardDiff.Complex` for v1.4 as a compatibility workaround.
- Use this to in the tests

and everything is good to go. The other alternative would be to move these behind an `@static if` definition, but I would be weary of having such central dispatches modified by such a core package outside of the system image. If you want invalidations, this is how you get invalidations.

ChrisRackauckas:
Handled by https://github.com/JuliaLang/julia/pull/36030

GiggleLiu:
Thanks! @ChrisRackauckas

DEKOHG:
Hello,
I just installed the newest version of DifferentialEquations.jl . However, I still got an error with stiff solvers (eg. Rodas4P) for complex numbers calculation. I have tried on both julia 1.8.2 and julia 1.7.2 and my DifferentialEquations is v7.6.0

ERROR: ArgumentError: Cannot create a dual over scalar type ComplexF64. If the type behaves as a scalar, define ForwardDiff.can_dual(::Type{ComplexF64}) = true.

petar-andrejic:
> Handled by [JuliaLang/julia#36030](https://github.com/JuliaLang/julia/pull/36030)

Not entirely handled, complex valued functions of multiple real arguments don't work, e.g.

```julia
julia> ForwardDiff.gradient(x->exp(im * (x[1] + x[2])), [1.0,2.0])

ERROR: DimensionMismatch: gradient(f, x) expects that f(x) is a real number. Perhaps you meant jacobian(f, x)?
```

davidschlegel:
> Hello, I just installed the newest version of DifferentialEquations.jl . However, I still got an error with stiff solvers (eg. Rodas4P) for complex numbers calculation. I have tried on both julia 1.8.2 and julia 1.7.2 and my DifferentialEquations is v7.6.0
> 
> ERROR: ArgumentError: Cannot create a dual over scalar type ComplexF64. If the type behaves as a scalar, define ForwardDiff.can_dual(::Type{ComplexF64}) = true.

I also still get the same error. Is there a way to manually change the way forward differentiation is handled?

ChrisRackauckas:
For DiffEq? You just set `autodiff=false` (`Rodas5P(autodiff=false)`) or pass a `jac` function into `ODEFunction` if you want full control of it.

davidschlegel:
Sorry, it was not clear from my question. Of course, setting `autodiff=false` works but is not efficient.
Can I autodiff a complex ODE in DiffEq? Currently it does not work out-of-the-box, but perhaps one can manually pass a Jacobian that is calculated using ForwardDiff...

ChrisRackauckas:
It uses `ForwardDiff` by default, and you can pass any Jacobian function you want via `jac` in the `ODEFunction`.

davidschlegel:
Exactly, but right now, this throws an error, when trying to autodiff a complex ODE.

ChrisRackauckas:
So then I don't get the question?

davidschlegel:
Right now, afaik, ForwardDiff does not handle automatic differentiation for complex functions out-of-the-box. However, DiffEq relies on ForwardDiff when computing jacobians and more. Is there any solution to still use automatic differentiation (eg using a different package) for DiffEq?

ChrisRackauckas:
Yes, just use `jac` in the `ODEFunction`.

PR #463, opened 2020-07-27
Allow value and partials to have distinct types

timholy:
This is one of those love/hate changes...in brief, it allows one to set up alternate rules for how the "epsilon" component of dual numbers behaves by changing the type of `Dual` from
```julia
struct Dual{T,V,N} <: Real
    value::V
    partials::Partials{N,V}
end
```
to
```julia
struct Dual{T,V,N,P} <: Real
    value::V
    partials::Partials{N,P}
end
```

Happy to discuss motivations for this change at my poster at JuliaCon (we should leave this open at least until then).

It also throws in a commit to change from `one` to `oneunit` in places where IMO the latter is technically the correct choice.

Someone with permissions should tag this "breaking change" (note the version bump to 0.11).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ForwardDiff"
 uuid = "f6369f11-7733-5829-9624-2563aa707210"
-version = "0.10.12"
+version = "0.11.0"
 
 [deps]
 CommonSubexpressions = "bbf7d656-a473-5ed7-a52c-81e309532950"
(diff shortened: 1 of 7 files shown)


ChrisRackauckas:
I don't think it's actually breaking because the view has been pretty clear that direct usage of Dual numbers is not part of the public facing API. I'd be willing to accept this.

I'd say we should probably start thinking about what a ForwardDiff 1.0 means. I think the library is quite mature, and would like to get "last things in" and just lock this library down.

oxinabox (approved):

Keno:
Leaving a reference to https://github.com/JuliaDiff/ForwardDiff.jl/issues/546 here, I'd like `Partials` also not be type-uniform, i.e. allowing each partial to be of a different type.

ChrisRackauckas:
Thinking about this some more, the Pumas type signatures are going to explode 100x in size because the use Duals of Duals of Duals, so now each of the partials is going to be printing out the dual of dual tag. We might need to fix Julia's type printing to not be so verbose by default if we're going down this route because this probably isn't the only case that would end up with a mess.

PR #467, opened 2020-08-27
Release 1.0

oxinabox:
Noone wants to change FowardDIff.
It is stable.
Even if one *does* want to change ForwardDiff, noone wants to change the API in a breaking way.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ForwardDiff"
 uuid = "f6369f11-7733-5829-9624-2563aa707210"
-version = "0.10.12"
+version = "1.0.0"
 
 [deps]
 CommonSubexpressions = "bbf7d656-a473-5ed7-a52c-81e309532950"


ChrisRackauckas:
I'd say we do https://github.com/JuliaDiff/ForwardDiff.jl/pull/419 and https://github.com/JuliaDiff/ForwardDiff.jl/pull/463, which I don't think are actually breaking changes, but can be considered breaking changes, and we can call that v1.0. Then, once it's v1.0, I'd say we can document the Dual number interface, https://github.com/JuliaDiff/ForwardDiff.jl/pull/379 because I'd hope no one would ever change that.

I'd like to get @KristofferC and @jrevels 's input though. But I think it's pretty safe to say, ForwardDiff works and if you want a very different AD then make a new package.

oxinabox:
I agree with those 2 PRs.

fredrikekre:
Please don't make a non-breaking 1.0.0 release.

oxinabox:
Its an unfortunate concisquence of julia's semver extension that 1.0.0 is always considered breaking.
So if one realises a package that is pre-1.0.0 is actually stable one has to delcare a breaking release to indicated that there will not be many breaking releases in the future.

But  in this case I am kind of happy to have it considered breaking because #463 is borderline breaking.
Probabably fine for most uses, but I do know that sometimes people do construct duals directly, for better or worse.

cossio:
Something holding this?

KristofferC:
If someone wants to do this they need to make a corresponding change to the registry that allows ForwardDiff 1.0 for those packages that are currently claiming compatibility with ForwardDiff 0.10.

mcabbott:
Does #481 count as breaking?

oscardssmith:
I don't know how to read.

PR #475, opened 2020-11-18
Update .travis.yml to test on latest release and move x86 testing

andreasnoack:
to Travis from AppVeyor


andreasnoack:
https://github.com/JuliaDiff/ForwardDiff.jl/blob/3fb16ac21ac06b7c42c807e898395ad3bd9f54ae/test/DualTest.jl#L382 seems to consistently fail on x86

andreasnoack:
It would be great if somebody could try to run this locally with 32 bit Julia and report more details. Otherwise, I think we should consider disabling the 32 bit testing.

andreasnoack:
It looks like we have spent all of our free Travis credit so we'd have to switch to GitHub Actions so I'll close here

PR #476, opened 2020-11-18
Relax print tests

oxinabox:
Wether or not a space is printed changes depending on the julia version.
So we use a regex.

I also hard coded to use `Int32` since that simpifies the regex

### DIFF
--- a/test/PartialsTest.jl
+++ b/test/PartialsTest.jl
@@ -136,7 +136,8 @@ for N in (0, 3), T in (Int, Float32, Float64)
 end
 
 io = IOBuffer()
-show(io, MIME("text/plain"), Partials((1, 2, 3)))
-@test String(take!(io)) == "3-element ForwardDiff.Partials{3,$Int}:\n 1\n 2\n 3"
+show(io, MIME("text/plain"), Partials((Int32(1), Int32(2), Int32(3))))
+str = String(take!(io))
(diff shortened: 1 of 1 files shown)


andreasnoack (approved):

KristofferC:
Why are we testing this at all? Doesn't it just directly call a `Base` function?

oxinabox:
> Why is this tested at all? Doesn't it just directly call a `Base` function?

That is a good question.
We do have 
https://github.com/JuliaDiff/ForwardDiff.jl/blob/master/src/partials.jl#L228
which itself looks kinda iffy.
Are we testing that that one *isn't* hit?

That code was added in 
https://github.com/JuliaDiff/ForwardDiff.jl/commit/1279080326440d24d5f1146082da79c10944214f
along with a ton of other things.

but that test was added in https://github.com/JuliaDiff/ForwardDiff.jl/commit/799c444136d355850d3646af4aa5f06955b6c600
Maybe @timholy  can illuminate what we were testing for.

andreasnoack:
The show method for Partials used to fail which was why the test was added in https://github.com/JuliaDiff/ForwardDiff.jl/pull/437. I've applied the proposal from https://github.com/JuliaDiff/ForwardDiff.jl/pull/476#issuecomment-729678758.

PR #477, opened 2020-11-18
Add ColPrac badge to the readme

oxinabox:
For all intents and purposes we already follow [ColPrac](https://github.com/SciML/ColPrac) here.
We should make it official, and add the bradge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,7 @@
 
 [![](https://img.shields.io/badge/docs-stable-blue.svg)](http://www.juliadiff.org/ForwardDiff.jl/stable)
 [![](https://img.shields.io/badge/docs-latest-blue.svg)](http://www.juliadiff.org/ForwardDiff.jl/latest)
+[![ColPrac: Contributor's Guide on Collaborative Practices for Community Packages](https://img.shields.io/badge/ColPrac-Contributor's%20Guide-blueviolet)](https://github.com/SciML/ColPrac)
 
 # ForwardDiff.jl
 


KristofferC:
When did that happen?

oxinabox:
> There are many things in "ColPrac" that we do not do.

AFAIK we do all of them
Or rather, when we don't do them it isn't by intent, but by mistake.


> Also, while I understand the good intention, it seems a bit odd to me for someone without a single commit to a repo to push a contributors guide and saying so definitely that it is followed. Maybe an issue with a suggestion to start following it is more apt?

My bad, I thought we had this discussion when ColPrac was written.
I guess we did not.
I certainly had this discussion with a bunch of people who have merge rights here, and mentioned JuliaDiff adopting them.
I may well not have had one specifically with people in the JuliaDiff org.

But fair enough, feel free discuss here before we merge.
Not sure we need a seperate issue.

KristofferC:
For example, these things are stuff that I think are unnecessary to have rules about and is best left up to the discretion of the maintainers:

- PRs must have 1 approval before they are merged.
- PRs by people with merge rights must have approval from someone else, who may or may not have merge rights (and then may merge their own PR).
- If the repository is in a state where there are unreleased changes for an extended period of time in preparation for a release, then the version in the Project.toml should be set to the version number of the intended release, with the -DEV suffix.
- The person who merged the PR should register the new release of the package.

andreasnoack:
> and mentioned JuliaDiff adopting them.

In reality, many of the `JuliaX` organizations are not coordinating units. At least I don't know where the JuliaDiff coordination takes place (and I'm an admin) so I think these discussions would have to take place within each repo.

Generally, I think it's fine to write down the guidelines of the repo. I suppose it could easier for newcomers if norms are written down and I think it could also be useful if you want to convince outsiders that a package is trustworthy. However, when reading @KristofferC list, I realize that the specific list of rules here might be too long/restrictive.

PR #478, opened 2020-11-19
WIP: Bump SpecialFunctions and Julia versions

willtebbutt:
#474 didn't appear to run on SpecialFunctions 1.0, so I'm forcing it here to see what happens.

Separately, the Julia versions in the CI appears to be out of date, testing 1.4 rather than 1.5.


andreasnoack:
For the CI adjustments, see https://github.com/JuliaDiff/ForwardDiff.jl/pull/475.

andreasnoack:
I suppose that it's possible to test with SepcialFuinctions@1 now since DiffRules has been upgraded in the meanwhile but let's see.

Issue #480, opened 2020-11-27
Inequality, ε-balls, and accidental structural zeros


mcabbott:
Consider this function:
```julia
sq(x) = x==1 ? one(x) : x^2

@test FiniteDifferences.central_fdm(5, 1)(sq, 1) ≈ 2.0
@test_broken ForwardDiff.derivative(sq, 1.0) == 2.0
```
Here `ForwardDiff` gets the wrong answer, according to your first calculus class: The derivative is defined by taking limits, evaluating `sq(x + ε)` for some small `ε`, and these always see the continuum `x^2` not the special point.

One to think about this is to say that `x==1` really means `abs(x-1) < ζ` with some tiny `ζ`, which we keep finite until we are sure we aren't confused. The calculus class assumption is that `ζ << ε`. 

The assumption of `ForwardDiff` is the opposite. Its `Dual(x,1)` encodes a perturbation `x + 1ε` with `ε` smaller than everything else around, and in particular `ε << ζ`. Or in other words, `sq` is viewed as being piecewise continuous, with a small flat area of width `2ζ`, which is still large enough for us to see that its slope is zero.

Of course nobody really writes contrived examples like `sq`. But they do write things like this:
```julia
function prod1(xs::Vector)
    p = one(eltype(xs))
    for x in xs
        p = p * x
    end
    p
end

function prod2(xs::Vector)
    p = one(eltype(xs))
    for x in xs
        p = p * x
        p == 0 && break  # exit early once you know the answer
    end
    p
end

@test ForwardDiff.gradient(prod1, [1,2,0,4,0,6]) == zeros(6)
@test_broken ForwardDiff.gradient(prod2, [1,2,0,4,0,6]) == zeros(6)
```

This has almost the same problem as #197, where `det(A)` tests for `istriu(A) || istril(A)` before calling a simpler branch. The fact that `f(x,y) == g(x,y)` when `y==0` does not imply that `df/dy == dg/dy`. So it seems AD ought not to take that branch.

In which case we want something like this:
```julia
Base.:(==)(x::Dual, y::Int) = x.value == y && iszero(x.partials)
Base.:(!=)(x::Dual, y::Int) = x.value != y || !iszero(x.partials)
```
This fixes the tests above, and (a slightly more careful version) fixes #197 and #407. 

[...]

PR #481, opened 2020-11-27
Change `==` to ignore measure-zero branches

### LINKED ISSUE #197: support for the determinant function
I naively tried `ForwardDiff.gradient(det, A)` and compared it to the output of `(det(A) * inv(A))' * A)`([Jacobi's Formula](https://en.wikipedia.org/wiki/Jacobi's_formula) for a real square matrix `A`. They should be the same value, but they are radically different.

What am I getting wrong here? Shouldn't ForwardDiff support `det` out of the box since it's a core function of Julia?

Thanks.

### LINKED ISSUE #407: det not handled correctly?
The following code gives different results for the LinearAlgebra det and a hand-written determinant function:
MWE
```
using LinearAlgebra
using ForwardDiff: derivative, gradient, jacobian, hessian

E, nu = (7.0e6, 0.3) 
lambda = E * nu / (1 + nu) / (1 - 2*(nu));
mu = E / 2. / (1+nu);

function strain6vto3x3t!(t::Array{T,2}, v::Array{T,1}) where {T}
    t[1,1] = v[1];
    t[2,2] = v[2];
    t[3,3] = v[3];
    t[1,2] = v[4]/2.;
    t[2,1] = v[4]/2.;
    t[1,3] = v[5]/2.;
    t[3,1] = v[5]/2.;
    t[3,2] = v[6]/2.;
    t[2,3] = v[6]/2.;
    return t
end

function strain3x3tto6v!(v::Array{T,1}, t::Array{T,2}) where {T}
    v[1] = t[1,1];
    v[2] = t[2,2];
    v[3] = t[3,3];
    v[4] = t[1,2] + t[2,1];
    v[5] = t[1,3] + t[3,1];
    v[6] = t[3,2] + t[2,3];
    return v
end

mydet(C) = begin
C[1,1] * C[2,2] * C[3,3] + 
C[1,2] * C[2,3] * C[3,1] + 
C[1,3] * C[2,1] * C[3,2] - 
C[1,3] * C[2,2] * C[3,1] - 
C[1,2] * C[2,1] * C[3,3] - 
C[1,1] * C[2,3] * C[3,2]
end

for i in 1:100
	a = rand(3, 3)
	@assert (det(a) - mydet(a)) / det(a) <= 1.0e-6
end

function strainenergy(Cv, lambda, mu)
	C = fill(zero(eltype(Cv)), 3, 3)
	strain6vto3x3t!(C, Cv)
	J = sqrt(mydet(C))
	lJ = log(J)
	return mu/2*(tr(C)-3) - mu*lJ + (lambda/2)*lJ^2;
end

function strainenergy1(Cv, lambda, mu)
	C = fill(zero(eltype(Cv)), 3, 3)
	strain6vto3x3t!(C, Cv)
	J = sqrt(det(C))
	lJ = log(J)
	return mu/2*(tr(C)-3) - mu*lJ + (lambda/2)*lJ^2;
end

Fn1 = [1.1 0 0; 0 1.0 0; 0 0 1.0]
Fn = [1.0 0 0; 0 1.0 0; 0 0 1.0]
C = Fn1'*Fn1;
Cv = fill(0.0, 6)
strain3x3tto6v!(Cv, C)
@show Da = hessian(Cv -> strainenergy(Cv, lambda, mu), Cv);
@show Da = hessian(Cv -> strainenergy1(Cv, lambda, mu), Cv);
```
Julia Version 1.3.0-alpha.0                                
Commit 6c11e7c2c4 (2019-07-23 01:46 UTC)                   
Platform Info:                                             
  OS: Windows (x86_64-w64-mingw32)                         
  CPU: Intel(R) Core(TM) i7-8705G CPU @ 3.10GHz            
  WORD_SIZE: 64                                            
  LIBM: libopenlibm                                        
  LLVM: libLLVM-6.0.1 (ORCJIT, skylake)      

    Status `C:\Users\PetrKrysl\.julia\environments\v1.3\Project.toml`               
                             
  [f6369f11] ForwardDiff v0.10.3

### LINKED ISSUE #490: Derivative of matrix inverse for diagonal matrix is not correct?
I am here to reporting some dubious results from the derivative of a matrix inversion for the diagonal matrix.

Package version: [f6369f11] ForwardDiff v0.10.14

`A = [0.5 0 0 ; 0 0.5 0; 0 0 0.5]`

`ForwardDiff.gradient(A -> sum(inv(A)), A)`
result:
```
3×3 Array{Float64,2}:    
 -4.0  -4.0  -4.0
  0.0  -4.0  -4.0
  0.0   0.0  -4.0
```

However, the analytical result should be:
`-inv(A) * ones(3,3) * inv(A)`
result:
```
3×3 Array{Float64,2}:
 -4.0  -4.0  -4.0
 -4.0  -4.0  -4.0
 -4.0  -4.0  -4.0
```

The Zygote gradient function 
`Zygote.gradient(A -> sum(inv(A)), A)[1]`

result:
```
3×3 Array{Float64,2}:
 -4.0  -4.0  -4.0
 -4.0  -4.0  -4.0
 -4.0  -4.0  -4.0
```

### LINKED ISSUE #506: Incorrect gradient with cascaded reshape, linear equation solver operations
For the following function A(η::Array) = η[1:2]'*(reshape(η[2+1:end],(2,2))\η[1:2]), I observed a weird behaviour with ForwardDiff. Suppose that we want to evaluate the gradient at x = [2.0, 2.0, 1.0, 0.0, 0.0, 1.0]. Then ForwardDiff.gradient(A,x) returns the output [4.0, 4.0, -4.0, 0.0, -4.0, -4.0]. However, I would expect 4th element of the gradient to be -4.0. If I change the vector x slightly as [2.0, 2.0, 1.0, 0.00001, 0.00001, 1.0], then the gradient turns out to be correct and I get the output [3.999960000399996, 3.999960000399996, -3.9999200011999845, -3.9999200011999845, -3.9999200011999845, -3.9999200011999845].

### LINKED ISSUE #536: Derivative lost in 3-arg mul
```
julia> a = randn(3,3)
b =3×3 Matrix{Float64}:
 ra  0.0237017   1.62282  -1.28889
 -0.186575   -2.00309   0.502572
 -0.413123   -1.99718  -0.174593

julia> b = randn(3,3)
3×3 Matrix{Float64}:
 -0.40664   1.80043     1.2399
  1.22559  -0.828653    1.38103
  1.40994  -0.0784539  -0.127629

julia> ForwardDiff.derivative(x -> sum(x*a*b), 0.0)
0.0

julia> ForwardDiff.derivative(x -> sum(x*(a*b)), 0.0)
-6.900806270529478
```

### LINKED ISSUE #542: Inconsistent behavior in assignment to sparse matrices of Dual numbers
Let us say that we have a sparse matrix where the entries are Dual numbers.
```julia
using ForwardDiff, SparseArrays
d = ForwardDiff.Dual(1.0, 2.0)
a = sparse([0 d])
```
Let us define two zero values with different partials. One with zero partials (true zero) and one with non-zero partials (false zero):
```julia
# 0, with ∂ = 1
d_false_zero = ForwardDiff.Dual(0.0, 1.0)
# 0, with ∂ = 0
d_true_zero = ForwardDiff.Dual(0.0, 0.0)
```
If we replace the existing entry at `[2]`, everything is well:
```julia
## Insertion into existing element keeps partials
a[2] = d_false_zero
println("Existing element:")
println(a[2].partials == d_false_zero.partials)
```
This gives the expected output:
```
Existing element:
true
```
Now let us try to insert our false zero into the first position, which we know isn't yet stored:
```julia
## Insertion into unstored element drops partials
a[1] = d_false_zero
println("Zero element:")
println(a[1].partials == d_false_zero.partials)
```
```
Zero element:
false
```
The partials were lost, due to the `iszero` check here in SparseArrays: https://github.com/JuliaLang/julia/blob/cb30aa7a089f4d5f1c9f834d96c3788f2e93ebcc/stdlib/SparseArrays/src/sparsematrix.jl#L2652

Some quick and ugly type piracy can fix this, but may have other unintended consequences:
```
import Base.iszero
Base.iszero(d::ForwardDiff.Dual) = iszero(d.value) && iszero(d.partials)
```
I'm not sure what the intended behavior should be, since the entry is technically zero, but not a neutral element for addition of Duals.

### LINKED ISSUE #551: The Hessian of x-> dot(x,H,x) should be H+H', not zero
Hi all,

ForwardDiff incorrectly computes the Hessian of three-argument-dot.

Cook up some (any) square matrix
julia> H = [1 2 3; 4 5 6; 7 8 9];

Compute Hessian of three-argument dot. Yields all zeros. Should yield H+H'.
julia> ForwardDiff.hessian(x->dot(x,H,x), zeros(3))
3×3 Array{Float64,2}:
 0.0  0.0  0.0
 0.0  0.0  0.0
 0.0  0.0  0.0

Manually expand that three argument dot call. Now result is the expected H+H':
julia> ForwardDiff.hessian(x->x'H*x, zeros(3))
3×3 Array{Float64,2}:
  2.0   6.0  10.0
  6.0  10.0  14.0
 10.0  14.0  18.0

Curious!

Cheers,

Wouter

mcabbott:
This changes the meaning of `==` for dual numbers, to demand that both real and ~~imaginary~~ dual parts match. 

Fixes #197, fixes #407. Edit -- also closes #490, closes #506, closes #536, closes #542, closes #551.

See #480 for too-long discussion of why I think this makes sense. And comments on functions like `>`, which this PR does not alter.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ForwardDiff"
 uuid = "f6369f11-7733-5829-9624-2563aa707210"
-version = "0.10.32"
+version = "0.11.0"
 
 [deps]
 CommonSubexpressions = "bbf7d656-a473-5ed7-a52c-81e309532950"
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ForwardDiff"
 uuid = "f6369f11-7733-5829-9624-2563aa707210"
-version = "0.10.30"
+version = "0.10.33"
 
 [deps]
 CommonSubexpressions = "bbf7d656-a473-5ed7-a52c-81e309532950"
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ForwardDiff"
 uuid = "f6369f11-7733-5829-9624-2563aa707210"
-version = "0.11.0"
+version = "0.10.30"
 
 [deps]
 CommonSubexpressions = "bbf7d656-a473-5ed7-a52c-81e309532950"
--- a/src/dual.jl
+++ b/src/dual.jl
@@ -384,9 +384,21 @@ for pred in UNARY_PREDICATES
     @eval Base.$(pred)(d::Dual) = $(pred)(value(d))
 end
 
-# BINARY_PREDICATES = Symbol[:isequal, :isless, :<, :>, :(==), :(!=), :(<=), :(>=)] # prelude.jl
+# Before PR#481 this loop ran over this list:
+# BINARY_PREDICATES = Symbol[:isequal, :isless, :<, :>, :(==), :(!=), :(<=), :(>=)]
+# Not a minimal set, as Base defines some in terms of others.
+for pred in [:isless, :<, :>, :(<=), :(>=)]
+    @eval begin
+        @define_binary_dual_op(
+            Base.$(pred),
+            $(pred)(value(x), value(y)),
+            $(pred)(value(x), y),
+            $(pred)(x, value(y)),
+        )
+    end
+end
 
-Base.iszero(x::Dual) = iszero(value(x)) && iszero(partials(x))
+Base.iszero(x::Dual) = iszero(value(x)) && iszero(partials(x))  # shortcut, equivalent to x == zero(x)
 
 for pred in [:isequal, :(==)]
     @eval begin
@@ -405,17 +417,6 @@ end
     (!=)(value(x), y)        || !iszero(partials(x)),
     (!=)(x, value(y))        || !iszero(partials(y)),
 )
-
-for pred in [:isless, :<, :>, :(<=), :(>=)]
-    @eval begin
-        @define_binary_dual_op(
-            Base.$(pred),
-            $(pred)(value(x), value(y)),
-            $(pred)(value(x), y),
-            $(pred)(x, value(y)),
-        )
-    end
-end
 
 ########################
 # Promotion/Conversion #
--- a/src/dual.jl
+++ b/src/dual.jl
@@ -390,18 +390,18 @@ for pred in [:isequal, :(==)]
     @eval begin
         @define_binary_dual_op(
             Base.$(pred),
-            $(pred)(value(x), value(y)) && $(pred)(partials(x),       partials(y)),
-            $(pred)(value(x), y)        && $(pred)(partials(x), zero(partials(x))),
-            $(pred)(x, value(y))        && $(pred)(zero(partials(y)), partials(y)),
+            $(pred)(value(x), value(y)) && $(pred)(partials(x), partials(y)),
+            $(pred)(value(x), y)        && iszero(partials(x)),
+            $(pred)(x, value(y))        && iszero(partials(y)),
         )
     end
 end
 
 @define_binary_dual_op(
     Base.:(!=),
-    (!=)(value(x), value(y)) || (!=)(partials(x),       partials(y)),
-    (!=)(value(x), y)        || (!=)(partials(x), zero(partials(x))),
-    (!=)(x, value(y))        || (!=)(zero(partials(y)), partials(y)),
+    (!=)(value(x), value(y)) || (!=)(partials(x), partials(y)),
+    (!=)(value(x), y)        || !iszero(partials(x)),
+    (!=)(x, value(y))        || !iszero(partials(y)),
 )
 
 for pred in [:isless, :<, :>, :(<=), :(>=)]
@@ -413,23 +413,11 @@ for pred in [:isless, :<, :>, :(<=), :(>=)]
             else
                 $(pred)(value(x), value(y))
             end,
-            if value(x) == y # only x is Dual
-                $(pred)(partials(x), zero(partials(x)))
-            else
-                $(pred)(value(x), value(y))
-            end,
-            if x == value(y) # only y is Dual
-                $(pred)(zero(partials(y)), partials(y))
-            else
-                $(pred)(value(x), value(y))
-            end,
+            $(pred)(value(x), y), # only x is Dual
+            $(pred)(x, value(y)), # only y is Dual
         )
     end
 end
-
-# @define_binary_dual_op(Base.:(==), false, false, false)
-# @define_binary_dual_op(Base.isequal, false, false, false)
-# @define_binary_dual_op(Base.:(!=), true, true, true)
 
 ########################
 # Promotion/Conversion #
--- a/src/dual.jl
+++ b/src/dual.jl
@@ -384,7 +384,9 @@ for pred in UNARY_PREDICATES
     @eval Base.$(pred)(d::Dual) = $(pred)(value(d))
 end
 
-# BINARY_PREDICATES = Symbol[:isequal, :isless, :<, :>, :(==), :(!=), :(<=), :(>=)]
+# BINARY_PREDICATES = Symbol[:isequal, :isless, :<, :>, :(==), :(!=), :(<=), :(>=)] # prelude.jl
+
+Base.iszero(x::Dual) = iszero(value(x)) && iszero(partials(x))
 
 for pred in [:isequal, :(==)]
     @eval begin
@@ -408,13 +410,9 @@ for pred in [:isless, :<, :>, :(<=), :(>=)]
     @eval begin
         @define_binary_dual_op(
             Base.$(pred),
-            if value(x) == value(y) # both Dual
-                $(pred)(partials(x), partials(y))
-            else
-                $(pred)(value(x), value(y))
-            end,
-            $(pred)(value(x), y), # only x is Dual
-            $(pred)(x, value(y)), # only y is Dual
+            $(pred)(value(x), value(y)),
+            $(pred)(value(x), y),
+            $(pred)(x, value(y)),
         )
     end
 end
--- a/src/dual.jl
+++ b/src/dual.jl
@@ -384,16 +384,52 @@ for pred in UNARY_PREDICATES
     @eval Base.$(pred)(d::Dual) = $(pred)(value(d))
 end
 
-for pred in BINARY_PREDICATES
+# BINARY_PREDICATES = Symbol[:isequal, :isless, :<, :>, :(==), :(!=), :(<=), :(>=)]
+
+for pred in [:isequal, :(==)]
     @eval begin
         @define_binary_dual_op(
             Base.$(pred),
-            $(pred)(value(x), value(y)),
-            $(pred)(value(x), y),
-            $(pred)(x, value(y))
+            $(pred)(value(x), value(y)) && $(pred)(partials(x),       partials(y)),
+            $(pred)(value(x), y)        && $(pred)(partials(x), zero(partials(x))),
+            $(pred)(x, value(y))        && $(pred)(zero(partials(y)), partials(y)),
         )
     end
 end
+
+@define_binary_dual_op(
+    Base.:(!=),
+    (!=)(value(x), value(y)) || (!=)(partials(x),       partials(y)),
+    (!=)(value(x), y)        || (!=)(partials(x), zero(partials(x))),
+    (!=)(x, value(y))        || (!=)(zero(partials(y)), partials(y)),
+)
+
+for pred in [:isless, :<, :>, :(<=), :(>=)]
+    @eval begin
+        @define_binary_dual_op(
+            Base.$(pred),
+            if value(x) == value(y) # both Dual
+                $(pred)(partials(x), partials(y))
+            else
+                $(pred)(value(x), value(y))
+            end,
+            if value(x) == y # only x is Dual
+                $(pred)(partials(x), zero(partials(x)))
+            else
+                $(pred)(value(x), value(y))
+            end,
+            if x == value(y) # only y is Dual
+                $(pred)(zero(partials(y)), partials(y))
+            else
+                $(pred)(value(x), value(y))
+            end,
+        )
+    end
+end
+
+# @define_binary_dual_op(Base.:(==), false, false, false)
+# @define_binary_dual_op(Base.isequal, false, false, false)
+# @define_binary_dual_op(Base.:(!=), true, true, true)
 
 ########################
 # Promotion/Conversion #
--- a/test/DualTest.jl
+++ b/test/DualTest.jl
@@ -30,7 +30,7 @@ ForwardDiff.:≺(::Int, ::Type{TestTag()}) = false
 ForwardDiff.:≺(::Type{TestTag}, ::Type{OuterTestTag}) = true
 ForwardDiff.:≺(::Type{OuterTestTag}, ::Type{TestTag}) = false
 
-@testset "Dual{TestTag(),$V,$N} and Dual{TestTag(),Dual{TestTag(),$V,$M},$N}" for N in (0,3), M in (0,4), V in (Int, Float32)
+@testset "Dual{Z,$V,$N} and Dual{Z,Dual{Z,$V,$M},$N}" for N in (0,3), M in (0,4), V in (Int, Float32)
     println("  ...testing Dual{TestTag(),$V,$N} and Dual{TestTag(),Dual{TestTag(),$V,$M},$N}")
 
     PARTIALS = Partials{N,V}(ntuple(n -> intrand(V), N))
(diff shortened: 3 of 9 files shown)


mcabbott:
Test failure on master is that `"3-element ForwardDiff.Partials{3, Int32}"` now has a space in it. 

Test failure on 1.0 might be my fault.

KristofferC:
I also think this makes sense. I am a little bit worried about the implications it will have though, perhaps there isn't a better way than to go ahead and try. Would it be considered a bugfix or a breaking change?

@jrevels, any thoughts here?

mcabbott:
Here's something this PR breaks:
```julia
julia> ForwardDiff.gradient(x -> sum(abs2, unique(x)), [1,2,1,2,3]) |> println
[2, 4, 0, 0, 6]  # before
[2, 4, 2, 4, 6]  # after

julia> hash(ForwardDiff.Dual(1,0)) === hash(ForwardDiff.Dual(1,2,3))
true  # unchanged
```
Maybe there is some way to avoid this?

Xref https://github.com/FluxML/Zygote.jl/issues/1053 (Zygote does not like `unique`)

stelmo:
Chiming in here, this patch solved [this issue for me that involved sparse arrays](https://discourse.julialang.org/t/incorrect-jacobian-using-forwarddiff-and-sparse-functions/79901). I would offer to help, but unfortunately I just don't know enough about auto diff :/

mcabbott:
Any help with making the tests pass would be welcome. They test many cases in a deeply nested set of loops, and it's a bit confusing to know which cases are meant to change, which are accidentally passing with certain random numbers, etc.

test/runtests.jl
```diff
@@ -1,39 +1,57 @@
...
-if VERSION >= v"1.5-"
-    println("Testing allocations...")
-    t = @elapsed include("AllocationsTest.jl")
-    println("done (took $t seconds).")
-end
+using ForwardDiff, Test, Random
+
+SEED = trunc(Int, time())
+println("##### Random.seed!($SEED), on VERSION == $VERSION")
+Random.seed!(SEED)
```
mcabbott, line 5: Previously it tested with a random global seed.

What was confusing is that sometimes this gives PRIMAL == PRIMAL2 (for integers), and sometimes not. It would be clearer to deliberately test both cases. But, for now, I just made it print the seed used, so that you can lock it to debug locally. 

Fixing the seed completely would test only one case per version of Julia.

mcabbott (commented):
Maybe I fixed the tests at last.

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "ForwardDiff"
 uuid = "f6369f11-7733-5829-9624-2563aa707210"
-version = "0.10.26"
+version = "0.11.0"
```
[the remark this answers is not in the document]
mcabbott: The downside is that 290 packages will all have to update their bounds:

https://juliahub.com/ui/Packages/ForwardDiff/k0ETY/0.10.28?page=2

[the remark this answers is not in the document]
KristofferC: If this is done, it should be 1.0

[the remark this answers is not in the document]
devmotion: Maybe it would also be good to enable nan safe mode by default.

test/DualTest.jl
```diff
@@ -219,6 +226,7 @@ for N in (0,3), M in (0,4), V in (Int, Float32)
 
     # Predicates #
     #------------#
+    @testset "Predicates" begin
 
     @test ForwardDiff.isconstant(zero(FDNUM))
     @test ForwardDiff.isconstant(one(FDNUM))
```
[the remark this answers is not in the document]
mcabbott: Last commit removed these.

But the rest seems to have rotted already, thanks to new test for complex numbers, perhaps?

mcabbott:
Latest commit proposes to make this a bugfix, since apparently it's what the authors of all the above issues expected, and making a breaking release sounds like it'll take another year. 

However, tests have already rotted, probably due to something in https://github.com/JuliaDiff/ForwardDiff.jl/pull/577/files. I have never met a package whose tests are more painful to work on.

test/DualTest.jl
```diff
@@ -499,10 +521,14 @@ for N in (0,3), M in (0,4), V in (Int, Float32)
                     else
                         @test dx isa Complex{<:Dual{TestTag()}}
                         @test dy isa Complex{<:Dual{TestTag()}}
-                        @test real(value(dx)) == real(actualval)
-                        @test real(value(dy)) == real(actualval)
-                        @test imag(value(dx)) == imag(actualval)
-                        @test imag(value(dy)) == imag(actualval)
```
mcabbott, line 505: These tests use method `value(x) = x`.

But they look like they are written to use a method `value(::Complex{<:Dual})`, which doesn't exist.

devmotion: Good catch, but it actually did the correct thing nevertheless - only changing the definition of `==` in this PR brakes it :stuck_out_tongue: `value(x::Complex{<:Dual}) = x`, so it's a no-op, and then `real` and `imag` just extract the different duals. Currently, then `somedual == somefloat` just checks `value(some_dual) == somefloat` (as intended in this test) but this PR change it to `value(some_dual) == somefloat && iszero(partials(some_dual))` which, of course, is not true in general and causes test errors.

mcabbott: So I discovered. The question is whether there ought to be `value(::Complex{<:Dual})`, probably yes, but probably not this PR.

stelmo:
Hi there, will this make it into master before 1.0?

src/dual.jl
```diff
@@ -384,13 +384,35 @@ for pred in UNARY_PREDICATES
     @eval Base.$(pred)(d::Dual) = $(pred)(value(d))
 end
 
-for pred in BINARY_PREDICATES
+# BINARY_PREDICATES = Symbol[:isequal, :isless, :<, :>, :(==), :(!=), :(<=), :(>=)] # prelude.jl
+
+Base.iszero(x::Dual) = iszero(value(x)) && iszero(partials(x))
```
devmotion, line 389: Maybe add a comment that this is just a more efficient implementation and consistent with `x == zero(x)` since `zero(x) = Dual(zero(value(x)), zero(partials(x)))` and `(==)(::Dual, ::Dual)` as defined below?

src/dual.jl
```diff
@@ -384,13 +384,35 @@ for pred in UNARY_PREDICATES
...
+        )
+    end
+end
+
+@define_binary_dual_op(
+    Base.:(!=),
+    (!=)(value(x), value(y)) || (!=)(partials(x), partials(y)),
+    (!=)(value(x), y)        || !iszero(partials(x)),
+    (!=)(x, value(y))        || !iszero(partials(y)),
+)
```
devmotion, line 407: Is this needed? Isn't this covered by the default definition of `!=` (https://github.com/JuliaLang/julia/blob/742b9abb4dd4621b667ec5bb3434b8b3602f96fd/base/operators.jl#L278) and the definition of `==` above? If it's needed (ambiguity issues?), then it might also be helpful to add a short comment that this just ensures that `!=(x, y) = !(x == y)`.

src/dual.jl
```diff
@@ -384,13 +384,35 @@ for pred in UNARY_PREDICATES
...
+
+for pred in [:isless, :<, :>, :(<=), :(>=)]
```
devmotion, line 409: Maybe change `BINARY_PREDICATES` to
```julia
const BINARY_PREDICATES = [:isless, :<, :>, :(<=), :(>=)]
```
and use here
```suggestion
for pred in BINARY_PREDICATES
```
? From a quick look at the source code it seems this is the only place where `BINARY_PREDICATES` was used.

devmotion (commented):
I think this change is reasonable.

Should some of the unary definitions be changed as well? In particular I am thinking about `isinteger` which is contained in `UNARY_PREDICATES` and hence leads to the definiton
```julia
isinteger(d::Dual) = isinteger(value(d))
```
However, that means that not every dual `d` with `isinteger(d)` can be converted to an `Integer` since we error if not `all(iszero, partials(d))`:
https://github.com/JuliaDiff/ForwardDiff.jl/blob/78c73afd9a21593daf54f61c7d0db67130cf29e1/src/dual.jl#L365-L368
I think the error is consistent with the modified definition of `==` in this PR since I guess we want `d == Integer(d)` whenever `Integer(d)` is defined/allowed. Hence probably `isinteger` should be changed?

src/dual.jl
mcabbott: The original code defines all of `[:isequal, :isless, :<, :>, :(==), :(!=), :(<=), :(>=)]`, even though some could call others via Base's definitions. Why, I'm not sure, maybe there were efficiency or ambiguity reasons. I guess I decided not to change that, to touch fewer things.

src/dual.jl
mcabbott: Seems a confusing name for a subset of them. Could be deleted. I didn't just in case someone has built on it.

src/dual.jl
devmotion: Yes, I assumed there might be a reason. Maybe just add a short comment explaining the reasoning behind this definition (i.e., it's just `!=(x, y) = !(x == y)`)?

src/dual.jl
devmotion: I don't mind, I just think it's confusing to keep `BINARY_PREDICATES` if it is not used anymore (I assumed you might just have forgotten to remove it). I don't think it would be too bad to reduce the number of functions in `BINARY_PREDICATES` - `UNARY_PREDICATES` etc. does also not include all possible unary functions but only those for which `@define_unary...` can be used.

mcabbott:
```suggestion
for pred in [:isless, :<, :>, :(<=), :(>=)]  # note that this list, and :(!=), isn't a minimal set, as Base defines some in terms of others. Left intact by #481
```

mcabbott:
The integer complaint is this, which is unchanged by this PR:
```julia

julia> isinteger(Dual(1.0, 0.2))  # is this wrong?
true

julia> convert(Int, Dual(1.0, 0.2))
ERROR: InexactError: Int(Int64, Dual{Nothing}(1.0,0.2))

julia> convert(Int, Dual(3.0, 0.0))
3
```
I'm not too sure what the right policy is here, has this ever come up in the wild? 

I don't think we should make quick arguments from apparent consistency. And I don't want to add anything more to this PR.

devmotion:
I'm sorry, I think my comment was a bit unclear. The main reason for why I think one should consider changing `isinteger` (and eg. `iseven` and `isodd`) is not to fix that `isinteger`/`Integer` inconsistency. But rather that

- these functions create also measure-zero branches that this PR tries to fix for other functions such as `==`,
- this PR introduces an inconsistency: Currently `isinteger(x::Dual)` is treated in the same way as the (slightly hypothetical) `x == typemin(T) || x == (typemin(T) + 1) || ... || x == typemax(T)` where `T = typeof(Integer(zero(x)))`, but with this PR they would be different. The first approach would neglect the partials whereas the second one would demand that they are zero.

Initially I was worried that these additional changes would be too far-fetched or possibly breaking. Only then I noticed the existing inconsistency with `isinteger`/`Integer` - which was somewhat relieving since I think it indicates that it is fine to change to `isinteger` (and other functions) as the current implementation seems to have problems that would be fixed by such a change.

mcabbott:
This PR is narrower than that, it fixes only `==` (and `isequal`) not all possible measure-zero branches. To start somewhere, and because equality is what seem to cause problems in the wild. 

It's possible that other changes should be considered, in follow-ups. `_isinteger(x) = x == trunc(Int, x)` is a plausible definition, which at present disagrees with `isinteger`. #480 also mentions `>`, which this PR does not touch, allowing this:
```
julia> x = ForwardDiff.Dual(1.0, 0.2)

julia> x > 1
false

julia> x < 1
false

julia> x == 1  # this PR
false
```

mcabbott:
Shall we merge this, before it goes stale again?

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "ForwardDiff"
 uuid = "f6369f11-7733-5829-9624-2563aa707210"
-version = "0.10.32"
+version = "0.10.30"
```
KristofferC, line 3:
```suggestion
version = "0.10.33"
```

One reason not to change version in PRs :P

mcabbott:
Thanks!

Test failures on master appear to be cause by changes in random numbers, now that 1 means 1.8? This (copied from CI logs) would pass with slightly larger tolerance:
```
julia> @test [1.7010290555795968e-6, 1.063143159737248e-6, 8.505145277897984e-7] ≈ [1.7010237080913283e-6, 1.0631398175570802e-6, 8.505118540456641e-7] rtol=1.0e-5
Test Passed

julia> @test [1.7010290555795968e-6, 1.063143159737248e-6, 8.505145277897984e-7] ≈ [1.7010237080913283e-6, 1.0631398175570802e-6, 8.505118540456641e-7] rtol=1.0e-6
Test Failed 
```

PR #489, opened 2020-12-23
Friendly errors on array vs. scalar mistakes

mcabbott:
The main purpose here is to change this:
```julia
julia> ForwardDiff.gradient(identity, rand(3))
ERROR: MethodError: no method matching extract_gradient!(::Type{ForwardDiff.Tag{typeof(identity), Float64}}, ::Vector{Vector{ForwardDiff.Dual{ForwardDiff.Tag{typeof(identity), Float64}, Float64, 3}}}, ::Vector{ForwardDiff.Dual{ForwardDiff.Tag{typeof(identity), Float64}, Float64, 3}})
Closest candidates are:
  extract_gradient!(::Type{T}, ::AbstractArray, ::ForwardDiff.Dual) where T at /Users/me/.julia/packages/ForwardDiff/qTmqf/src/gradient.jl:79
  extract_gradient!(::Type{T}, ::AbstractArray, ::Real) where T at /Users/me/.julia/packages/ForwardDiff/qTmqf/src/gradient.jl:78
  extract_gradient!(::Type{T}, ::DiffResults.DiffResult, ::ForwardDiff.Dual) where T at /Users/me/.julia/packages/ForwardDiff/qTmqf/src/gradient.jl:72
  ...
```
into this:
```julia
julia> ForwardDiff.gradient(identity, rand(3))  # with this PR
ERROR: ArgumentError: gradient(f,x) expects that f(x) be a real number. Perhaps you meant jacobian(f,x)?
```
since confusing these functions seems like a common mistake. For precedent:
```julia
julia> Zygote.gradient(identity, ones(3))
ERROR: Output should be scalar; gradients are not defined for output [1.0, 1.0, 1.0]
```
While I was doing this, I also made friendly errors for some wrong input types, as those also seem easy to confuse:
```julia
julia> ForwardDiff.gradient(identity, 3)  # with this PR
ERROR: ArgumentError: gradient(f,x) expects that x is an array. Perhaps you meant derivative(f,x)?
...

julia> Zygote.gradient(identity, 3)
(1,)
```
Although the error in this case was at least a MethodError of a function you called. And perhaps adding methods just to provide errors is poor style? Happy to remove these but perhaps worth discussing.
```julia
julia> ForwardDiff.gradient(identity, 3)
ERROR: MethodError: no method matching gradient(::typeof(identity), ::Int64)
Closest candidates are:
  gradient(::Any, ::StaticArrays.StaticArray) at /Users/me/.julia/packages/ForwardDiff/qTmqf/src/gradient.jl:44
  gradient(::Any, ::StaticArrays.StaticArray, ::ForwardDiff.GradientConfig) at /Users/me/.julia/packages/ForwardDiff/qTmqf/src/gradient.jl:45
  gradient(::Any, ::StaticArrays.StaticArray, ::ForwardDiff.GradientConfig, ::Val) at /Users/me/.julia/packages/ForwardDiff/qTmqf/src/gradient.jl:46
  ...
 ```

### DIFF
--- a/src/derivative.jl
+++ b/src/derivative.jl
@@ -69,7 +69,7 @@ Set `check` to `Val{false}()` to disable tag checking. This can lead to perturba
     return result
 end
 
-derivative(f, x::AbstractArray) = throw(ArgumentError("derivative(f,x) expects that x is a real number. Perhaps you meant gradient(f,x)?"))
+derivative(f, x::AbstractArray) = throw(DimensionMismatch("derivative(f, x) expects that x is a real number. Perhaps you meant gradient(f, x)?"))
 
 #####################
 # result extraction #
--- a/src/derivative.jl
+++ b/src/derivative.jl
@@ -69,6 +69,8 @@ Set `check` to `Val{false}()` to disable tag checking. This can lead to perturba
     return result
 end
 
+derivative(f, x::AbstractArray) = throw(ArgumentError("derivative(f,x) expects that x is a real number. Perhaps you meant gradient(f,x)?"))
+
 #####################
 # result extraction #
 #####################
--- a/src/gradient.jl
+++ b/src/gradient.jl
@@ -91,12 +91,18 @@ function extract_gradient_chunk!(::Type{T}, result::DiffResult, dual, index, chu
     return result
 end
 
+extract_gradient_chunk!(::Type, result, dual::AbstractArray, index, chunksize) = throw(GRAD_ERROR)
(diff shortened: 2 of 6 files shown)


ChrisRackauckas (approved):
Major QoL improvement. I like it. But should those be ArgumentErrors? I guess so because it's errors with the `f` argument, but it seems a little odd.

mcabbott:
Maybe `DimensionMismatch` is closer? 

Needs tests, BTW, but I thought better to discuss first.

KristofferC:
Nice! Would be good with some tests though to make sure these code paths are hit.

Issue #497, opened 2021-01-17
Derivative through sincospi gives StackOverflowError


mchitre:
On Julia 1.6.0-beta1

MWE:
```julia-repl
julia> f(x) = sum(sincospi(x));
julia> ForwardDiff.derivative(f, 0.0)
ERROR: StackOverflowError:
Stacktrace:
 [1] sincospi(x::ForwardDiff.Dual{ForwardDiff.Tag{typeof(f), Float64}, Float64, 1})
   @ Base.Math ./special/trig.jl:934
```

but this works:
```julia-repl
julia> mysincospi(x) = (sin(pi*x), cos(pi*x));
julia> f(x) = sum(mysincospi(x));
julia> ForwardDiff.derivative(f, 0.0)
3.141592653589793
```

andreasnoack:
These is probably a definition missing in https://github.com/JuliaDiff/DiffRules.jl/blob/master/src/rules.jl so it would be great if you could prepare a PR.

mchitre:
Will do shortly

mchitre:
Investigated further. The problem is that Julia Base (`base/special/trig.jl`) defines:
```julia
sincospi(x::Real) = sincospi(float(x))
```
which results in a self-call since `float(::Dual) isa Real` (but not an `AbstractFloat`, for which there is a specialized version defined).

The equivalent definition for `sincos()` is different in Julia Base and doesn't have this problem:
```julia
_sincos(x) = (sin(x), cos(x))
sincos(x) = _sincos(float(x))
```
It simplifies to the regular `sin()` and `cos()`, for which rules are already defined.

So my questions are:
1.  Should this be fixed by defining a rule in `DiffRules.jl`, or should it be fixed in Julia Base?
2. If it is to be fixed in `DiffRules.jl`, how does one define a rule for a function that returns a tuple? I could not find any other rule that needs to deal with this, and returning a tuple of derivatives from the rule doesn't seem to work. See below.

---
Returning a tuple from a diff rule does not work:

```julia
@define_diffrule Base.sincospi(x)             = :( (π * cospi($x), -π * sinpi($x))     )
```
fails:
```julia
julia> f1(x) = sum(sincospi(x));

julia> ForwardDiff.derivative(f1, 0.0)
ERROR: MethodError: no method matching *(::Tuple{Float64, Float64}, ::ForwardDiff.Partials{1, Float64})
Closest candidates are:
[...]

PR #499, opened 2021-01-18
Fix for derivative through sincospi gives StackOverflowError

mchitre:
Fixes https://github.com/JuliaDiff/ForwardDiff.jl/issues/497

Adds:
- [x] Specialized version of `Base.sincospi(::Dual)` to avoid stackoverflow error
- [x] Test case for the specialized `Base.sincospi(::Dual)`

NOTE: Julia 1.6 required

### DIFF
--- a/src/dual.jl
+++ b/src/dual.jl
@@ -612,9 +612,11 @@ end
 # sincospi #
 #----------#
 
-@inline function Base.sincospi(d::Dual{T}) where T
-    sd, cd = sincospi(value(d))
-    return (Dual{T}(sd, cd * π * partials(d)), Dual{T}(cd, -sd * π * partials(d)))
+if VERSION >= v"1.6-"
+    @inline function Base.sincospi(d::Dual{T}) where T
+        sd, cd = sincospi(value(d))
+        return (Dual{T}(sd, cd * π * partials(d)), Dual{T}(cd, -sd * π * partials(d)))
+    end
 end
 
 ###################
--- a/src/dual.jl
+++ b/src/dual.jl
@@ -612,7 +612,7 @@ end
 # sincospi #
 #----------#
 
-if VERSION >= v"1.6-"
+if VERSION >= v"1.6.0-DEV.292"
     @inline function Base.sincospi(d::Dual{T}) where T
         sd, cd = sincospi(value(d))
         return (Dual{T}(sd, cd * π * partials(d)), Dual{T}(cd, -sd * π * partials(d)))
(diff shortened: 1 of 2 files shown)


src/dual.jl
```diff
@@ -609,6 +609,14 @@ end
     return (Dual{T}(sd, cd * partials(d)), Dual{T}(cd, -sd * partials(d)))
 end
 
+# sincospi #
+#----------#
+
+@inline function Base.sincospi(d::Dual{T}) where T
+    sd, cd = sincospi(value(d))
+    return (Dual{T}(sd, cd * π * partials(d)), Dual{T}(cd, -sd * π * partials(d)))
```
charleskawczynski, line 617:
```suggestion
    return (Dual{T}(sd, cd * T(π) * partials(d)), Dual{T}(cd, -sd * T(π) * partials(d)))
```
Do we need to wrap in `T` to be type stable?

andreasnoack: The existing version should be fine since the type of `cd * π` will be decided by `cd`.

charleskawczynski: Ah, that's awesome. So there are special promotion rules with literals?

andreasnoack: It's special `Irrational` magic.
```julia
julia> typeof(π)
Irrational{:π}
```
The precision of them in binary operations is determined by the other argument.

andreasnoack:
Why is it that 1.6 is required?

mchitre:
Because `sincospi()` is introduced in 1.6.

src/dual.jl
```diff
@@ -609,6 +609,16 @@ end
     return (Dual{T}(sd, cd * partials(d)), Dual{T}(cd, -sd * partials(d)))
 end
 
+# sincospi #
+#----------#
+
+if VERSION >= v"1.6-"
```
andreasnoack, line 615:
```suggestion
if VERSION >= v"1.6.0-DEV.292"
```

test/DualTest.jl
```diff
@@ -449,6 +449,10 @@ for N in (0,3), M in (0,4), V in (Int, Float32)
 
     @test all(map(dual_isapprox, ForwardDiff.sincos(FDNUM), (sin(FDNUM), cos(FDNUM))))
 
+    if VERSION >= v"1.6-"
```
andreasnoack, line 452:
```suggestion
    if VERSION >= v"1.6.0-DEV.292"
```

PR #503, opened 2021-01-29
Overload div for Dual

sethaxen:
With this, `div` as well as functions that call it, like `fld` and `cld`, no longer error when passed a `Dual`.

### DIFF
--- a/src/dual.jl
+++ b/src/dual.jl
@@ -279,6 +279,12 @@ Base.trunc(d::Dual) = trunc(value(d))
 Base.round(::Type{R}, d::Dual) where {R<:Real} = round(R, value(d))
 Base.round(d::Dual) = round(value(d))
 
+if VERSION ≥ v"1.4"
+    Base.div(x::Dual, y::Dual, r::RoundingMode) = div(value(x), value(y), r)
+else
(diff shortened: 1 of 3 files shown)


oxinabox (approved):

src/dual.jl
```diff
@@ -279,6 +279,12 @@ Base.trunc(d::Dual) = trunc(value(d))
 Base.round(::Type{R}, d::Dual) where {R<:Real} = round(R, value(d))
 Base.round(d::Dual) = round(value(d))
 
+if VERSION ≥ v"1.4"
+    Base.div(x::Dual, y::Dual, r::RoundingMode) = div(value(x), value(y), r)
+else
+    Base.div(x::Dual, y::Dual) = div(value(x), value(y))
+end
```
oxinabox, line 286: Do we also need overloads for `div` when one of the arguments is not a Dual?
e.g. for if literals are involved?

sethaxen: What is a literal?

oxinabox: e.g. if the code is `div(10, x)`

sethaxen: Ah okay, yes that is implicitly supported via promotion. I just added an explicit test.

PR #508, opened 2021-03-03
Int(::Dual) and Integer(::Dual)

dlfivefifty:
This adds convert routines for integers. I need this to fix the following bug:
```julia
julia> length(UnitRange(Dual(1.0), Dual(2.0)))
ERROR: MethodError: no method matching Integer(::Dual{Nothing, Float64, 0})
Closest candidates are:
  (::Type{T})(::T) where T<:Number at boot.jl:760
  (::Type{T})(::AbstractChar) where T<:Union{AbstractChar, Number} at char.jl:50
  (::Type{T})(::Base.TwicePrecision) where T<:Number at twiceprecision.jl:243
  ...
Stacktrace:
 [1] unsafe_length(r::UnitRange{Dual{Nothing, Float64, 0}})
   @ Base ./range.jl:559
 [2] length(r::UnitRange{Dual{Nothing, Float64, 0}})
   @ Base ./range.jl:561
 [3] top-level scope
   @ REPL[4]:1
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ForwardDiff"
 uuid = "f6369f11-7733-5829-9624-2563aa707210"
-version = "0.10.18"
+version = "0.10.19"
 
 [deps]
 CommonSubexpressions = "bbf7d656-a473-5ed7-a52c-81e309532950"
--- a/src/dual.jl
+++ b/src/dual.jl
@@ -305,8 +305,14 @@ end
 @inline Base.one(d::Dual) = one(typeof(d))
 @inline Base.one(::Type{Dual{T,V,N}}) where {T,V,N} = Dual{T}(one(V), zero(Partials{N,V}))
 
-@inline Base.Int(d::Dual) = Int(value(d))
-@inline Base.Integer(d::Dual) = Integer(value(d))
+@inline function Base.Int(d::Dual)
+    all(iszero, partials(d)) || throw(InexactError(:Int, Int, d))
+    Int(value(d))
+end
+@inline function Base.Integer(d::Dual)
+    all(iszero, partials(d)) || throw(InexactError(:Integer, Integer, d))
+    Integer(value(d))
+end
 
 @inline Random.rand(rng::AbstractRNG, d::Dual) = rand(rng, value(d))
 @inline Random.rand(::Type{Dual{T,V,N}}) where {T,V,N} = Dual{T}(rand(V), zero(Partials{N,V}))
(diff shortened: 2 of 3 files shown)


KristofferC:
I think this should only work if the dual part is zero (cf the behavior of complex numbers).

dlfivefifty:
I think I agree:
```julia
julia> Integer(nextfloat(1.0))
ERROR: InexactError: Int64(1.0000000000000002)
Stacktrace:
 [1] Int64
   @ ./float.jl:723 [inlined]
 [2] Integer(x::Float64)
   @ Core ./boot.jl:786
 [3] top-level scope
   @ REPL[1]:1
```
But in that case I should also overload `unsafe_length(::UnitRange{<:Dual})` so that 
```julia
julia> UnitRange(Dual(1,1), Dual(2,0)) |> collect
2-element Vector{Dual{Nothing, Int64, 1}}:
 Dual{Nothing}(1,1)
 Dual{Nothing}(2,1)
```
works.

dlfivefifty:
Ha! We're in luck, because `partials(step(r)) == partials(first(r))` the default code for `unsafe_length(::UnitRange)` just works as is..

ranocha:
Bump. This looks like a useful addition to me :+1:

dlfivefifty:
Can this be merged? Just ran into it, again

dlfivefifty:
The test failure looks unrelated...

KristofferC:
I agree that the test failures don't look directly related to this but when I look at the CI run for earlier commits on master (https://github.com/JuliaDiff/ForwardDiff.jl/commits/master) they all seem green so it is a bit surprising why this would suddenly start happening.

KristofferC:
CI error also happens in https://github.com/JuliaDiff/ForwardDiff.jl/pull/527. It seems to be something with the StaticArray tests so maybe it is a new version of StaticArrays that caused it?

KristofferC:
I bisected and https://github.com/JuliaArrays/StaticArrays.jl/pull/907 is when the ForwardDiff test failure was introduced.

dlfivefifty:
Note the values depend on the sample in `x = rand(3,3)`. If I do `Random.seed!(0)` I get the slightly different result:
```julia
julia> @test DiffResults.value(result1) == DiffResults.value(result)
Test Failed at REPL[39]:1
  Expression: DiffResults.value(result1) == DiffResults.value(result)
   Evaluated: 0.0 == 1.297445353195765e-6
```

dlfivefifty:
I'm debugging this now. FYI I believe `vector_mode_dual_eval` should be renamed `vector_mode_dual_eval!`

https://github.com/JuliaDiff/ForwardDiff.jl/blob/909976d719fdbd5fec91a159c6e2d808c45a770f/src/apiutils.jl#L40

PR #513, opened 2021-03-23
eigen and eigvals for Symmetric, Hermitian, and SymTridiagonal

dlfivefifty:
I still need to add more tests but wanted to make sure that this is the right repos for this change. I suspect its too complicated for DiffRules.jl


andreasnoack:
This is great to have and I agree that it's a better fit for this package than DiffRules.

PR #523, opened 2021-06-15
Overload floatmin, floatmax

dlfivefifty:
I need these for differentiating code with adaptive convergence checks.


dlfivefifty:
I'm clearly losing my mind as I made the exact same fix in #525

PR #525, opened 2021-06-22
Overload floatmin, floatmax

dlfivefifty:
This is needed to support adaptivity with infinite arrays. 

For a some-what esoteric example, with this PR we can differentiate a renormalized Bessel function from it's 3-term recurrence:
```julia
julia> J̃ = z -> (LazyBandedMatrices.SymTridiagonal(-2*(0:∞)/z, ones(∞)) \ [1; zeros(∞)])[1]
#9 (generic function with 1 method)

julia> J̃(0.5) ≈ besselj(0,0.5)/besselj(1,0.5)
true

julia> derivative(J̃,0.5) ≈ derivative(x -> besselj(0,x)/besselj(1,x), 0.5)
true
```

More interesting examples are adaptively solving ODEs with parameters via spectral methods.

### DIFF
--- a/src/dual.jl
+++ b/src/dual.jl
@@ -684,7 +684,7 @@ function Base.show(io::IO, d::Dual{T,V,N}) where {T,V,N}
 end
 
 for op in (:(Base.typemin), :(Base.typemax), :(Base.floatmin), :(Base.floatmax))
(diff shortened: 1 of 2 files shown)


mcabbott (approved):
Needs tests, otherwise seems fine.

dlfivefifty:
I've added tests

dlfivefifty:
Can we merge, please?

PR #528, opened 2021-07-01
WIP vector_mode_dual_eval -> vector_mode_dual_eval!

dlfivefifty:
Still need to change elsewhere, this is just a reminder to rename `vector_mode_dual_eval` to make it clear it mutates its inputs (was confusing when debugging).

### DIFF
--- a/src/jacobian.jl
+++ b/src/jacobian.jl
@@ -144,7 +144,7 @@ reshape_jacobian(result::DiffResult, ydual, xdual) = reshape_jacobian(DiffResult
 ###############
 
 function vector_mode_jacobian(f::F, x, cfg::JacobianConfig{T,V,N}) where {F,T,V,N}
-    ydual = vector_mode_dual_eval(f, x, cfg)
+    ydual = vector_mode_dual_eval!(f, x, cfg)
     ydual isa AbstractArray || throw(JACOBIAN_ERROR)
     result = similar(ydual, valtype(eltype(ydual)), length(ydual), N)
     extract_jacobian!(T, result, ydual, N)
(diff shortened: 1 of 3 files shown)


KristofferC:
Bump

dlfivefifty:
The 1.0 failure seems unrelated

KristofferC:
I guess you would kind of think that this modifies `x` and not `cfg`. Perhaps the order should be swapped? What do you think?

dlfivefifty:
Agreed (though only subject to the "kind of" qualifier... I think anything that mutates any of the arguments should use `!`).

I'll change it

src/jacobian.jl
```diff
@@ -144,7 +144,7 @@ reshape_jacobian(result::DiffResult, ydual, xdual) = reshape_jacobian(DiffResult
 ###############
 
 function vector_mode_jacobian(f::F, x, cfg::JacobianConfig{T,V,N}) where {F,T,V,N}
-    ydual = vector_mode_dual_eval!(f, x, cfg)
+    ydual = vector_mode_dual_eval!(cfg, f, x)
```
KristofferC, line 147: even in `!` functions I think you put the function first to allow for `do` syntax. So it would be `f, cfg, x`.

dlfivefifty: You are really trying to make me regret this...

ok I've changed it again

dlfivefifty:
Hmm now the failure is on Julia 1, but also seems unrelated to this PR...

KristofferC:
Yeah, there is some numerical thing we need to investigate at some point..

PR #530, opened 2021-07-04
Ignore rules with more than two arguments

mcabbott:
DiffRules.jl supports N-ary rules. These aren't automatically handled by ForwardDiff, but I don't think their existence need cause this package not to be usable.

If the two packages must be coupled more tightly than via Semver, then perhaps a test checking that all rules have Dual definitions would be appropriate. And produce a test failure, while allowing use of other, unaffected, parts of the package. 

Closes #529, I think. 

Edit to add context:
- DiffRules v1.1 and v1.2 contain 3-ary rules. These were removed from 1.2.1 as a temporary measure, and the General registry changed so that existing versions of ForwardDiff are marked incompatible with 1.1 and 1.2.
- The present Project.toml declares compatibility with DiffRules 1.0. I believe that tagging a new release without this PR will mean it is marked compatible with all 1.0 versions, and thus combinations which crash on launch will be legal according to the registry.

PR #534, opened 2021-07-15
Avoid conj override message

dlfivefifty:
At the moment we have
```julia
julia> using ForwardDiff
[ Info: Precompiling ForwardDiff [f6369f11-7733-5829-9624-2563aa707210]
WARNING: Method definition conj(ForwardDiff.Dual{T, V, N} where N where V) where {T} in module ForwardDiff at /Users/sheehanolver/Projects/ForwardDiff.jl/src/dual.jl:390 overwritten at /Users/sheehanolver/Projects/ForwardDiff.jl/src/dual.jl:201.
  ** incremental compilation may be fatally broken for this module **
```
due to `conj` being added to DiffRules.jl


src/dual.jl
```diff
@@ -387,8 +387,6 @@ Base.AbstractFloat(d::Dual{T,V,N}) where {T,V,N} = convert(Dual{T,promote_type(V
 # General Mathematical Operations #
 ###################################
 
-@inline Base.conj(d::Dual) = d
```
KristofferC, line 390: If we don't want to drop all the previous DiffRules compact, a `hasmethod` check could be used instead?

dlfivefifty: Do we care about maintaining compatibility with old versions of DiffRules? My personal preference is to always drop old versions as it means less maintenance headaches.

KristofferC: If it is so easy, like in this case, might as well, or?

dlfivefifty: Fine, can you add the change as  a suggestion? (I'm not 100% sure how to do it in a way that's precompile-compatible.)

KristofferC: Should just be

```jl
if !hasmethod(conj, Tuple{ForwardDiff.Dual})
    @inline Base.conj(d::Dual) = d
end
```

Not sure how to add that as a suggestion.

dlfivefifty: Use "```suggestion"

```suggestion
if !hasmethod(conj, Tuple{ForwardDiff.Dual})
    @inline Base.conj(d::Dual) = d
end
```

KristofferC (approved):

mcabbott:
Closes #532.

The other option would be to remove that from DiffRules. DiffRules mostly has holomorphic functions, although it does have `abs`.

src/dual.jl
mcabbott: Is `hasmethod` going to work here? The method you are checking for is defined in just below this in the file.

mcabbott: And, even with this mechanism, I think it needs new bounds in Project.toml before tagging. 

In the registry all past versions are marked incompatible with DiffRules 1.1 & 1.2, see https://github.com/JuliaRegistries/General/pull/40252 and friends. The next tagged version should probably accept `<= 1.02 + >= 1.21`.

dlfivefifty: I'm not sure what you are suggestion. Perhaps make a ```suggestion ?

mcabbott: I made a PR, weeks ago. But read my comments carefully, there are two concerns. One is whether this `hasmethod` actually works as claimed, did you test it with DiffRules 0.9 say? The other is that tagging this would re-declare compatibility of this package with versions of DiffRules which contain 3-ary rules, which cause it to crash, #530.

KristofferC: Let's go with the compat restricted version then. I don't think it will be any big problem in practice.

PR #535, opened 2021-07-16
fix float(::Dual) and add float(::Type{<:Dual})

### LINKED ISSUE #362: `ForwardDiff.derivative(float, 1)` returns `Float16(1.0)`
Why the conversion to `Float16`? This only occurs with integer `1`. 

`ForwardDiff.derivative(float, 1.0)` works fine.

### LINKED ISSUE #492: float(::Dual) disagrees with Base
I was surprised by this:
```
using ForwardDiff: Dual, value
x1 = Dual.(1:4, 2:5, 3:6)
value.(float.(x1)) # Vector{Float16}
float.(value.(x1)) # Vector{Float64}
```
While one could argue that Julia should have different promotion / widening rules, it seems hard to argue that Dual types should disagree. This example caught me when trying to wrap a function which only accepts Float32 & Float64, as seems fairly common for C libraries. 

Edit -- xref #419, about AbstractFloat. 

And #362, about Int -> Float16. Example from [here](https://discourse.julialang.org/t/what-functions-might-break-forwarddiff-jl/64694/8) is `ForwardDiff.derivative(first∘sincos, 1) isa Float16`

stevengj:
Fixes #492, fixes #362, closes #419.

The `Base.float` function in base supports numeric types, not just values:
```
julia> float(Complex{Int})
ComplexF64 (alias for Complex{Float64})
```
and by exploiting this we obtain floating-point promotion consistent with `Base`.

### DIFF
--- a/src/dual.jl
+++ b/src/dual.jl
@@ -380,8 +380,8 @@ Base.convert(::Type{Dual{T,V,N}}, x) where {T,V,N} = Dual{T}(convert(V, x), zero
 Base.convert(::Type{Dual{T,V,N}}, x::Number) where {T,V,N} = Dual{T}(convert(V, x), zero(Partials{N,V}))
 Base.convert(::Type{D}, d::D) where {D<:Dual} = d
 
-Base.float(d::Dual{T,V,N}) where {T,V,N} = convert(Dual{T,promote_type(V, Float16),N}, d)
-Base.AbstractFloat(d::Dual{T,V,N}) where {T,V,N} = convert(Dual{T,promote_type(V, Float16),N}, d)
+Base.float(::Type{Dual{T,V,N}}) where {T,V,N} = Dual{T,float(V),N}
+Base.float(d::Dual) = convert(float(typeof(d)), d)
 
 ###################################
 # General Mathematical Operations #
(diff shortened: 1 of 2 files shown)


stevengj:
The 
```
Test threw exception
  DomainError with -1.1858610145546527e18:
  sqrt will only return a complex result if called with a complex argument. Try sqrt(Complex(x)).
```
makes no sense to me.  I actually got a similar error at some point locally, and then it mysteriously disappeared when I re-ran the test — makes me think it is a bug in Julia itself.

KristofferC:
I think I recall seeing something similar to that. It clearly isn't related to the PR.

mcabbott:
Re `NESTED_FDNUM ^ PRIMAL`, the rule for `^` attempts to avoid a `DomainError ` from `log`, and was changed recently. (Both 1.0 and 1 CI are DiffRules v1.2.1, the latest.) Is it possible this is somehow to blame? I don't see how though.

 https://github.com/JuliaDiff/DiffRules.jl/blob/master/src/rules.jl#L88

Note also that I think #530 needs to be merged before tagging anything.

stevengj:
Is this mergeable, then?  If not, what needs to be done?

dlfivefifty:
We also need `Base.big`. Can you add that here, or shall I do it in a separate PR?

PR #538, opened 2021-07-23
Add `FloatNN(::Dual)`

mzgubic:
Needed for Zygote on ChainRules 1.0 (only the `Float64` case strictly speaking)

<s>
Similar to the recently merged https://github.com/JuliaDiff/ForwardDiff.jl/pull/508/files

</s>

### DIFF
--- a/src/dual.jl
+++ b/src/dual.jl
@@ -314,6 +314,10 @@ end
     Integer(value(d))
 end
 
+for F in (:Float16, :Float32, :Float64)
+    @eval @inline Base.$F(d::Dual) = $F(value(d))
+end
+
 @inline Random.rand(rng::AbstractRNG, d::Dual) = rand(rng, value(d))
 @inline Random.rand(::Type{Dual{T,V,N}}) where {T,V,N} = Dual{T}(rand(V), zero(Partials{N,V}))
 @inline Random.rand(rng::AbstractRNG, ::Type{Dual{T,V,N}}) where {T,V,N} = Dual{T}(rand(rng, V), zero(Partials{N,V}))
--- a/src/dual.jl
+++ b/src/dual.jl
@@ -314,10 +314,6 @@ end
     Integer(value(d))
 end
 
-for F in (:Float16, :Float32, :Float64)
-    @eval @inline Base.$F(d::Dual{T,V,N}) where {T,V,N} = convert(Dual{T,promote_type(V, $F),N}, d)
-end
-
 @inline Random.rand(rng::AbstractRNG, d::Dual) = rand(rng, value(d))
 @inline Random.rand(::Type{Dual{T,V,N}}) where {T,V,N} = Dual{T}(rand(V), zero(Partials{N,V}))
 @inline Random.rand(rng::AbstractRNG, ::Type{Dual{T,V,N}}) where {T,V,N} = Dual{T}(rand(rng, V), zero(Partials{N,V}))
@@ -386,6 +382,10 @@ Base.convert(::Type{D}, d::D) where {D<:Dual} = d
 
 Base.float(d::Dual{T,V,N}) where {T,V,N} = convert(Dual{T,promote_type(V, Float16),N}, d)
 Base.AbstractFloat(d::Dual{T,V,N}) where {T,V,N} = convert(Dual{T,promote_type(V, Float16),N}, d)
+
+for F in (:Float16, :Float32, :Float64)
+    @eval @inline Base.$F(d::Dual{T,V,N}) where {T,V,N} = convert(Dual{T,promote_type(V, $F),N}, d)
+end
 
 ###################################
 # General Mathematical Operations #
--- a/src/dual.jl
+++ b/src/dual.jl
@@ -315,7 +315,7 @@ end
 end
 
 for F in (:Float16, :Float32, :Float64)
-    @eval @inline Base.$F(d::Dual) = $F(value(d))
+    @eval @inline Base.$F(d::Dual{T,V,N}) where {T,V,N} = convert(Dual{T,promote_type(V, $F),N}, d)
 end
 
 @inline Random.rand(rng::AbstractRNG, d::Dual) = rand(rng, value(d))
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ForwardDiff"
 uuid = "f6369f11-7733-5829-9624-2563aa707210"
-version = "0.10.19"
+version = "0.10.20"
 
 [deps]
 CommonSubexpressions = "bbf7d656-a473-5ed7-a52c-81e309532950"
--- a/test/DerivativeTest.jl
+++ b/test/DerivativeTest.jl
@@ -100,4 +100,8 @@ end
     @test_throws DimensionMismatch ForwardDiff.derivative(sum, fill(2pi, 3))
 end
 
+@testset "function that calls a floating point constructor" begin
+    @test ForwardDiff.derivative(x->2*Float32(x), 3) === 2f0
+end
+
 end # module
--- a/test/DualTest.jl
+++ b/test/DualTest.jl
@@ -519,9 +519,8 @@ end
 end
 
 @testset "Float" begin
-    x = Dual(1.0, 0)
     for F in (Float16, Float32, Float64)
-        @test F(1.0) == F(x)
+        @test F(Dual(3, 5)) == Dual(F(3), F(5))
     end
 end
   
--- a/test/DualTest.jl
+++ b/test/DualTest.jl
@@ -517,6 +517,13 @@ end
     @test length(UnitRange(Dual(1.5), Dual(3.5))) == 3
     @test length(UnitRange(Dual(1.5,1), Dual(3.5,3))) == 3
 end
+
+@testset "Float" begin
+    x = Dual(1.0, 0)
+    for F in (Float16, Float32, Float64)
+        @test F(1.0) == F(x)
+    end
+end
   
 if VERSION >= v"1.6.0-rc1"
     @testset "@printf" begin


src/dual.jl
```diff
@@ -314,6 +314,10 @@ end
     Integer(value(d))
 end
 
+for F in (:Float16, :Float32, :Float64)
+    @eval @inline Base.$F(d::Dual) = $F(value(d))
```
oxinabox, line 318:
```suggestion
    @eval @inline function Base.$F(d::Dual)
        all(iszero, partials(d)) || throw(InexactError($(nameof(F)), $F, d))
        $F(value(d))
    
```
Else we are just unexpectedly dropping the partials.


like https://github.com/JuliaDiff/ForwardDiff.jl/pull/508/files#diff-7f9840ac912342d25066b9e92a4f472a21750c5a88bf8b13232172d45bf844f4R313

oxinabox:
> Needed for Zygote on ChainRules 1.0 (only the Float64 case strictly speaking)

Why does Zygote need this?
This is pretty weird to need.
Feels like Zygote is in the wrong.

mzgubic:
It's the `ProjectTo` really that errors:
```julia
diagonal hessian: Error During Test at /Users/mzgubic/JuliaEnvs/Zygote.jl/test/utils.jl:22
  Got exception outside of a @test
  MethodError: no method matching Float64(::ForwardDiff.Dual{Nothing, Float64, 6})
  Closest candidates are:
    (::Type{T})(::Real, ::RoundingMode) where T<:AbstractFloat at rounding.jl:200
    (::Type{T})(::T) where T<:Number at boot.jl:760
    (::Type{T})(::AbstractChar) where T<:Union{AbstractChar, Number} at char.jl:50
    ...
  Stacktrace:
    [1] convert(#unused#::Type{Float64}, x::ForwardDiff.Dual{Nothing, Float64, 6})
      @ Base ./number.jl:7
    [2] (::ChainRulesCore.ProjectTo{Float64, NamedTuple{(), Tuple{}}})(dx::ForwardDiff.Dual{Nothing, Float64, 6})
      @ ChainRulesCore ~/JuliaEnvs/Zygote.jl/dev/ChainRulesCore/src/projection.jl:144
    [3] ^_pullback
      @ ~/JuliaEnvs/Zygote.jl/dev/ChainRules/src/rulesets/Base/fastmath_able.jl:172 [inlined]
```

oxinabox:
I suspect that the correct thing to do here is not to drop the partial part.
These are not actually the mathematical objects called dual numbers, which can simply be projected onto the reals.
They are the abstraction of forwards-mode AD, which is normally isomorphic to the dual number, but not this case I think.

I think the correct thing to do is to end up with more dual numbers.
Right now we have
```julia
julia> d = ForwardDiff.Dual(1,2)
Dual{Nothing}(1,2)

julia> AbstractFloat(d)
Dual{Nothing}(1.0,2.0)
```

So matching the behavour of `AbstractFloat` would be to convert the partials and the values to a `Float64` etc.
Then if Zygote really wants to drop the partials, it can by doing so explicitly.

src/dual.jl
oxinabox: @mzgubic  said in slack that this does not work, Zygote is still broken with this
```
julia>   dx, dy = diaghessian(f34, xs, y)
ERROR: InexactError: Float64(Dual{Nothing}(0.2883004137872829,-0.36786533559880646,-0.006843046852434196,-0.0,-0.0,-0.0,-0.0))
```

oxinabox: I can confirm that the following fixes Zygote

```suggestion
    @eval @inline Base.$F(d::Dual{T,V,N}) where {T,V,N} = convert(Dual{T,promote_type(V, $F),N}, d)
```

Code is based on what `AbstractFloat(::Dual)` currently does

oxinabox:
I am now more convinced this is correct.
It is not a conversation of the Dual to a Float64 (etc)
It is using a forward mode operator overloading AD to compute the derivative of the Float64 (etc) constructor.
Which is to apply `Float64` to the partial as well.

It is kinda gross that to do so it has to overload the constructor to return the `Dual{Float64}` type.
But that is operator overloading AD for you.

YingboMa (dismissed):

oxinabox:
@YingboMa  raised the good questions about how many invalidations this would cause.
It would make sense for this to be pretty invaldiating, since it would be quite possible for world-splitting to be causing it to depend on all uses of `Float64` returning a `Float64` (which I think it should but again operator overloading ADs are gross)

But checking with  SnoopCompile (and remembering to `--startup=no` this time)
```
julia> using SnoopCompileCore;

julia> invalidations = @snoopr begin 
       using ForwardDiff
       end
```
It seems 
ForwardDiff#master has 1108 invalidations
This PR has only 896 invalidations
I am not sure how exactly that is possible.

KristofferC:
This needs a test for when this is a problem for ForwardDiff. What bug does this fix? The test just verifies that the implementation does what the implementation does but it lacks a motivating example.

oxinabox:
> This needs a test for when this is a problem for ForwardDiff. What bug does this fix? 

Done. 
I have added code that would error before, but now returns a correct result.

mcabbott:
Late to the party, but this seems a bit odd. 

If I understand right what's happening is this:
```
julia> using ChainRulesCore, ForwardDiff

julia> p = ProjectTo(1.0)  # Float64 in forward pass
ProjectTo{Float64}()

julia> p(ForwardDiff.Dual(1.0, true, false))  # Dual in backward pass
ERROR: MethodError: no method matching Float64(::ForwardDiff.Dual{Nothing, Float64, 2})

julia> ProjectTo(ForwardDiff.Dual(1.0, true, false))  # this one we made not fussy
ProjectTo{Real}()
```
Could it be fixed elsewhere? The reason `p` converts to Float64 is mainly to protect Float32 from accidental promotion (or at least stop it propagating.) Maybe that method should only apply to AbstractFloat? 

Or be done through an intermediate function which means "convert the numeric type to Float64, but preserve the Dual / units / etc"?

Second, this PR doesn't seem to achieve that, since the FloatN constructor doesn't always change the base numeric type. And it leads to weird error messages:
```
julia> Float32(Dual(1.0, true, false)) |> dump
Dual{Nothing, Float64, 2}
  value: Float64 1.0
  partials: ForwardDiff.Partials{2, Float64}
    values: Tuple{Float64, Float64}
      1: Float64 1.0
      2: Float64 0.0

julia> rand(3)[1] = Dual(1.0, true, false)
ERROR: TypeError: in typeassert, expected Float64, got a value of type Dual{Nothing, Float64, 2}
Stacktrace:
 [1] setindex!(A::Vector{Float64}, x::Dual{Nothing, Float64, 2}, i1::Int64)
```

oxinabox:
> Second, this PR doesn't seem to achieve that, since the FloatN constructor doesn't always change the base numeric type. 

Right now it has some promotion stuff in there.
Because it's based off the one for AbstractFloat.
So it gives the hight of the current type and the requested one 
Possibly that could be removed?
I am not sure if it is required to make nested calls work.

> And it leads to weird error messages:

That error message seems fine to me.
What else would it say?
Same thing but in method error?


> Or be done through an intermediate function which means "convert the numeric type to Float64, but preserve the Dual / units / etc"?

Could be a thing.
Not sure it would be possible without ForwardDiff overloading something from ChainRulesCore.
But I guess we could have a fallback for other things that subtype `Real` of leave them alone, they probably know what they are doing.

I do still think this PR makes perfectly fine sense though.
It is operator overloading AD of the constructor.

mcabbott:
> Right now it has some promotion stuff in there.

If `Float32(Dual(...))` means anything, it surely means some kind of implicit broadcast over the numbers within the `Dual`, right? Which promotion defeats. (The tests are only for integers.)

Crossed out above, but https://github.com/JuliaDiff/ForwardDiff.jl/pull/508 recently made `Int(::Dual)` work, *but*  not in this sense of converting the base numbers. It returns an `Int`, and throws an InexactError if it doesn't fit.

I remain a little concerned that making `T(x)::T` fail is too weird. Especially for basic bitstypes. How widely would this idea be expected to propagate? `Float32(RGB(0.1,0.2,0.3))` does not work (with ImageCore.jl); I see that `Float32(1u"m")` does actually work --- Unitful.jl does it `for f in (:float, :BigFloat, :Float64, :Float32, :Float16)`. And of course things like `Float32(Vec(1,2,3,4))` from VectorizationBase.jl do work.

> Not sure it would be possible without ForwardDiff overloading something from ChainRulesCore

Yes, I have no idea where would be the right place for such a "convert the numbers inside this" to live, if it isn't overloaded on the constructors. 

I do still think the idea of making `ProjectTo` less strict here deserves some thought. In fact I thought that, on some iteration of that, someone convinced me that `Zygote.hessian` would always have the same Dual numbers (same Tag) forward and backward passes. What's the MWE that triggers this?

> That error message seems fine to me. What else would it say?

Independent of this PR, it's possible that it should say something more helpful. This appears to be the leading way to use ForwardDiff wrong, so possibly some friendly message (explaining that your buffer needs a wide enough type) would be a good idea. 

But the basic fact that you cannot convert Duals to Floats because this would silently lose derivatives is essential to understanding how this package works. I don't think the non-conversion in this PR can in fact silently lose information; it seems super-important to be sure of that.

oxinabox:
> If `Float32(Dual(...))` means anything other than conversion to Float32`, then surely it means some kind of implicit broadcast over the numbers within the `Dual`, right? 

Duals are not numbers.
Not in the sense that `Complex` are.
They are things to let operator overloading AD happen.
Which just so happens to mostly agree with mathematical motion of a dual number with the `sqrt(0)!=0` thing.
Which kinda makes this whole thing a question of semantics.

But anyway, that's why it's constructors don't return the right types.
Other operator overloading ADs do this also.
Nabla does this to many constructor.



> Which promotion defeats. (The tests are only for integers.)

Yeah, I will check tomorrow to see if removing promotion is a thing.
I am pretty sure it should be, just need to write some tests.


> I do still think the idea of making ProjectTo less strict here deserves some thought.

Out of scope for this package though.

 
> In fact I thought that, on some iteration of that, someone convinced me that Zygote.hessian would always have the same Dual numbers (same Tag) forward and backward passes. What's the MWE that triggers this?

That was me.
I thought Zygote.hessian did reverse over forwards, but it does forwards over reverse.
Which i should have released in retrospect, that is the fast direction

KristofferC:
What is special about Float64?

Isn't what you saying that for any type `T`, `T(::Dual)` should "broadcast" `T` onto the real and dual part? This feels kind of similar to the `convert(::Type{Dual}, x::T) where {T}` idea in some packages which causes a lot of invalidation issues.

mcabbott:
> Not in the sense that Complex are. They are things to let operator overloading AD happen.

Yes. My example above was Vec for SIMD, which is similarly a way of threading more information through code which was written for scalars. Another I just thought of is Measurements.jl, which FWIW does not allow such conversions:
```julia
julia> λ = measurement(0, 0.1)
0.0 ± 0.1

julia> typeof(λ)
Measurement{Float64}

julia> Float32(λ)
ERROR: MethodError: no method matching Float32(::Measurement{Float64})
```

dlfivefifty:
Wouldn't it be better to add `float32(x)` and `float64(x)` functions to avoid any issues with invalidation?

dlfivefifty:
Or better yet `float(Val(32), x)` and `float(Val(64), x)`

devmotion:
I think the definition of `Float64` in this PR is reasonable for operator overloading AD (and might be the right thing to do).

However, I am worried that `Int` (and `Integer`) and `Float64` etc. would behave differently and that `Float64(x)::Float64` would not be satisfied. A common use case for explicit `Float64` constructors or conversions seems to be that the code actually requires values of type `Float64`, but this would require `Float64(d::Dual) = Float64(value(d))` (probably together with a check of the partials, similar to `Int` and `Integer`). I came across this discussion here since e.g. it would break
```julia
using ForwardDiff
using Random

struct IsoNormal{V<:AbstractVector}
    mu::V
end

function Random.rand!(rng::Random.AbstractRNG, x::AbstractVector, d::IsoNormal)
    length(x) == length(d.mu) || throw(DimensionMismatch())
    randn!(rng, x)
    x .+= d.mu
    return x
end

Base.rand(d::IsoNormal) = rand(Float64, d)
Base.rand(::Type{T}, d::IsoNormal) where {T} = Base.rand(Random.GLOBAL_RNG, T, d)
Base.rand(rng::Random.AbstractRNG, d::IsoNormal) = rand(rng, Float64, d)
function Base.rand(rng::AbstractRNG, ::Type{T}, d::IsoNormal) where {T}
    return rand!(rng, Vector{T}(undef, length(d.mu)), d)
end

rand(IsoNormal(zeros(2))) # works, returns `Vector{Float64}`
rand(Float32, IsoNormal(zeros(2))) # works, returns `Vector{Float32}`
rand(typeof(ForwardDiff.Dual(0f0)), IsoNormal(zeros(2))) # works (but probably uncommon), returns `Dual`

rand(IsoNormal([ForwardDiff.Dual(0.0) for _ in 1:2])) # fails since `Float64(d::Dual)` not defined or a `Dual` as in this PR
rand(Float32, IsoNormal([ForwardDiff.Dual(0.0) for _ in 1:2])) # fails since `Float64(d::Dual)` not defined or a `Dual` as in this PR
rand(typeof(ForwardDiff.Dual(0f0)), IsoNormal([ForwardDiff.Dual(0.0) for _ in 1:2])) # works as expected
```
which came up in https://github.com/JuliaStats/Distributions.jl/pull/1433 (it's not clear from this example but intentionally the default type `T` is chosen as `Float64` independently of `mu` there).

KristofferC:
> I think the definition of Float64 in this PR is reasonable for operator overloading AD (and might be the right thing to do).

>  A common use case for explicit Float64 constructors or conversions seems to be that the code actually requires values of type Float64, b

Yes, if someone writes explicitly `Float64` then they do want a `Float64`. You can take this argument that you want `Constructor(d::Dual) == Dual(Constructor(d.val), 0.0)` for a huge number of constructors and add overloads after overload and it will still not work well. At that point, you are "outside" the generic method system that ForwardDiff is designed for and things will likely error out just a little bit later.

Issue #181, opened 2016-12-26
dynamic NaN-safe mode switching


jrevels:
Follow-up issue to #179. From my comment there:

> In a future PR, we could add `AbstractConfig` constructors like `GradientConfig(x, Chunk{N}, NaNSafe{true})`. The generated dual numbers would then contain partials of type `Partials{N,T,true}`, which would dispatch to the `NaN`-safe versions of methods. Downstream modules could then support passing that option through their APIs (or simply setting it to a reasonable default for themselves internally).

cossio:
This would be very nice to have!

taylormcd:
Is it possible we could use [Preferences.jl](https://github.com/JuliaPackaging/Preferences.jl) to enable/disable NaN-safe mode now?  See https://github.com/JuliaLang/Pkg.jl/issues/458.  I think that would effectively solve this long-standing issue.

devmotion:
NaN-safe mode switching via the Preferences.jl system was added in https://github.com/JuliaDiff/ForwardDiff.jl/pull/539.

PR #539, opened 2021-07-28
use the preference system for nan safe mode

KristofferC:


KristofferC:
Closes https://github.com/JuliaDiff/ForwardDiff.jl/issues/181 I think

devmotion (approved):
My knowledge about Preferences is only based on its README, but according to this it looks good to me. Seems to match the instructions in https://github.com/JuliaPackaging/Preferences.jl#conditional-loading.

chriselrod:
How does one `@set_preferences!`?

devmotion:
You use `set_preferences!`: https://juliadiff.org/ForwardDiff.jl/dev/user/advanced/#Fixing-NaN/Inf-Issues

chriselrod:
```julia
julia> using ForwardDiff, Preferences

julia> set_preferences!(ForwardDiff, "nansafe_mode" => true)
```
Ah, thanks!

PR #541, opened 2021-08-03
Rules for FFTs

dlfivefifty:
This is picking up where #495 left off. Some simple examples of derivatives involving `fft` are now working.

I wasn't able to get `dct` or `r2r` working, yet, as they are defined in FFTW.jl. See https://github.com/JuliaMath/AbstractFFTs.jl/issues/56


MilesCranmer:
Hey, what is the timeline on this? Would be very helpful to have this for a project I'm working on.

dlfivefifty:
If this is going to be merged I think it's ready.

Though it might be better in a separate package

dlfivefifty:
@MilesCranmer I'm going close this PR, move the code to https://github.com/JuliaApproximation/FastTransformsForwardDiff.jl and add support for FFTW.jl as I think its too half-baked to be in ForwardDiff.jl proper.

PR #544, opened 2021-08-31
Call Random.seed! to avoid inconsistent test failure

dlfivefifty:
PR tests sometimes fail, I believe due to floating point errors. See e.g.

https://github.com/JuliaDiff/ForwardDiff.jl/pull/541

This should hopefully fix it.

### DIFF
--- a/test/DualTest.jl
+++ b/test/DualTest.jl
@@ -14,6 +14,7 @@ import Calculus
 struct TestTag end
 
 samerng() = MersenneTwister(1)
+Random.seed!(132)
 
 # By lower-bounding the Int range at 2, we avoid cases where differentiating an
 # exponentiation of an Int value would cause a DomainError due to reducing the


KristofferC:
While at it, perhaps moving to StableRNGs.jl would make sense?

dlfivefifty:
But then we’d have to catch every ` rand ` call and pass the ` rng `? Seems more trouble than it’s worth since we test on different Julia versions and can change the seed if a future version breaks the tests

mcabbott:
Is it just one or two tests? If it is, it might be worth figuring out why, instead of avoiding such cases.

In https://github.com/JuliaDiff/ForwardDiff.jl/runs/3340080509?check_suite_focus=true the failure is this, which looks a bit familiar:
```
450
Error During Test at /home/runner/work/ForwardDiff.jl/ForwardDiff.jl/test/DualTest.jl:400
451
  Test threw exception
452
453
  DomainError with -2.98738277122415e18:
454
  sqrt will only return a complex result if called with a complex argument. Try sqrt(Complex(x)).
455
  Stacktrace:
456
    [1] throw_complex_domainerror(f::Symbol, x::Float64)
457
      @ Base.Math ./math.jl:33
458
    [2] sqrt
459
      @ ./math.jl:582 [inlined]
460
    [3] sqrt
461
      @ ~/work/ForwardDiff.jl/ForwardDiff.jl/src/dual.jl:203 [inlined]
462
    [4] generic_norm2(x::ForwardDiff.Partials{3, ForwardDiff.Dual{Main.DualTest.TestTag(), Int64, 4}})
463
      @ LinearAlgebra /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/generic.jl:509
464
    [5] norm2
465
      @ /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/generic.jl:558 [inlined]
466
    [6] norm(itr::ForwardDiff.Partials{3, ForwardDiff.Dual{Main.DualTest.TestTag(), Int64, 4}}, p::Int64)
467
      @ LinearAlgebra /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/generic.jl:627
468
    [7] norm
469
      @ /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/generic.jl:625 [inlined]
470
    [8] isapprox(x::ForwardDiff.Partials{3, ForwardDiff.Dual{Main.DualTest.TestTag(), Int64, 4}}, y::ForwardDiff.Partials{3, ForwardDiff.Dual{Main.DualTest.TestTag(), Float64, 4}}; atol::Int64, rtol::Float64, nans::Bool, norm::typeof(LinearAlgebra.norm))
471
      @ LinearAlgebra /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/generic.jl:1659
472
    [9] isapprox
473
      @ /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/generic.jl:1657 [inlined]
474
   [10] dual_isapprox(a::ForwardDiff.Dual{Main.DualTest.TestTag(), ForwardDiff.Dual{Main.DualTest.TestTag(), Int64, 4}, 3}, b::ForwardDiff.Dual{Main.DualTest.TestTag(), ForwardDiff.Dual{Main.DualTest.TestTag(), Float64, 4}, 3})
475
      @ Main.DualTest ~/work/ForwardDiff.jl/ForwardDiff.jl/test/DualTest.jl:24
```

dlfivefifty:
This bug has been around for months. If it was easy to figure out, someone would have. And it’s most likely not even a bug, just an issue if numbers are large producing slightly higher error than expected in the tests.

In any case, it’s standard practice to seed RNG in tests to avoid these issues

mcabbott:
If `isapprox` were returning `false` I'd be happier to write it off to rounding error, as you say. But this is a bit stranger, and I worry a bit it's telling us something is wrong with `^`.

dlfivefifty:
Probably we need to overload `norm(::Partials, p)` as its falling back to the generic version. It's unclear why 

https://github.com/JuliaLang/julia/blob/49e3aecd5966a2af0b064c0314cd61c1338abc00/stdlib/LinearAlgebra/src/generic.jl#L496

might cause sqrt of a negative number

odow:
Bump. This is pretty painful: https://github.com/JuliaDiff/ForwardDiff.jl/pull/553#issuecomment-961610627

odow:
Fixed in #554

Issue #548, opened 2021-10-03
Segfault from collecting a range of duals


ChrisRackauckas:
MWE:

```julia
using ForwardDiff
function f(p)
    sum(collect(0.0:p[1]:p[2]))
end
ForwardDiff.gradient(f, [0.2,25.0])
```

KristofferC:
Always good to check with segfaults:

```
julia  -q --check-bounds=yes

julia> using ForwardDiff

julia> function f(p)
           sum(collect(0.0:p[1]:p[2]))
       end
f (generic function with 1 method)

julia> ForwardDiff.gradient(f, [0.2,25.0])
ERROR: BoundsError: attempt to access 124-element Vector{ForwardDiff.Dual{ForwardDiff.Tag{typeof(f), Float64}, Float64, 2}} at index [125]
Stacktrace:
 [1] setindex!
   @ ./array.jl:839 [inlined]
 [2] vcat(rs::StepRange{ForwardDiff.Dual{ForwardDiff.Tag{typeof(f), Float64}, Float64, 2}, ForwardDiff.Dual{ForwardDiff.Tag{typeof(f), Float64}, Float64, 2}})
   @ Base ./range.jl:1036
 [3] collect
   @ ./range.jl:1043 [inlined]
 [4] f(p::Vector{ForwardDiff.Dual{ForwardDiff.Tag{typeof(f), Float64}, Float64, 2}})
   @ Main ./REPL[1]:2
 [5] vector_mode_dual_eval
   @ ~/.julia/packages/ForwardDiff/5gUap/src/apiutils.jl:37 [inlined]
 [6] vector_mode_gradient(f::typeof(f), x::Vector{Float64}, cfg::ForwardDiff.GradientConfig{ForwardDiff.Tag{typeof(f), Float64}, Float64, 2, Vector{ForwardDiff.Dual{ForwardDiff.Tag{typeof(f), Float64}, Float64, 2}}})
   @ ForwardDiff ~/.julia/packages/ForwardDiff/5gUap/src/gradient.jl:106
 [7] gradient(f::Function, x::Vector{Float64}, cfg::ForwardDiff.GradientConfig{ForwardDiff.Tag{typeof(f), Float64}, Float64, 2, Vector{ForwardDiff.Dual{ForwardDiff.Tag{typeof(f), Float64}, Float64, 2}}}, ::Val{true})
[...]

PR #549, opened 2021-10-03
define ArithmeticStyle for Dual

KristofferC:
Fixes https://github.com/JuliaDiff/ForwardDiff.jl/issues/548


ChrisRackauckas (approved):

ChrisRackauckas:
Nightly failure unrelated. Thanks for the quick fix. This looks reasonable enough.

PR #552, opened 2021-11-02
Use specified threshold instead of DEFAULT_CHUNK_THRESHOLD in pickchunksize

chriselrod:


KristofferC (approved):
Wow..

chriselrod:
> Wow..

Git blame says it's been like this for 4 years!

chriselrod:
Is this test always so unreliable?
```julia
Error During Test at /home/runner/work/ForwardDiff.jl/ForwardDiff.jl/test/DualTest.jl:400
  Test threw exception
  DomainError with -8.830391984918102e18:
```

KristofferC:
Yes, they are.

PR #554, opened 2021-11-09
Fix flakey tests in DualTest.jl

odow:
This was pretty painful for new PRs: https://github.com/JuliaDiff/ForwardDiff.jl/pull/553#issuecomment-961610627

Closes #544


KristofferC:
Thanks for tracking this down.

PR #555, opened 2021-11-19
Manual vectorization of the accumulation of `Partials`

YingboMa:
The original goal for manual vectorization was to decrease the compile time, as I thought SLP vectorizer could take a long time to run. Unfortunately, it doesn't seem to improve the compile time at all.

BUT!!! Just when I am about to stash my changes, I found out that this improves the runtime by about 20% for the Jacobian computation of a chemical differential equation system.

Master:
image

This branch:
image

The benchmark script:
```julia
using DiffEqBase, ForwardDiff, BenchmarkTools
const k1=.35e0
const k2=.266e2
const k3=.123e5
const k4=.86e-3
const k5=.82e-3
const k6=.15e5
const k7=.13e-3
const k8=.24e5
const k9=.165e5
const k10=.9e4
const k11=.22e-1
const k12=.12e5
const k13=.188e1
const k14=.163e5
const k15=.48e7
const k16=.35e-3
const k17=.175e-1
const k18=.1e9
const k19=.444e12
const k20=.124e4
const k21=.21e1
const k22=.578e1
const k23=.474e-1
const k24=.178e4
const k25=.312e1

function rhs!(dy,y,p,t)
    @inbounds begin
        r1  = k1 *y[1]
        r2  = k2 *y[2]*y[4]
        r3  = k3 *y[5]*y[2]
        r4  = k4 *y[7]
        r5  = k5 *y[7]
        r6  = k6 *y[7]*y[6]
        r7  = k7 *y[9]
        r8  = k8 *y[9]*y[6]
        r9  = k9 *y[11]*y[2]
        r10 = k10*y[11]*y[1]
        r11 = k11*y[13]
        r12 = k12*y[10]*y[2]
        r13 = k13*y[14]
        r14 = k14*y[1]*y[6]
        r15 = k15*y[3]
        r16 = k16*y[4]
        r17 = k17*y[4]
        r18 = k18*y[16]
        r19 = k19*y[16]
        r20 = k20*y[17]*y[6]
        r21 = k21*y[19]
        r22 = k22*y[19]
        r23 = k23*y[1]*y[4]
        r24 = k24*y[19]*y[1]
        r25 = k25*y[20]

        dy[1]  = -r1-r10-r14-r23-r24+
        r2+r3+r9+r11+r12+r22+r25
        dy[2]  = -r2-r3-r9-r12+r1+r21
        dy[3]  = -r15+r1+r17+r19+r22
        dy[4]  = -r2-r16-r17-r23+r15
        dy[5]  = -r3+r4+r4+r6+r7+r13+r20
        dy[6]  = -r6-r8-r14-r20+r3+r18+r18
        dy[7]  = -r4-r5-r6+r13
        dy[8]  = r4+r5+r6+r7
        dy[9]  = -r7-r8
        dy[10] = -r12+r7+r9
        dy[11] = -r9-r10+r8+r11
        dy[12] = r9
        dy[13] = -r11+r10
        dy[14] = -r13+r12
        dy[15] = r14
        dy[16] = -r18-r19+r16
        dy[17] = -r20
        dy[18] = r20
        dy[19] = -r21-r22-r24+r23+r25
        dy[20] = -r25+r24
    end
    nothing
end
f! = DiffEqBase.UJacobianWrapper(rhs!, 0.0, nothing)

u0 = zeros(20)
u0[2]  = 0.2
u0[4]  = 0.04
u0[7]  = 0.1
u0[8]  = 0.3
u0[9]  = 0.01
u0[17] = 0.007

J = similar(u0, length(u0), length(u0))
du = similar(u0)
cfg = ForwardDiff.JacobianConfig(f!, du, u0, ForwardDiff.Chunk(8))
#tinf = @snoopi_deep ForwardDiff.jacobian!(J, f!, du, u0, cfg)
@time ForwardDiff.jacobian!(J, f!, du, u0, cfg)
display(@benchmark ForwardDiff.jacobian!($J, $f!, $du, $u0, $cfg))
println()
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -14,6 +14,7 @@ Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 SpecialFunctions = "276daf66-3868-5448-9aa4-cd146d93841b"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
+VectorizationBase = "3d5dd08c-fd9d-11e8-17fa-ed2836048c2f"
 
 [compat]
 Calculus = "0.2, 0.3, 0.4, 0.5"
--- a/Project.toml
+++ b/Project.toml
@@ -14,7 +14,6 @@ Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 SpecialFunctions = "276daf66-3868-5448-9aa4-cd146d93841b"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
-VectorizationBase = "3d5dd08c-fd9d-11e8-17fa-ed2836048c2f"
 
 [compat]
 Calculus = "0.2, 0.3, 0.4, 0.5"
--- a/src/ForwardDiff.jl
+++ b/src/ForwardDiff.jl
@@ -8,6 +8,7 @@ if VERSION >= v"1.6"
 end
 using Random
 using LinearAlgebra
+using VectorizationBase
 
 import Printf
 import NaNMath
--- a/src/ForwardDiff.jl
+++ b/src/ForwardDiff.jl
@@ -8,7 +8,6 @@ if VERSION >= v"1.6"
 end
 using Random
 using LinearAlgebra
-using VectorizationBase
 
 import Printf
 import NaNMath
--- a/src/partials.jl
+++ b/src/partials.jl
@@ -300,27 +300,33 @@ end
     end
 end
 
-@generated function minus_tuple(tup::NTuple{N,T}) where {N,T}
-    T <: SIMDType || return tupexpr(i -> :(-tup[$i]), N)
+if VERSION >= v"1.4" # fsub requires LLVM 8 (Julia 1.4)
+    @generated function minus_tuple(tup::NTuple{N,T}) where {N,T}
+        T <: SIMDType || return tupexpr(i -> :(-tup[$i]), N)
 
-    S = julia_type_to_llvm_type(T)
-    VT = NTuple{N, VecElement{T}}
-    if T <: SIMDFloat
-        llvmir = """
-        %res = fneg nsz contract <$N x $S> %0
-        ret <$N x $S> %res
-        """
-    else
-        llvmir = """
-        %res = sub <$N x $S> zeroinitializer, %0
-        ret <$N x $S> %res
-        """
+        S = julia_type_to_llvm_type(T)
+        VT = NTuple{N, VecElement{T}}
+        if T <: SIMDFloat
+            llvmir = """
+            %res = fneg nsz contract <$N x $S> %0
+            ret <$N x $S> %res
+            """
+        else
+            llvmir = """
+            %res = sub <$N x $S> zeroinitializer, %0
+            ret <$N x $S> %res
+            """
+        end
+
+        quote
+            $(Expr(:meta, :inline))
+            ret = Base.llvmcall($llvmir, $VT, Tuple{$VT}, $VT(tup))
+            Base.@ntuple $N i->ret[i].value
+        end
     end
-
-    quote
-        $(Expr(:meta, :inline))
-        ret = Base.llvmcall($llvmir, $VT, Tuple{$VT}, $VT(tup))
-        Base.@ntuple $N i->ret[i].value
+else
+    @generated function minus_tuple(tup::NTuple{N,T}) where {N,T}
+        return tupexpr(i -> :(-tup[$i]), N)
     end
 end
 
--- a/src/partials.jl
+++ b/src/partials.jl
@@ -197,29 +197,174 @@ end
     return tupexpr(i -> :(rand(V)), N)
 end
 
-@generated function scale_tuple(tup::NTuple{N}, x) where N
-    return tupexpr(i -> :(tup[$i] * x), N)
+const SIMDFloat = Union{Float64, Float32}
+const SIMDInt = Union{
+                       Int128, Int64, Int32, Int16, Int8,
+                       UInt128, UInt64, UInt32, UInt16, UInt8,
+                       Bool
+                     }
+const SIMDType = Union{SIMDFloat, SIMDInt}
+
+function julia_type_to_llvm_type(@nospecialize(T::DataType))
+    T === Float64 ? "double" :
+    T === Float32 ? "float"  :
+    T <: Union{Int128,UInt128} ? "i128" :
+    T <: Union{Int64,UInt64} ? "i64" :
+    T <: Union{Int32,UInt32} ? "i32" :
+    T <: Union{Int16,UInt16} ? "i16" :
+    T <: Union{Bool,Int8,UInt8} ? "i8" :
+    error("$T cannot be mapped to a LLVM type")
 end
 
-@generated function div_tuple_by_scalar(tup::NTuple{N}, x) where N
-    return tupexpr(i -> :(tup[$i] / x), N)
+function llvmir_scalar_to_vec(@nospecialize(T::DataType), n::Int, vname::String)
+    S = julia_type_to_llvm_type(T)
+    el = string("ele", vname)
+    """
+      %$el = insertelement <$n x $S> undef, $S %0, i32 0
+      %$vname = shufflevector <$n x $S> %$el, <$n x $S> undef, <$n x i32> zeroinitializer
+    """
 end
 
-@generated function add_tuples(a::NTuple{N}, b::NTuple{N})  where N
-    return tupexpr(i -> :(a[$i] + b[$i]), N)
+@generated function scale_tuple(tup::NTuple{N,T1}, x::S1) where {N,T1,S1}
+    (T1 <: SIMDType && S1 <: SIMDType) || return tupexpr(i -> :(tup[$i] * x), N)
+
+    T = promote_type(T1, S1)
+    S = julia_type_to_llvm_type(T)
+    VT = NTuple{N, VecElement{T}}
+    op = T <: SIMDFloat ? "fmul nsz contract" : "mul"
+    llvmir = """
+    %el = insertelement <$N x $S> undef, $S %1, i32 0
+    %vx = shufflevector <$N x $S> %el, <$N x $S> undef, <$N x i32> zeroinitializer
+    %res = $op <$N x $S> %0, %vx
+    ret <$N x $S> %res
+    """
+
+    quote
+        $(Expr(:meta, :inline))
+        t = Base.@ntuple $N i->$T(tup[i])
+        ret = Base.llvmcall($llvmir, $VT, Tuple{$VT, $T}, $VT(t), $T(x))
+        Base.@ntuple $N i->ret[i].value
+    end
 end
 
-@generated function sub_tuples(a::NTuple{N}, b::NTuple{N})  where N
-    return tupexpr(i -> :(a[$i] - b[$i]), N)
+@generated function div_tuple_by_scalar(tup::NTuple{N,T1}, x::S1) where {N,T1,S1}
+    (T1 <: SIMDType && S1 <: SIMDType) || return tupexpr(i -> :(tup[$i] / x), N)
+
+    T = typeof(one(T1) / one(S1))
+    S = julia_type_to_llvm_type(T)
+    VT = NTuple{N, VecElement{T}}
+    op = T <: SIMDFloat ? "fdiv nsz contract" : "div"
+    llvmir = """
+    %el = insertelement <$N x $S> undef, $S %1, i32 0
+    %vx = shufflevector <$N x $S> %el, <$N x $S> undef, <$N x i32> zeroinitializer
+    %res = $op <$N x $S> %0, %vx
+    ret <$N x $S> %res
+    """
+
+    quote
+        $(Expr(:meta, :inline))
+        t = Base.@ntuple $N i->$T(tup[i])
+        ret = Base.llvmcall($llvmir, $VT, Tuple{$VT, $T}, $VT(t), $T(x))
+        Base.@ntuple $N i->ret[i].value
+    end
 end
 
-@generated function minus_tuple(tup::NTuple{N}) where N
-    return tupexpr(i -> :(-tup[$i]), N)
+@generated function add_tuples(a::NTuple{N,T1}, b::NTuple{N,S1}) where {N,T1,S1}
+    (T1 <: SIMDType && S1 <: SIMDType) || return tupexpr(i -> :(a[$i] + b[$i]), N)
+
+    T = promote_type(T1, S1)
+    S = julia_type_to_llvm_type(T)
+    VT = NTuple{N, VecElement{T}}
+    op = T <: SIMDFloat ? "fadd nsz contract" : "add"
+    llvmir = """
+    %res = $op <$N x $S> %0, %1
+    ret <$N x $S> %res
+    """
+
+    quote
+        $(Expr(:meta, :inline))
+        at = Base.@ntuple $N i->$T(a[i])
+        bt = Base.@ntuple $N i->$T(b[i])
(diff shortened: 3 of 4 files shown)


YingboMa:
With chunk size = 4, there is about 10% improvement.

Master:
image

This branch:
image

KristofferC:
Is the speed improvement here just from the injected fastmath flags? For example, ForwardDiff already auto-vectorizes seemingly ok:

```
julia> @code_native d * 3.0
        .section        __TEXT,__text,regular,pure_instructions
; ┌ @ dual.jl:144 within `*'
        movq    %rdi, %rax
; │ @ dual.jl:243 within `*' @ float.jl:332
        vbroadcastsd    %xmm0, %ymm1
        vmulpd  (%rsi), %ymm1, %ymm1
...

julia> @code_native d + d
        .section        __TEXT,__text,regular,pure_instructions
; ┌ @ dual.jl:139 within `+'
        movq    %rdi, %rax
; │ @ dual.jl:422 within `+' @ float.jl:326
        vmovupd (%rsi), %ymm0
        vaddpd  (%rdx), %ymm0, %ymm0
...
```

I think before this is merged we need to figure out what optimization LLVM misses with the normal tuples and why going to llvmcall is strictly needed. Also, instead of hand-rolling our own LLVM SIMD everywhere, if this is really required, wouldn't SIMD.jl be easier to use? 

Also, please use `@btime` unless the histogram and all other numbers are of any use (which it isn't here since we just want the value most close to the truth (which the minimum already shows best)

YingboMa:
> For example, ForwardDiff already auto-vectorizes seemingly ok:

`@code_native` on a small expression is highly misleading. How well SLP vectorizer performs depends on the context, if that is inline in some complex code with indexing and such, it won't produce the same code quality.

> Is the speed improvement here just from the injected fastmath flags?

It is not:

```julia
➜  ForwardDiff git:(master) julia --math-mode=fast --startup-file=no ~/pollu.jl
  1.289900 seconds (4.68 M allocations: 241.667 MiB, 18.83% gc time, 99.99% compilation time)
BenchmarkTools.Trial: 10000 samples with 73 evaluations.
 Range (min … max):  870.356 ns …   3.028 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     890.062 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   911.691 ns ± 113.366 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

   █▇▃▂▁▂▂  ▁                                                   ▁
  ████████▇██▇▇▇▆▆▆▅▅▆▃▄▅▆▅▅▅▄▄▄▃▃▅▃▄▄▅▅▅▅▄▄▃▄▅▄▃▄▂▃▂▂▃▅▂▃▃▂▄▃▅ █
  870 ns        Histogram: log(frequency) by time       1.47 μs <

 Memory estimate: 0 bytes, allocs estimate: 0.
➜  ForwardDiff git:(master) julia --math-mode=fast --startup-file=no ~/pollu.jl
  1.137902 seconds (4.68 M allocations: 241.667 MiB, 13.23% gc time, 99.99% compilation time)
BenchmarkTools.Trial: 10000 samples with 71 evaluations.
 Range (min … max):  882.451 ns …   3.489 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     895.127 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   918.756 ns ± 126.937 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▇█▄▃▁▁▂  ▁                                                    ▁
  ███████▇██▇▆▆▅▇▅▄▆▄▃▅▅▅▅▂▄▃▄▅▃▄▅▅▅▅▄▅▄▄▃▃▄▄▄▄▄▃▃▄▄▄▄▃▃▃▄▃▃▃▄▃ █
  882 ns        Histogram: log(frequency) by time       1.53 μs <

 Memory estimate: 0 bytes, allocs estimate: 0.
```

> Also, please use @btime unless the histogram and all other numbers are of any use (which it isn't here since we just want the value most close to the truth (which the minimum already shows best))

Not necessarily, if you look at the histogram, most of the samples are no where near the min time. Why would you say min is the most close to the truth? If anything it's misleading to believe that this function would really run that fast in a real world setting.

KristofferC:
> `@code_native` on a small expression is highly misleading.  How well SLP vectorizer performs depends on the context, if that is inlined in some complex code with indexing and such, it won't produce the same code quality.

The end result is still the optimization of the sum of the parts. 

> Not necessarily, if you look at the histogram, most of the samples are no where near the min time. Why would you say min is the most close to the truth? 

Yes, because your computer has noise which is always additive. The fastest run is the one with the least noise. We are not writing an HTTP server here where there is scheduling and we care about percentiles etc. The mode of the samples is close to the min (and the more samples you take the more the mode will move towards the min).

> Also, I like those colors, thus, the screenshots

I agree that colors are good if they show something useful.

YingboMa:
> The end result is still the optimization of the sum of the parts. For example here, the contract fast math option used here could perhaps make a difference when doing optimization on the whole.

That's not true. If that were the case, `--math-mode=fast` should speed the master branch to this branch's speed as well. If you actually see the native code for the rhs function, the manual vectorization is better. Also, the speed up is proven empirically, so I don't see the dispute. SLP objectively produces worse code if there's complex code surrounding the unrolled part as shown in the benchmark.

Maybe you can also run the benchmark to see if there's a speed up.

> Yes, because your computer has noise which is always additive. The fastest run is the one with the least noise. We are not writing an HTTP server here where there is scheduling and we care about percentiles etc. The mode of the samples is close to the min (and the more samples you take the more the mode will move towards the min).

ForwardDiff.jl doesn't run on perfect computers. I don't see how you can prove changes in native code cannot produce pathological cases that I mentioned in here.

> Further, I am not sure that changes in the native code don't shift the min/mean/median in unintuitive ways. For instance, maybe the min time is better, yet, the mean is much worse, leading to an overall pessimization in real world use cases.

KristofferC:
> That's not true. If that were the case, --math-mode=fast should speed the master branch to this branch's speed as well. If you actually see the native code for the rhs function, the manual vectorization is better.

I am not saying that the contract is the actual cause, I am saying it could be. Or one of the other fast math flags introduced. 

> ForwardDiff.jl doesn't run on perfect computers. I don't see how you can prove changes in native code cannot produce pathological cases that I mentioned in here.

See https://youtu.be/vrfYLlR8X8k?t=915 and the following ~10 min. No one cares about the particular noise of your computer, we want to benchmark the algorithm as well as we can, that's the most unbiased result.

---------------

I am getting some really strange things. If I run with the local ForwardDiff:

```
~/JuliaPkgs/ForwardDiff.jl tags/v0.10.23*> julia -q --project

julia> using ForwardDiff

julia> d = ForwardDiff.Dual(1.0, 2.0, 3.0, 4.0, 5.0);

julia> @code_native d * 3.0
        .text
; ┌ @ dual.jl:144 within `*'
        movq    %rdi, %rax
; │ @ dual.jl:243 within `*' @ float.jl:332
        vmovsd  (%rsi), %xmm2                   # xmm2 = mem[0],zero
; │ @ dual.jl:244 within `*' @ partials.jl:84 @ partials.jl:0
        vmovsd  8(%rsi), %xmm1                  # xmm1 = mem[0],zero
; │ @ dual.jl:244 within `*' @ partials.jl:84 @ partials.jl:95
; │┌ @ float.jl:452 within `isfinite'
; ││┌ @ float.jl:329 within `-'
        vsubsd  %xmm0, %xmm0, %xmm3
        vxorpd  %xmm4, %xmm4, %xmm4
; ││└
; ││┌ @ float.jl:401 within `==' @ float.jl:365
        vucomisd        %xmm4, %xmm3
; │└└
; │ @ dual.jl:244 within `*' @ partials.jl:84 @ partials.jl:0
        vmovsd  16(%rsi), %xmm3                 # xmm3 = mem[0],zero
; │ @ dual.jl:244 within `*' @ partials.jl:84 @ partials.jl:95
        jne     L33
        jnp     L72
L33:
        vucomisd        %xmm4, %xmm1
        jne     L72
        jp      L72
; │ @ dual.jl:244 within `*' @ partials.jl:84 @ partials.jl:0
        vxorpd  %xmm5, %xmm5, %xmm5
; │ @ dual.jl:244 within `*' @ partials.jl:84 @ partials.jl:95
        vucomisd        %xmm5, %xmm3
        jne     L72
        jp      L72
; │┌ @ partials.jl:37 within `iszero'
; ││┌ @ partials.jl:167 within `iszero_tuple'
; │││┌ @ partials.jl:172 within `macro expansion'
; ││││┌ @ float.jl:365 within `=='
        vmovsd  24(%rsi), %xmm4                 # xmm4 = mem[0],zero
....
```

If I run with the versioned:

```
(@v1.6) pkg> activate --temp
  Activating new environment at `/tmp/jl_p9UUFZ/Project.toml`

(jl_p9UUFZ) pkg> add ForwardDiff@0.10.23

julia> using ForwardDiff

julia> d = ForwardDiff.Dual(1.0, 2.0, 3.0, 4.0, 5.0);

julia> @code_native d * 3.0
        .text
; ┌ @ dual.jl:144 within `*'
        movq    %rdi, %rax
; │ @ dual.jl:243 within `*' @ float.jl:332
        vbroadcastsd    %xmm0, %ymm1
        vmulpd  (%rsi), %ymm1, %ymm1
; │ @ dual.jl:244 within `*' @ partials.jl:84 @ partials.jl:111
; │┌ @ partials.jl:200 within `scale_tuple'
; ││┌ @ partials.jl:157 within `macro expansion'
; │││┌ @ float.jl:332 within `*'
        vmulsd  32(%rsi), %xmm0, %xmm0
; │└└└
; │ @ dual.jl:244 within `*'
        vmovupd %ymm1, (%rdi)
        vmovsd  %xmm0, 32(%rdi)
        vzeroupper
        retq
        nop
; └
```

Why is the local one so trash?

YingboMa:
I get:
```julia
julia> @code_native debuginfo=:none d * 3.0
        .section        __TEXT,__text,regular,pure_instructions
        movq    %rdi, %rax
        vmulsd  (%rsi), %xmm0, %xmm1
        vbroadcastsd    %xmm0, %ymm0
        vmulpd  8(%rsi), %ymm0, %ymm0
        vmovsd  %xmm1, (%rdi)
        vmovupd %ymm0, 8(%rdi)
        vzeroupper
        retq
        nop
```

> I am not saying that the contract is the actual cause, I am saying it could be. Or one of the other fast math flags introduced. Or something else. But it is not magic.

What do you mean by magic? SLP is not magic. So maybe you can also accept that LLVM cannot always somehow deduces the way to group unrolled expression into optimal SIMD code. Here's the benchmark without flags with `@btime` (just for you :-P)
```julia
➜  ForwardDiff git:(myb/vec) ✗ julia --startup-file=no ~/pollu.jl
  0.979201 seconds (4.80 M allocations: 243.740 MiB, 13.99% gc time, 99.99% compilation time)
  833.165 ns (0 allocations: 0 bytes)
➜  ForwardDiff git:(myb/vec) ✗ julia --startup-file=no ~/pollu.jl
  1.007542 seconds (4.80 M allocations: 243.740 MiB, 14.13% gc time, 99.99% compilation time)
  842.387 ns (0 allocations: 0 bytes)
➜  ForwardDiff git:(myb/vec) ✗ julia --startup-file=no ~/pollu.jl
  0.991858 seconds (4.80 M allocations: 243.740 MiB, 13.58% gc time, 99.99% compilation time)
  773.643 ns (0 allocations: 0 bytes)
```

```julia
➜  ForwardDiff git:(master) julia --startup-file=no ~/pollu.jl
  1.106445 seconds (4.68 M allocations: 241.588 MiB, 20.34% gc time, 99.99% compilation time)
  902.947 ns (0 allocations: 0 bytes)
➜  ForwardDiff git:(master) julia --startup-file=no ~/pollu.jl
  0.952378 seconds (4.68 M allocations: 241.588 MiB, 13.72% gc time, 99.99% compilation time)
  903.885 ns (0 allocations: 0 bytes)
➜  ForwardDiff git:(master) julia --startup-file=no ~/pollu.jl
  0.968903 seconds (4.68 M allocations: 241.588 MiB, 13.50% gc time, 99.99% compilation time)
  921.333 ns (0 allocations: 0 bytes)
```
Indeed, I have a noisy computer. But the speed up is still there.

KristofferC:
>  I get:

Ok, I had 

```
❯ cat LocalPreferences.toml
[ForwardDiff]
nansafe_mode = true
```

locally, which apparently trashes the whole thing.

KristofferC:
This is what I get locally (with 1.6):

PR:

```
~/JuliaPkgs/ForwardDiff.jl myb/vec* 23s
❯ julia --project ~/JuliaTests/fdiffbench.jl
  0.874255 seconds (5.05 M allocations: 275.217 MiB, 7.34% gc time, 99.99% compilation time)
  1.278 μs (0 allocations: 0 bytes)
```

master:

```
~/JuliaPkgs/ForwardDiffOrig.jl master* 8s
❯ julia --project ~/JuliaTests/fdiffbench.jl
  0.820081 seconds (4.31 M allocations: 241.401 MiB, 6.69% gc time, 99.99% compilation time)
  1.316 μs (0 allocations: 0 bytes)
```

KristofferC:
With this branch, locally I get:

```
❯ julia --project ~/JuliaTests/fdiffbench.jl
  0.770050 seconds (4.75 M allocations: 260.031 MiB, 7.25% gc time, 99.99% compilation time)
  995.077 ns (0 allocations: 0 bytes)
```

If I remove the changes to `div_tuple_by_scalar` I get:

```
❯ julia --project ~/JuliaTests/fdiffbench.jl
  0.743958 seconds (4.75 M allocations: 260.048 MiB, 7.28% gc time, 99.99% compilation time)
  875.692 ns (0 allocations: 0 bytes)
```

Can you repro that?

YingboMa:
I cannot observe the difference:
```
➜  ForwardDiff git:(myb/vec) julia --startup-file=no ~/pollu.jl
  1.128815 seconds (4.80 M allocations: 243.678 MiB, 13.70% gc time, 99.99% compilation time)
  827.237 ns (0 allocations: 0 bytes)
➜  ForwardDiff git:(myb/vec) ✗ julia --startup-file=no ~/pollu.jl
  0.971470 seconds (4.80 M allocations: 243.692 MiB, 13.62% gc time, 99.99% compilation time)
  831.758 ns (0 allocations: 0 bytes)
```

In fact, I changed the div function to
```julia
    return tupexpr(i -> :((println("Hello")); return tup[$i] / x), N)
```
and nothing was printed, so I am pretty sure it won't change anything :-p

KristofferC:
> and nothing was printed, so I am pretty sure it won't change anything :-p

Maybe it is some code layout effect then (https://easyperf.net/blog/2018/01/18/Code_alignment_issues) because it is quite consistent for me. I keep doing it over and over :P Or it was noise. :man_shrugging:

KristofferC:
I seem to get the same performance (and the same number of instructions using)

```jl
using SIMD

function scale_tuple(tup::NTuple{N}, x) where N
    return @fastmath Tuple(Vec(tup) * x)
end

function div_tuple_by_scalar(tup::NTuple{N}, x) where N
    return @fastmath Tuple(Vec(tup) / x)
end

function add_tuples(a::NTuple{N}, b::NTuple{N})  where N
    return @fastmath Tuple(Vec(a) + Vec(b))
end

function sub_tuples(a::NTuple{N}, b::NTuple{N})  where N
    return @fastmath Tuple(Vec(a) - Vec(b))
end

function minus_tuple(tup::NTuple{N}) where N
    return @fastmath Tuple(-Vec(tup))
end

function mul_tuples(a::NTuple{N,V1}, b::NTuple{N,V2}, afactor::S1, bfactor::S2) where {N,V1,V2,S1,S2}
    af = Vec{N,V1}(afactor)
    bf = Vec{N,V2}(bfactor)
    @fastmath Tuple(muladd(Vec(a), af, bf * Vec(b)))
end
```

Removing `@fastmath` adds like 10 instructions. Being able to use the muladd in `mul_tuples` is nice, indeed.

KristofferC:
I put up a PR so it is easier to compare: https://github.com/JuliaDiff/ForwardDiff.jl/pull/557

KristofferC:
Incorporated into https://github.com/JuliaDiff/ForwardDiff.jl/pull/557.

PR #557, opened 2021-11-19
use SIMD.jl for explicit vectorization of partial operations

KristofferC:
Alternative to https://github.com/JuliaDiff/ForwardDiff.jl/pull/555

Co-authored-by: Yingbo Ma <mayingbo5@gmail.com>

Using the benchmark in #555

I extracted the code from the benchmark in https://github.com/JuliaDiff/ForwardDiff.jl/pull/555  (https://gist.github.com/KristofferC/a3915eec407491b7a4fc8a7d81c16fbe) and put it into https://uica.uops.info/. The result is (with some various tools) in cycles per iteration:

Explicit:
```
uiCA | 265.50
IACA 3.0 | 245.00
llvm-mca | 210.06
```

Master:
```
uiCA | 371.00
IACA 3.0 | 318.00
llvm-mca | 263.11
```

So it does seem that the strategy in #555 is quite good.


KristofferC:
Since this seems to have the same performance improvements as #555 (and also some better compile-time behavior) as well as not having to duplicate all the llvmcall logic that exists in SIMD.jl I think it is better to go with this one.

PR #561, opened 2021-11-20
define sin/cos on Dual to use sincos

KristofferC:
A bit more efficient since `sincos` can share some computation between the `sin` and `cos`:

PR:

```jl
julia> d = Dual(1.0,2.0,3.0,4.0,5.0)

julia> @btime sin($d)
  10.050 ns (0 allocations: 0 bytes)
Dual{Nothing}(0.8414709848078965,1.0806046117362795,1.6209069176044193,2.161209223472559,2.701511529340699)
```

Master:

```jl
julia> d = Dual(1.0,2.0,3.0,4.0,5.0)

julia> @btime sin($d)
  13.638 ns (0 allocations: 0 bytes)
Dual{Nothing}(0.8414709848078965,1.0806046117362795,1.6209069176044193,2.161209223472559,2.701511529340699)
```

### DIFF
--- a/src/dual.jl
+++ b/src/dual.jl
@@ -393,7 +393,7 @@ Base.float(d::Dual) = convert(float(typeof(d)), d)
 ###################################
 
 for (M, f, arity) in DiffRules.diffrules(filter_modules = nothing)
-    if (M, f) in ((:Base, :^), (:NaNMath, :pow), (:Base, :/), (:Base, :+), (:Base, :-))
+    if (M, f) in ((:Base, :^), (:NaNMath, :pow), (:Base, :/), (:Base, :+), (:Base, :-), (:Base, :sin), (:Base, :cos))
         continue  # Skip methods which we define elsewhere.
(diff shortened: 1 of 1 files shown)


src/dual.jl
```diff
@@ -622,12 +622,19 @@ end
...
 
-# sincos #
+# sin/cos #
 #--------#
+function Base.sin(d::Dual{T}) where T
+    s, c = sincos(value(d))
+    return Dual{T}(s, c * partials(d))
+end
 
-@inline sincos(x) = (sin(x), cos(x))
```
KristofferC, line 628: I don't know why this was here.

chriselrod (approved):
Might be worth doing similar for `sincosd`, at least if `isdefined(Base, :sincos)`?

YingboMa (approved):

KristofferC:
> Might be worth doing similar for `sincosd`

Base defines `sincosd(x)` as `sind(x), cosd(x)` so it doesn't seem there is any performance to gain there.

PR #568, opened 2021-12-12
Fix `ldexp` test errors

devmotion:
This PR fixes the test error caused by https://github.com/JuliaDiff/DiffRules.jl/pull/73:
```julia
julia>] test ForwardDiff
...
...auto-testing Base.ldexp with 2 arguments
ERROR: LoadError: MethodError: no method matching ldexp(::Int64, ::Float64)
Closest candidates are:
  ldexp(::ForwardDiff.Dual{Tx}, ::AbstractFloat) where Tx at ~/.julia/packages/ForwardDiff/tZ5o1/src/dual.jl:144
  ldexp(::ForwardDiff.Dual{Tx}, ::Real) where Tx at ~/.julia/packages/ForwardDiff/tZ5o1/src/dual.jl:144
  ldexp(::Integer, ::ForwardDiff.Dual{Ty}) where Ty at ~/.julia/packages/ForwardDiff/tZ5o1/src/dual.jl:145
  ...
Stacktrace:
 [1] ldexp(x::ForwardDiff.Dual{Main.DualTest.TestTag(), Int64, 1}, y::Float64)
   @ ForwardDiff ~/.julia/packages/ForwardDiff/tZ5o1/src/dual.jl:243
...
```

### DIFF
--- a/test/DualTest.jl
+++ b/test/DualTest.jl
@@ -444,24 +444,23 @@ for N in (0,3), M in (0,4), V in (Int, Float32)
                 end
             elseif arity == 2
                 derivs = DiffRules.diffrule(M, f, :x, :y)
+                x, y = if f === :ldexp
+                    rand(), rand(1:10)
+                elseif f === :mod
+                    13 + rand(), 5 + rand() # make sure x/y is not an integer
+                else
+                    rand(1:10), rand()
+                end
                 @eval begin
-                    x, y = rand(1:10), rand()
+                    x, y = $x, $y
                     dx = $M.$f(Dual{TestTag()}(x, one(x)), y)
                     dy = $M.$f(x, Dual{TestTag()}(y, one(y)))
                     actualdx = $(derivs[1])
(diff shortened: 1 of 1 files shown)


test/DualTest.jl
```diff
@@ -444,24 +444,23 @@ for N in (0,3), M in (0,4), V in (Int, Float32)
                 end
             elseif arity == 2
                 derivs = DiffRules.diffrule(M, f, :x, :y)
+                x, y = if f === :ldexp
+                    rand(), rand(1:10)
+                elseif f === :mod
+                    13 + rand(), 5 + rand() # make sure x/y is not an integer
```
devmotion, line 450: Random values where `x/y` where (close to) an integer caused problems in DiffRules (https://github.com/JuliaDiff/DiffRules.jl/pull/78), so I thought it's good to apply the same fix here as well.

mcabbott (approved):

PR #577, opened 2022-02-23
Fix DiffRules-based definitions for complex-valued functions

devmotion:
Currently, complex-valued functions (with real inputs!) for which derivatives are defined in DiffRules don't work with ForwardDiff since the return values for DiffRules-based definitions is enforced to be a `Dual` but should be a `Complex{<:Dual}` in these cases. E.g., the following example fails currently (and is therefore excluded from the tests):
```julia
julia> using ForwardDiff, SpecialFunctions

julia> hankelh1(1, 2.0)
0.5767248077568733 - 0.10703243154093756im

julia> hankelh1(1, ForwardDiff.Dual(2.0))
ERROR: ArgumentError: Cannot create a dual over scalar type Any. If the type behaves as a scalar, define ForwardDiff.can_dual(::Type{Any}) = true.
Stacktrace:
 [1] throw_cannot_dual(V::Type)
   @ ForwardDiff ~/.julia/packages/ForwardDiff/PBzup/src/dual.jl:41
 [2] ForwardDiff.Dual{Nothing, Any, 1}(value::ComplexF64, partials::ForwardDiff.Partials{1, Any})
   @ ForwardDiff ~/.julia/packages/ForwardDiff/PBzup/src/dual.jl:18
 [3] Dual
   @ ~/.julia/packages/ForwardDiff/PBzup/src/dual.jl:60 [inlined]
 [4] Dual
   @ ~/.julia/packages/ForwardDiff/PBzup/src/dual.jl:64 [inlined]
 [5] Dual
   @ ~/.julia/packages/ForwardDiff/PBzup/src/dual.jl:67 [inlined]
 [6] Dual
   @ ~/.julia/packages/ForwardDiff/PBzup/src/dual.jl:71 [inlined]
 [7] hankelh1(x::Int64, y::ForwardDiff.Dual{Nothing, Float64, 0})
   @ ForwardDiff ~/.julia/packages/ForwardDiff/PBzup/src/dual.jl:249
 [8] top-level scope
   @ REPL[9]:1
```

This PR fixes this issue and returns values of type `Complex{<:Dual}` if the primal function value is a `Complex` number. This fixes the example above:
```julia
julia> using ForwardDiff, SpecialFunctions

julia> hankelh1(1, 2.0)
0.5767248077568733 - 0.10703243154093756im

julia> hankelh1(1, ForwardDiff.Dual(2.0))
Dual{Nothing}(0.5767248077568733) - Dual{Nothing}(0.10703243154093756)*im
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ForwardDiff"
 uuid = "f6369f11-7733-5829-9624-2563aa707210"
-version = "0.10.25"
+version = "0.10.27"
 
 [deps]
 CommonSubexpressions = "bbf7d656-a473-5ed7-a52c-81e309532950"
--- a/src/dual.jl
+++ b/src/dual.jl
@@ -195,6 +195,38 @@ macro define_ternary_dual_op(f, xyz_body, xy_body, xz_body, yz_body, x_body, y_b
     return esc(defs)
 end
 
+# Support complex-valued functions such as `hankelh1`
+function dual_definition_retval(::Val{T}, val::Real, deriv::Real, partial::Partials) where {T}
+    return Dual{T}(val, deriv * partial)
+end
+function dual_definition_retval(::Val{T}, val::Real, deriv1::Real, partial1::Partials, deriv2::Real, partial2::Partials) where {T}
+    return Dual{T}(val, _mul_partials(partial1, partial2, deriv1, deriv2))
+end
+function dual_definition_retval(::Val{T}, val::Complex, deriv::Union{Real,Complex}, partial::Partials) where {T}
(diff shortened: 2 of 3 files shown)


devmotion:
Bump :slightly_smiling_face:

devmotion:
Bump 🙂 It would be very helpful, in DiffRules and downstream packages, to fix this bug.

andreasnoack:
There have been a couple of discussions of support for complex numbers. Might be good to address the issues here.

andreasnoack:
Okay. Then let's get this one in.

andreasnoack (approved):

PR #580, opened 2022-03-08
Define `precision` for `Dual`

devmotion:
Similar to `eps`, this PR adds an implementation of `precision(::Dual)` and `precision(::Type{<:Dual})`, with support of the `base` keyword argument in Julia 1.8.

This is useful e.g. in LogExpFunctions (https://github.com/JuliaStats/LogExpFunctions.jl/pull/37#issuecomment-1062235555) to define precision-specific branches.


cossio:
Can this be merged?

tpapp (approved):
LGTM.

PR #582, opened 2022-04-25
DEFAULT_CHUNK_THRESHOLD is configurable via Preferences.jl

lassepe:
Context: I would like to set the `DEFAULT_CHUNK_THRESHOLD` for a project that uses ForwardDiff.jl through Zygote.jl. Since configuring the chunk size used in `Zygote.forwarddiff` is not supported by the Zygote API, I figured it may be okay to just expose this as a preference here.

### DIFF
--- a/docs/src/user/advanced.md
+++ b/docs/src/user/advanced.md
@@ -84,6 +84,12 @@ julia> @time ForwardDiff.gradient!(out, rosenbrock, x, cfg);
   0.281853 seconds (4 allocations: 160 bytes)
 ```
 
+The underlying heuristic will compute a suitable chunk size smaller or equal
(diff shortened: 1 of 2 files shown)


andreasnoack (approved):

andreasnoack (commented):
It would probably be a good idea to add this to the docs as well.

lassepe:
Thanks for the feedback, @andreasnoack. I added documentation to this feature. I should also say that as of https://github.com/FluxML/Zygote.jl/pull/1213, the chunk size is exposed in `Zygote.forwarddiff`. Therefore, the original use case for this feature is already fixed. Nevertheless, I think it may be a nice feature for other settings.

Feel free to merge or close as you see fit.

PR #585, opened 2022-05-06
Add definition for `logabsgamma`

devmotion:
This PR adds a definition of `SpecialFunctions.logabsgamma(::Dual)`. SpecialFunctions is a dependency of ForwardDiff already, and in constrast to most other functions in SpecialFunctions derivatives of `logabsgamma` can't be defined via DiffRules since `logabsgamma` is not a real-valued function but returns a tuple `(log(abs(gamma(x)), sign(gamma(x)))`.

### DIFF
--- a/src/dual.jl
+++ b/src/dual.jl
@@ -742,6 +742,15 @@ function LinearAlgebra.eigen(A::SymTridiagonal{<:Dual{Tg,T,N}}) where {Tg,T<:Rea
     Eigen(λ,Dual{Tg}.(Q, tuple.(parts...)))
 end
 
+# SpecialFunctions.logabsgamma           #
+# Derivative is not defined in DiffRules #
+#----------------------------------------#
+
+function SpecialFunctions.logabsgamma(d::Dual{T,<:Real}) where {T}
+    x = value(d)
+    y, s = SpecialFunctions.logabsgamma(x)
+    return (Dual{T}(y, SpecialFunctions.digamma(x) * partials(d)), s)
+end
 
 ###################
 # Pretty Printing #
--- a/test/DualTest.jl
+++ b/test/DualTest.jl
@@ -540,7 +540,8 @@ for N in (0,3), M in (0,4), V in (Int, Float32)
         @test dual_isapprox(f(PRIMAL, PRIMAL2, FDNUM3), Dual{TestTag()}(f(PRIMAL, PRIMAL2, PRIMAL3), PARTIALS3))
     end
 
-    @test all(map(dual_isapprox, logabsgamma(FDNUM), (loggamma(abs(FDNUM)), sign(gamma(FDNUM)))))
+    @test dual_isapprox(logabsgamma(FDNUM)[1], loggamma(abs(FDNUM)))
+    @test dual_isapprox(logabsgamma(FDNUM)[2], sign(gamma(FDNUM)))
 end
 
 @testset "Exponentiation of zero" begin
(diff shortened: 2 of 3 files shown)


src/dual.jl
```diff
@@ -742,6 +742,15 @@ function LinearAlgebra.eigen(A::SymTridiagonal{<:Dual{Tg,T,N}}) where {Tg,T<:Rea
...
 end
 
+# SpecialFunctions.logabsgamma           #
+# Derivative is not defined in DiffRules #
+#----------------------------------------#
+
+function SpecialFunctions.logabsgamma(d::Dual{T,<:Real}) where {T}
+    x = value(d)
+    y, s = SpecialFunctions.logabsgamma(x)
+    return (Dual{T}(y, SpecialFunctions.digamma(x) * partials(d)), s)
```
devmotion, line 752: Maybe this should return rather
```suggestion
    return (Dual{T}(y, SpecialFunctions.digamma(x) * partials(d)), Dual{T}(s, zero(s)))
```
to be consistent with `sign(::Dual)`?

oxinabox: it doesn't really matter one way or the other does it? promotion will fix it when it matters.

test/DualTest.jl
```diff
@@ -539,6 +539,8 @@ for N in (0,3), M in (0,4), V in (Int, Float32)
         @test dual_isapprox(f(FDNUM, PRIMAL2, PRIMAL3), Dual{TestTag()}(f(PRIMAL, PRIMAL2, PRIMAL3), PRIMAL2*PARTIALS))
         @test dual_isapprox(f(PRIMAL, PRIMAL2, FDNUM3), Dual{TestTag()}(f(PRIMAL, PRIMAL2, PRIMAL3), PARTIALS3))
     end
+
+    @test all(map(dual_isapprox, logabsgamma(FDNUM), (loggamma(abs(FDNUM)), sign(gamma(FDNUM)))))
```
oxinabox, line 543: Is the following the same:

```suggestion
    @test dual_isapprox(logabsgamma(FDNUM)[1], loggamma(abs(FDNUM)))
    @test dual_isapprox(logabsgamma(FDNUM)[2], sign(gamma(FDNUM)))
```
if it is the same then this seems clearer.

oxinabox (approved):
I'm not a regular committer here.
But the math on this checks out, 
and other than one comment on style, I think the tests look sufficient.

I would say give this a little time, then merge it if noone else raising objections.
Maybe 1 more day, since it has been open 7 days already.

src/dual.jl
devmotion: No, I don't think it matters. It felt easier for downstream computations and somewhat stronger to drop the partials completely. Only afterwards I noticed the inconsistency with `sign` and started to wonder if we should return a zero partial instead.

devmotion: Hmm.... Maybe we should be consistent with `sign` at all since it makes a difference in an example such as:
```julia
julia> using ForwardDiff

julia> f(x) = Inf * x

julia> g(x) = sign(x)

julia> g(x::ForwardDiff.Dual) = sign(ForwardDiff.value(x))
g (generic function with 2 methods)

julia> ForwardDiff.derivative(f ∘ g, 1.0)
0.0

julia> ForwardDiff.derivative(f ∘ sign, 1.0)
NaN
```

The "stronger" definition that drops the partial completely doesn't run into the NaN-safe mode issues and hence does not care about the non-finite result.

Maybe it's an argument for defining
```julia
Base.sign(x::Dual{<:Any,<:Real}) = sign(value(x))
```
rather than changing this PR though.

PR #587, opened 2022-05-22
Define `SpecialFunctions.gamma_inc` for `ForwardDiff.Dual`

devmotion:
This PR adds a definition of `SpecialFunctions.gamma_inc(::Real, ::Dual, ::Integer)`.

Similar to `logabsgamma` (added in https://github.com/JuliaDiff/ForwardDiff.jl/pull/585) this function returns a tuple and hence is not handled by DiffRules.

As in the CR definitions in SpecialFunctions, only the derivatives with respect to the second argument are defined since derivatives wrt to the first argument are given in terms of (regularized) hypergeometric functions and hence would require additional dependencies (see https://github.com/JuliaMath/SpecialFunctions.jl/issues/317). The third argument indicates the desired accuracy, and hence the function is not differentiable wrt to the third argument anyway.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ForwardDiff"
 uuid = "f6369f11-7733-5829-9624-2563aa707210"
-version = "0.10.30"
+version = "0.10.31"
 
 [deps]
 CommonSubexpressions = "bbf7d656-a473-5ed7-a52c-81e309532950"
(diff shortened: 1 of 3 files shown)


andreasnoack:
Any idea why this one is failing on nightly?

devmotion:
No, I'll have a look when I'm back from vacation next week.

devmotion:
I reran the tests and it seems they pass now. Maybe some recent change in Julia nightly or SpecialFunctions fixed the test errors.

andreasnoack (approved):

PR #592, opened 2022-08-04
check for `iszero(partials(x))` in pow

chriselrod:


chriselrod:
As suggested by @andreasnoack, I think it's also worth considering `StatsFuns.xlogy` in place of `x * log(y)` within `ForwardDiff`.

devmotion:
> As suggested by @andreasnoack, I think it's also worth considering `StatsFuns.xlogy` in place of `x * log(y)` within `ForwardDiff`.

As you might know, StatsFuns just reexports `LogExpFunctions.xlogy`, and since ForwardDiff already depends on LogExpFunctions this would not even require any additional dependencies.

andreasnoack (approved):

PR #599, opened 2022-09-17
use StaticArraysCore

### LINKED ISSUE #591: use StaticArraysCore.jl
if a package needs to do AD through a `StaticArray`, then they need to load `StaticArrays` anyway

mcabbott:
Closes #591. Should reduce loading time by a factor of 5 or so. Needs https://github.com/JuliaDiff/DiffResults.jl/pull/22 

Needs Julia 1.6. ~~What thoughts on dropping Julia 1.0 support that in a patch release?~~

~~The version with #481 is not yet tagged. If that does throw up any surprises, then best to be able to make a patch release correcting them... so we should require Julia 1.6 before tagging that, rather than just after?~~

Master is 0.11-DEV, and requires Julia 1.6.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -13,7 +13,7 @@ Preferences = "21216c6a-2e73-6563-6e65-726566657250"
 Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 SpecialFunctions = "276daf66-3868-5448-9aa4-cd146d93841b"
-StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
+StaticArraysCore = "1e83bf80-4336-4d27-bf5d-d5a4f845583c"
 
 [compat]
 Calculus = "0.5"
@@ -25,7 +25,8 @@ LogExpFunctions = "0.3"
 NaNMath = "1"
 Preferences = "1"
 SpecialFunctions = "1, 2"
-StaticArrays = "1.5"
+StaticArrays = "1.5.7"
+StaticArraysCore = "1.3.0"
 julia = "1.6"
 
 [extras]
@@ -33,7 +34,8 @@ Calculus = "49dc2e85-a5d0-5ad3-a950-438e2897f1b9"
 DiffTests = "de460e47-3fe3-5279-bb4a-814414816d5d"
 InteractiveUtils = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
+StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["Calculus", "DiffTests", "SparseArrays", "Test", "InteractiveUtils"]
+test = ["Calculus", "DiffTests", "SparseArrays", "StaticArrays", "Test", "InteractiveUtils"]
--- a/src/ForwardDiff.jl
+++ b/src/ForwardDiff.jl
@@ -2,7 +2,8 @@ module ForwardDiff
 
 using DiffRules, DiffResults
 using DiffResults: DiffResult, MutableDiffResult, ImmutableDiffResult
-using StaticArrays
+using StaticArraysCore
+using StaticArraysCore: StaticArray, StaticMatrix
 if VERSION >= v"1.6"
     using Preferences
 end
--- a/src/apiutils.jl
+++ b/src/apiutils.jl
@@ -18,14 +18,14 @@ end
 # vector mode function evaluation #
 ###################################
 
-@generated function dualize(::Type{T}, x::StaticArray) where T
-    N = _static_length(StaticArraysCore.Size(x))
+@generated function dualize(::Type{T}, x::S) where {T, S<:StaticArray}
+    N = _static_length(StaticArraysCore.Size(S))
     dx = Expr(:tuple, [:(Dual{T}(x[$i], chunk, Val{$i}())) for i in 1:N]...)
-    V = StaticArraysCore.similar_type(x, Dual{T,eltype(x),N})
     return quote
+        V = StaticArraysCore.similar_type(S, Dual{$T, $(eltype(x)), $N})
         chunk = Chunk{$N}()
         $(Expr(:meta, :inline))
-        return $V($(dx))
+        return V($(dx))
     end
 end
 
--- a/src/apiutils.jl
+++ b/src/apiutils.jl
@@ -21,7 +21,7 @@ end
 @generated function dualize(::Type{T}, x::StaticArray) where T
     N = length(x)
     dx = Expr(:tuple, [:(Dual{T}(x[$i], chunk, Val{$i}())) for i in 1:N]...)
-    V = StaticArrays.similar_type(x, Dual{T,eltype(x),N})
+    V = StaticArraysCore.similar_type(x, Dual{T,eltype(x),N})
     return quote
         chunk = Chunk{$N}()
         $(Expr(:meta, :inline))
--- a/src/apiutils.jl
+++ b/src/apiutils.jl
@@ -19,7 +19,7 @@ end
 ###################################
 
 @generated function dualize(::Type{T}, x::StaticArray) where T
-    N = length(x)
+    N = _static_length(StaticArraysCore.Size(x))
     dx = Expr(:tuple, [:(Dual{T}(x[$i], chunk, Val{$i}())) for i in 1:N]...)
     V = StaticArraysCore.similar_type(x, Dual{T,eltype(x),N})
     return quote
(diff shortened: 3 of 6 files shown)


mcabbott:
With these two PRs:
```julia
julia> @time_imports using ForwardDiff
      6.3 ms  IrrationalConstants
      0.9 ms  DiffRules
      2.7 ms  StaticArraysCore
      2.3 ms  DiffResults
     19.3 ms  Preferences
      1.2 ms  OpenLibm_jll
      0.3 ms  NaNMath
      0.1 ms  Compat
     82.6 ms  ChainRulesCore
      3.6 ms  DocStringExtensions 68.58% compilation time
      0.7 ms  ChangesOfVariables
      0.6 ms  InverseFunctions
      0.7 ms  LogExpFunctions
      0.3 ms  JLLWrappers
      2.9 ms  OpenSpecFun_jll 76.75% compilation time
     15.5 ms  SpecialFunctions
     27.6 ms  MacroTools
      0.3 ms  CommonSubexpressions
    101.4 ms  ForwardDiff
```
Before, StaticArrays completely dominates:
```julia
julia> @time_imports using ForwardDiff
      6.4 ms  IrrationalConstants
      0.7 ms  DiffRules
      2.4 ms  StaticArraysCore
    608.7 ms  StaticArrays
      2.4 ms  DiffResults
     20.7 ms  Preferences
      1.2 ms  OpenLibm_jll
...
```

Project.toml
```diff
@@ -13,27 +13,29 @@ Preferences = "21216c6a-2e73-6563-6e65-726566657250"
...
+DiffResults = "1.1.0"
 DiffRules = "1.4.0"
 DiffTests = "0.0.1, 0.1"
 LogExpFunctions = "0.3"
 NaNMath = "0.2.2, 0.3, 1"
 Preferences = "1"
 SpecialFunctions = "0.8, 0.9, 0.10, 1.0, 2"
-StaticArrays = "0.8.3, 0.9, 0.10, 0.11, 0.12, 1.0"
-julia = "1"
+StaticArrays = "1.5.7"
```
fredrikekre, line 28:
```suggestion
```

mcabbott: I guess that's fine, loading StaticArraysCore implies this bound during testing?

mcabbott:
CI gives errors like this, which don't show up when running simple examples locally:
```
  ...testing specialized StaticArray codepaths
Hessians: Error During Test at /home/runner/work/ForwardDiff.jl/ForwardDiff.jl/test/HessianTest.jl:106
  Test threw exception
  Expression: ForwardDiff.hessian(prod, sx) == actual
  MethodError: no method matching length(::Type{StaticArraysCore.SMatrix{3, 3, Float64, 9}})
  The applicable method may be too new: running in world age 32433, while current world is 32506.
  Closest candidates are:
    length(::Type{SA}) where SA<:(Union{LinearAlgebra.Adjoint{T, <:Union{StaticArraysCore.StaticArray{Tuple{var"#s2"}, T, 1} where var"#s2", ...) at ~/.julia/packages/StaticArrays/NOLon/src/abstractarray.jl:2 (method too new to be called from this world context.)
    length(::Union{LinearAlgebra.Adjoint{T, <:Union{StaticArraysCore.StaticArray{...) at ~/.julia/packages/StaticArrays/NOLon/src/abstractarray.jl:1 (method too new to be called from this world context.)
    ...
  Stacktrace:
   [1] #s24#84
     @ ~/work/ForwardDiff.jl/ForwardDiff.jl/src/apiutils.jl:22 [inlined]
   [2] var"#s24#84"(T::Any, ::Any, #unused#::Any, x::Any)
     @ ForwardDiff ./none:0
   [3] (::Core.GeneratedFunctionStub)(::Any, ::Vararg{Any})
     @ Core ./boot.jl:582
   [4] static_dual_eval(#unused#::Type{ForwardDiff.Tag{ForwardDiff.var"#118#119"{typeof(prod)}, Float64}}, f::ForwardDiff.var"#118#119"{typeof(prod)}, x::StaticArraysCore.SMatrix{3, 3, Float64, 9})
     @ ForwardDiff ~/work/ForwardDiff.jl/ForwardDiff.jl/src/apiutils.jl:32
   [5] vector_mode_jacobian(f::Function, x::StaticArraysCore.SMatrix{3, 3, Float64, 9})
     @ ForwardDiff ~/work/ForwardDiff.jl/ForwardDiff.jl/src/jacobian.jl:185
   [6] jacobian
     @ ~/work/ForwardDiff.jl/ForwardDiff.jl/src/jacobian.jl:85 [inlined]
   [7] hessian(f::Function, x::StaticArraysCore.SMatrix{3, 3, Float64, 9})
     @ ForwardDiff ~/work/ForwardDiff.jl/ForwardDiff.jl/src/hessian.jl:71
   [8] macro expansion
     @ /opt/hostedtoolcache/julia/1.8.1/x64/share/julia/stdlib/v1.8/Test/src/Test.jl:464 [inlined]
     ```

**Comment by devmotion** [comment]
My suspicion is that the type piracy in StaticArrays is actually a problem here: it defines functions (that it does not own) for types in StaticArraysCore (that it does not own) which are not available when the `@generated` functions in ForwardDiff are defined. IIRC I saw similar issues related to generated functions in DiffEqBase a while ago. It might be fine locally if you e.g. load StaticArrays before ForwardDiff.

**Comment by mcabbott** [comment]
Yes that's my guess too. I did not try very hard to replicate locally, but I did use ForwardDiff first, before loading StaticArrays. 

Scrolling through it looks like all are from `length(::Type{...})`, so just defining that in SACore ought to solve this?

File: Project.toml (hunk 1 above)
Comment (fredrikekre) [reply Project.toml]: I don't think it is allowed to have compat for non-dependencies. If this is needed you can keep the StaticArrays dependency (without `using` it).

**Comment by devmotion** [comment]
> Scrolling through it looks like all are from `length(::Type{...})`, so just defining that in SACore ought to solve this?

Yes, seems that might be sufficient for ForwardDiff. However, I guess these issues could also occur in other packages for different functions and hence ideally StaticArrays would fix these type piracies.

**Comment by fredrikekre** [comment]
The only way to fix (all of) them would be to merge the two packages again, no?

**Comment by devmotion** [comment]
Hmm, I am not familiar with the internals of StaticArrays but from a quick look it indeed seems that most of StaticArrays is type piracy currently - and hence, as far as I understand, potentially problematic when working with `@generated` functions. If one actually tries to fix that, maybe one approach could be to move basic types and definitions for Base functions to something like StaticArraysCore and definitions for stdlibs to separate packages (e.g. StaticArraysLinearAlgebra etc.)?

**Comment by thchr** [comment]
This seems to run counter to most of the point of having a *Core.jl package: the parent package _must_ commit a great deal of type-piracy. For StaticArrays, it must be nearly 100% since most of its functionality comes from extending Base/stdlib functions.

We could in principle put all `@generated` methods into StaticArraysCore.jl but that will come at the cost of increasing its load-time.
It seems a better path to try to find out how many of the relevant `@generated` methods indeed still need to be `@generated`.

**Comment by devmotion** [comment]
Sure, I see that one has to commit type piracy somewhere, so the ideal I mentioned above won't be possible.

However, with the current setup to me it seems that one can't use Base functionality of static arrays inside of `@generated` functions when only depending on StaticArraysCore. AFAICT the problem are not the generated functions in StaticArrays but generated functions in downstream packages, similar to the behaviour observed in https://github.com/SciML/DiffEqBase.jl/issues/507#issuecomment-629707067.

Maybe that could be avoided for Base functions by moving all definitions with Base functions to StaticArraysCore (without stdlibs) but maybe already that would increase compilation and loading times of StaticArraysCore too much...

**Comment by mcabbott** [comment]
> seems that one can't use Base functionality of static arrays inside of `@generated` functions 

It might be narrower than that. Functions which act on the instance can't be called until you have one, but functions which accept the type can be called outside the returned quote, and these are the problem. Minimal example:
```julia
# Core package
struct M1 end
foo(x::Int) = x

# Foreign package
@generated function gen(x::T) where T
  # :(foo(x))  # no problem
  foo(x)     # problem
  # :(foo($T)) # no problem
end
gen(1)

# Big package
foo(::M1) = 33
foo(::Type{M1}) = 99

# User
gen(M1())
```
The problem code here uses `similar_type` and `length` (although I think Base defines length on a type, so this is not a generic use):

https://github.com/JuliaDiff/ForwardDiff.jl/blob/c6dc2098b5201943cdb8433ee963078e6dcc7900/src/apiutils.jl#L21-L25

mcabbott:
I think this is done. The above `@generated` world-age problems are solved by not calling `length`, and by moving `similat_type` out of the generated body & into the quote.

KristofferC:
This could be made into an extension which would also solve the loading time.

mcabbott:
Maybe there would be less reason to create StaticArraysCore now that extensions exist. But unless there's a plan to deprecate the core package, maybe ForwardDiff may as well use it?

The next heaviest thing is ChainRulesCore, which isn't directly loaded. Perhaps all the packages for which ForwardDiff defines rules out to extensions would be a good idea.

fredrikekre:
> Maybe there would be less reason to create StaticArraysCore now that extensions exist. But unless there's a plan to deprecate the core package, maybe ForwardDiff may as well use it?

Hah, just posted https://github.com/JuliaArrays/StaticArraysCore.jl/issues/15. If people can agree on that plan, it might be easier to not port more packages to using `StaticArraysCore`.

mcabbott:
Is it certain the extension feature will make 1.9?

I see there are [22 packages](https://juliahub.com/ui/Packages/StaticArraysCore/Tzw28/1.4.0?page=2) using StaticArraysCore, but I suppose no problem to have some mixture of mechanisms for a while.

KristofferC:
> Is it certain the extension feature will make 1.9?

It's in the backport branch at least.

PR #605, opened 2022-11-11
dual exponent

ptiede:
This issue came up in https://github.com/FluxML/Zygote.jl/pull/1324#issuecomment-1311196153. 

The problem is that on Julia 1.6, the GPU was sensing it could hit a code branch where `exponent(::Dual)` was being called, and unfortunately, `exponent(::Dual)` does not exist, so the GPU threw a dynamic call error and gave up. This pull request fixes this.

Solving this problem is pretty easy since exponent: Real --> Int function, so you can take a similar path compared to `trunc` and define the Dual argument to return the primal value, which I did.

### DIFF
--- a/src/dual.jl
+++ b/src/dual.jl
@@ -332,6 +332,8 @@ Base.fld(x::Dual, y::Dual) = fld(value(x), value(y))
 
 Base.cld(x::Dual, y::Dual) = cld(value(x), value(y))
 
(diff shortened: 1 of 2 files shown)


mcabbott (approved):
Looks good, assuming tests pass

ptiede:
The tests passed! I think we are good to merge.

Issue #609, opened 2022-11-15
Violating assumptions about Dual comparisons


gpeairs:
There are already a few issues talking about #481 (#606, #607) and related problems (#480). I've been running into breakage hinted at in [this comment](https://github.com/JuliaDiff/ForwardDiff.jl/pull/481#issuecomment-1171251680) and [this workaround](https://github.com/JuliaDiff/ForwardDiff.jl/issues/607#issuecomment-1314299091) where
```jl
julia> x, y = 1, ForwardDiff.Dual(1,1);

julia> x < y || x > y || x == y
false

julia> x >= y && x <= y
true
```
This seems to be an issue for piecewise functions and wrapper types that assume `x <= y` is equivalent to `x == y || x < y`.

For example, something like this previously worked:
```jl
import ForwardDiff

struct WrappedNumber{T<:Number}
    value::T
end
for op in (:isless, :(==))
    @eval begin
        Base.$op(x::WrappedNumber, y::WrappedNumber) = ($op)(x.value, y.value)
        Base.$op(x::Number, y::WrappedNumber) = ($op)(x, y.value)
        Base.$op(x::WrappedNumber, y::Number) = ($op)(x.value, y)
    end
end
for op in (:*, :+, :-)
    @eval begin
        Base.$op(x::WrappedNumber, y::WrappedNumber) = WrappedNumber(($op)(x.value, y.value))
        Base.$op(x::Number, y::WrappedNumber) = WrappedNumber(($op)(x, y.value))
        Base.$op(x::WrappedNumber, y::Number) = WrappedNumber(($op)(x.value, y))
    end
end

function f_piecewise(x)
    lb = WrappedNumber(0.0)
    ub = WrappedNumber(1.0)
    (lb <= x < ub) && return 0.5*x*x
    x >= ub && return 0.5*ub*ub + (x - ub)
    return 0.0*x
end

function ForwardDiff.derivative(f, x::WrappedNumber{T}) where T
    r = f(WrappedNumber(ForwardDiff.Dual(x.value, one(T))))
    return ForwardDiff.partials(r.value)[1]
end
```
And now:
```jl
julia> ForwardDiff.derivative(f_piecewise, WrappedNumber(0.99))
0.99

julia> ForwardDiff.derivative(f_piecewise, WrappedNumber(1.01))
1.0

julia> ForwardDiff.derivative(f_piecewise, WrappedNumber(1.0))
0.0 # Used to be 1.0
```
[...]

PR #613, opened 2022-12-02
Version 0.11

mcabbott:
After https://github.com/JuliaRegistries/General/pull/73304 we should probably tag changes as 0.11.

This PR just updates the version number. It drops Julia 1.0 to enable #599 without further surprises.  

We may wish to include other things in 0.11:
* `<=` as in #609
* `isinteger` and friends: https://github.com/JuliaDiff/ForwardDiff.jl/pull/481#issuecomment-1171251680
* Any of this list: https://github.com/JuliaDiff/ForwardDiff.jl/pull/481#discussion_r866863465

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -16,16 +16,16 @@ SpecialFunctions = "276daf66-3868-5448-9aa4-cd146d93841b"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 
 [compat]
-Calculus = "0.2, 0.3, 0.4, 0.5"
+Calculus = "0.5"
 CommonSubexpressions = "0.3"
-DiffResults = "0.0.1, 0.0.2, 0.0.3, 0.0.4, 1.0.1"
-DiffRules = "1.4.0"
-DiffTests = "0.0.1, 0.1"
+DiffResults = "1.1"
+DiffRules = "1.4"
+DiffTests = "0.1"
 LogExpFunctions = "0.3"
-NaNMath = "0.2.2, 0.3, 1"
+NaNMath = "1"
 Preferences = "1"
-SpecialFunctions = "0.8, 0.9, 0.10, 1.0, 2"
-StaticArrays = "0.8.3, 0.9, 0.10, 0.11, 0.12, 1.0"
+SpecialFunctions = "1, 2"
+StaticArrays = "1.5"
 julia = "1.6"
 
 [extras]
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ForwardDiff"
 uuid = "f6369f11-7733-5829-9624-2563aa707210"
-version = "0.10.33"
+version = "0.11-DEV"
 
 [deps]
 CommonSubexpressions = "bbf7d656-a473-5ed7-a52c-81e309532950"
@@ -26,7 +26,7 @@ NaNMath = "0.2.2, 0.3, 1"
 Preferences = "1"
 SpecialFunctions = "0.8, 0.9, 0.10, 1.0, 2"
 StaticArrays = "0.8.3, 0.9, 0.10, 0.11, 0.12, 1.0"
-julia = "1"
+julia = "1.6"
 
 [extras]
 Calculus = "49dc2e85-a5d0-5ad3-a950-438e2897f1b9"


mcabbott:
Test failure on 1.8 is a tolerance issue.

<details>

```
436
       ...auto-testing SpecialFunctions.bessely with 2 arguments
[437](https://github.com/JuliaDiff/ForwardDiff.jl/actions/runs/3601997063/jobs/6068433780#step:6:440)
  ...testing Dual{TestTag(),Int64,3} and Dual{TestTag(),Dual{TestTag(),Int64,0},3}
[438](https://github.com/JuliaDiff/ForwardDiff.jl/actions/runs/3601997063/jobs/6068433780#step:6:441)
Dual{Z,Int64,3} and Dual{Z,Dual{Z,Int64,0},3}: Test Failed at /home/runner/work/ForwardDiff.jl/ForwardDiff.jl/test/DualTest.jl:596
[439](https://github.com/JuliaDiff/ForwardDiff.jl/actions/runs/3601997063/jobs/6068433780#step:6:442)
  Expression: ≈(partials(pq[i]), PARTIALS * Calculus.derivative((x->begin
[440](https://github.com/JuliaDiff/ForwardDiff.jl/actions/runs/3601997063/jobs/6068433780#step:6:443)
                    (gamma_inc(a, x, ind...))[i]
[441](https://github.com/JuliaDiff/ForwardDiff.jl/actions/runs/3601997063/jobs/6068433780#step:6:444)
                end), 1 + PRIMAL), rtol = tol)
[442](https://github.com/JuliaDiff/ForwardDiff.jl/actions/runs/3601997063/jobs/6068433780#step:6:445)
   Evaluated: Partials(2.911603324688561e-7, 3.8821377662514144e-7, 2.911603324688561e-7) ≈ Partials(2.91160034554385e-7, 3.8821337940584666e-7, 2.91160034554385e-7) (rtol=1.0e-6)
[443](https://github.com/JuliaDiff/ForwardDiff.jl/actions/runs/3601997063/jobs/6068433780#step:6:446)
Stacktrace:
```

</details>

devmotion:
I think before doing that (also since discussions and decisions might take some time) it would be useful to tag a non-breaking 0.10 release with https://github.com/JuliaDiff/ForwardDiff.jl/pull/605. That is, I think it would be good to revert https://github.com/JuliaDiff/ForwardDiff.jl/pull/481 on the master branch and not only yank the latest release. That would also allow us to include more non-breaking fixes and features in 0.10 that come up before this PR is merged.

In general, maybe it would be useful to backport bug fixes for some time to 0.10 once 0.11 is released due to the widespread use of 0.10 (sinilar to what we do in SpecialFunctions where currently bug fixes for version 2 are backported to 1).

mcabbott:
I think better just to make a branch for backports to 0.10, and then cherry pick whatever commits desired onto that. That's ends up with the right graph of changes, right? The branching and tagging can be done before 0.11.

Reverting seems (in my experience) to lead to changes rotting quietly somewhere. There was already far too much fighting with tests & re-basing in 481.

mcabbott:
There was already a release-0.3 branch from way back when, presumably for the same purpose. So now there's a new one:

https://github.com/JuliaDiff/ForwardDiff.jl/tree/release-0.10

j-fu:
Would love to see #615 (I think it is not breaking though formally changes API, ymmv...).

mcabbott:
#615 looks safe.

Something new we could consider for 0.11 is to use ["package extensions"]( https://github.com/JuliaLang/julia/pull/47695) rather than directly loading so many other packages.

mcabbott:
Merging since v0.10.34 is now tagged on [release-0.10](https://github.com/JuliaDiff/ForwardDiff.jl/tree/release-0.10) branch, https://github.com/JuliaRegistries/General/pull/73859 .

j-fu:
As for the package extensions: IMHO ForwardDiff loading many packages is no problem (compared to other packages). And for at least 1.6 backward compatibility one would have to [mimic them via Requires.jl](https://github.com/JuliaLang/Pkg.jl/pull/3264/files#diff-1af5f877eb4497fc1f22daf47044d0958aa02ab39cc6da8ef052624870d75d28), thus losing compat guarantees for the corresponding packages on those older versions AFAIK. So I think this would introduce brittleness into an otherwise rock stable (from my experience) package  with not much additional value.

mcabbott:
It could I think always load them on old versions (as now) rather than using Requires.

PR #615, opened 2022-12-08
Support specializing on functions

### LINKED ISSUE #516: type instabilities when  passing a function to jacobian!
Hi, I am encountering type instabilities whan calling `jacobian!`.  As I use this in the inner loop of a  finite volume operator assembly, this is serious performatnce hit. 

A remedy could  be to type-annotate the callback `f!. Here is an MWE. `myjacobian!` just modifies `jacobian!` 
(see 
https://github.com/JuliaDiff/ForwardDiff.jl/blob/909976d719fdbd5fec91a159c6e2d808c45a770f/src/jacobian.jl#L75 )
in this respect..

```
using ForwardDiff
using DiffResults

function myjacobian!(result::Union{AbstractArray,DiffResults.DiffResult}, f!::F, y::AbstractArray, x::AbstractArray, cfg::ForwardDiff.JacobianConfig{T} = ForwardDiff.JacobianConfig(f!, y, x), ::Val{CHK}=Val{true}()) where {F,T,CHK}
    CHK && ForwardDiff.checktag(T, f!, x)
    if ForwardDiff.chunksize(cfg) == length(x)
        ForwardDiff.vector_mode_jacobian!(result, f!, y, x, cfg)
    else
        ForwardDiff.chunk_mode_jacobian!(result, f!, y, x, cfg)
    end
    return result
end

function test()
    u=[1,2.0]
    function f(y,u)
        y[1]=2*u[1]-u[2]
        y[2]= u[2]
    end
    result=DiffResults.JacobianResult(u)
    y=similar(u)
    cfg=ForwardDiff.JacobianConfig(f, y,u)
    @time ForwardDiff.jacobian!(result,f,y,u,cfg)
    @time myjacobian!(result,f,y,u,cfg)
    DiffResults.jacobian(result)
end
```

The output of the second run is:

```
  0.000010 seconds (1 allocation: 64 bytes)
  0.000003 seconds
2×2 Matrix{Float64}:
 2.0  -1.0
 0.0   1.0
```
`vector_mode_jacobian` has this type annotation (and I use this in the moment as a drop-in replacement,
though it is not exported).

May be this needs to be changed for all similar API functions. Not sure however where this can have unintended consequences.

j-fu:
See [Performance Tips](https://docs.julialang.org/en/v1/manual/performance-tips/#Be-aware-of-when-Julia-avoids-specializing) .

In many places this was not implemented. This in particular would fix #516.

### DIFF
--- a/src/derivative.jl
+++ b/src/derivative.jl
@@ -22,8 +22,8 @@ stored in `y`.
 
 Set `check` to `Val{false}()` to disable tag checking. This can lead to perturbation confusion, so should be used with care.
 """
-@inline function derivative(f!, y::AbstractArray, x::Real,
-                            cfg::DerivativeConfig{T} = DerivativeConfig(f!, y, x), ::Val{CHK}=Val{true}()) where {T, CHK}
+@inline function derivative(f!::F, y::AbstractArray, x::Real,
+                            cfg::DerivativeConfig{T} = DerivativeConfig(f!, y, x), ::Val{CHK}=Val{true}()) where {F, T, CHK}
     CHK && checktag(T, f!, x)
     ydual = cfg.duals
     seed!(ydual, y)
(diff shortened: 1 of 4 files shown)


KnutAM:
I'm not sure what the release plan for releasing 0.11 is, is backporting this fix to 0.10 desired? 
(I can re-implement it on that branch if desired, I don't know if there is an easier way to do the backport)

PR #616, opened 2022-12-10
Remove `if VERSION ≥ v"1.4"` etc.

mcabbott:
since master only supports 1.6+

### DIFF
--- a/test/DualTest.jl
+++ b/test/DualTest.jl
@@ -178,10 +178,8 @@ ForwardDiff.:≺(::Type{OuterTestTag}, ::Type{TestTag}) = false
         @test div(NESTED_FDNUM, PRIMAL2) === div(PRIMAL, PRIMAL2)
         @test div(PRIMAL, NESTED_FDNUM2) === div(PRIMAL, PRIMAL2)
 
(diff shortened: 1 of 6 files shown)


test/DualTest.jl
```diff
@@ -589,7 +585,7 @@ ForwardDiff.:≺(::Type{OuterTestTag}, ::Type{TestTag}) = false
         @test pq isa Tuple{Dual{TestTag()},Dual{TestTag()}}
         # We have to adjust tolerances if lower accuracy is requested
         # Therefore we don't use `dual_isapprox`
-        tol = V === Float32 ? 5f-4 : 1e-6
+        tol = V === Float32 ? 5f-4 : 1e-5
```
devmotion, line 588: This seems unrelated?

devmotion:
Apart from the question above the PR seems fine :slightly_smiling_face:

test/DualTest.jl
mcabbott: I think this is an intermittent failure on 1.8, e.g. here:

https://github.com/JuliaDiff/ForwardDiff.jl/actions/runs/3601997063/jobs/6068433780#step:6:445

PR #619, opened 2022-12-20
make extract_gradient[_chunk]! GPU compatible

marius311:
Fixes this:
```julia
ForwardDiff.gradient(sum, cu(rand(10))) # Error: scalar indexing
```
My adhoc tests suggest no performance impact on 1.8.3. for `result::Vector`, however, there is an extra `(2 allocations: 80 bytes)` for `result::Matrix` I think due to a reshape/view thing. 

I'm not sure if there's some other form of the code that works without scalar indexing and without this allocation, but open to suggestions. Another alternative is to write some gluecode here or in CUDA.jl that handles CuArray specifically, but that seem more brittle. 

I'd vote this solution is OK (pending other feedback) for the benefit of CUDA compatibility with generic code, but thats just IMO.

### DIFF
--- a/src/gradient.jl
+++ b/src/gradient.jl
@@ -80,12 +80,12 @@ function extract_gradient!(::Type{T}, result::DiffResult, dual::Dual) where {T}
 end
 
 extract_gradient!(::Type{T}, result::AbstractArray, y::Real) where {T} = fill!(result, zero(y))
-extract_gradient!(::Type{T}, result::AbstractArray, dual::Dual) where {T}= copyto!(result, partials(T, dual))
+extract_gradient!(::Type{T}, result::AbstractArray, dual::Dual) where {T} =
+    extract_gradient_chunk!(T, result, dual, 1, npartials(dual))
 
 function extract_gradient_chunk!(::Type{T}, result, dual, index, chunksize) where {T}
-    offset = index - 1
-    for i in 1:chunksize
-        result[i + offset] = partials(T, dual, i)
+    map!(view(Base.ReshapedArray(result, (length(result),), ()), index:index+chunksize-1), 1:chunksize) do i
+        @inbounds partials(T, dual, i)
     end
     return result
 end
--- a/Project.toml
+++ b/Project.toml
@@ -32,8 +32,9 @@ julia = "1.6"
 Calculus = "49dc2e85-a5d0-5ad3-a950-438e2897f1b9"
 DiffTests = "de460e47-3fe3-5279-bb4a-814414816d5d"
 InteractiveUtils = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
+JLArrays = "27aeb0d3-9eb9-45fb-866b-73c2ecf80fcb"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["Calculus", "DiffTests", "SparseArrays", "Test", "InteractiveUtils"]
+test = ["Calculus", "DiffTests", "SparseArrays", "Test", "InteractiveUtils", "JLArrays"]
--- a/test/AllocationsTest.jl
+++ b/test/AllocationsTest.jl
@@ -37,4 +37,26 @@ convert_test_574() = convert(ForwardDiff.Dual{Nothing,ForwardDiff.Dual{Nothing,F
   
 end
 
+@testset "Test extract_gradient! allocations" begin
+    T = Float64
+    @testset "vector-mode size(result)=$size" for size in [(4,), (2,2)]
+        dual = ForwardDiff.Dual(0, (rand(T, size...)...,))
+        y = Array{T}(undef, size)
+        alloc = @allocated ForwardDiff.extract_gradient!(Nothing, y, dual)
+        alloc = @allocated ForwardDiff.extract_gradient!(Nothing, y, dual)
+        @test alloc == 0
+    end
+    @testset "chunk-mode size(result)=$size" for size in [(DEFAULT_CHUNK_THRESHOLD+1,), (DEFAULT_CHUNK_THRESHOLD+1, DEFAULT_CHUNK_THRESHOLD+1)]
+        Npartials = DEFAULT_CHUNK_THRESHOLD÷2
+        dual = ForwardDiff.Dual(0, (rand(T, Npartials...)...,))
+        y = Array{T}(undef, size)
+        alloc = @allocated ForwardDiff.extract_gradient_chunk!(Nothing, y, dual, 2, Npartials)
+        alloc = @allocated ForwardDiff.extract_gradient_chunk!(Nothing, y, dual, 2, Npartials)
+        @test alloc == 0
+        alloc = @allocated ForwardDiff.extract_gradient_chunk!(Nothing, y, dual, 2, Npartials-1)
+        alloc = @allocated ForwardDiff.extract_gradient_chunk!(Nothing, y, dual, 2, Npartials-1)
+        @test alloc == 0
+    end
+end
+
 end
--- a/test/GradientTest.jl
+++ b/test/GradientTest.jl
@@ -8,6 +8,8 @@ using ForwardDiff
 using ForwardDiff: Dual, Tag
 using StaticArrays
 using DiffTests
+using JLArrays
+JLArrays.allowscalar(false)
 
 include(joinpath(dirname(@__FILE__), "utils.jl"))
 
@@ -148,6 +150,26 @@ end
     @test isequal(ForwardDiff.gradient(t -> t[1]^t[2], [0.0,  0.5]), [Inf, NaN])
     @test isequal(ForwardDiff.gradient(t -> t[1]^t[2], [0.0,  1.5]), [0.0, 0.0])
 end
+
+
+##############################################
+# test GPUArray compatibility (via JLArrays) #
+##############################################
+
+println("  ...testing GPUArray compatibility (via JLArrays)")
+
+@testset "size = $(size(x))" for x in JLArray.([
+    rand(1), 
+    rand(DEFAULT_CHUNK_THRESHOLD+1), 
+    rand(1,1), 
+    rand(DEFAULT_CHUNK_THRESHOLD+1,DEFAULT_CHUNK_THRESHOLD+1), 
+    rand(1,1,1)
+])
+    
+    @test ForwardDiff.gradient(prod, x) isa typeof(x)
+
+end
+
 
 #############
 # bug fixes #


src/gradient.jl
```diff
@@ -80,13 +80,13 @@ function extract_gradient!(::Type{T}, result::DiffResult, dual::Dual) where {T}
 end
 
 extract_gradient!(::Type{T}, result::AbstractArray, y::Real) where {T} = fill!(result, zero(y))
-extract_gradient!(::Type{T}, result::AbstractArray, dual::Dual) where {T}= copyto!(result, partials(T, dual))
+function extract_gradient!(::Type{T}, result::AbstractArray, dual::Dual) where {T}
+    result[:] .= partials(T, dual)
```
mcabbott, line 84: Why the the `[:]`?
```
julia> @btime $(rand(3,3))[:] .= 0;
  min 27.973 ns, mean 35.812 ns (2 allocations, 80 bytes)

julia> @btime $(rand(3,3)) .= 0;
  min 4.541 ns, mean 4.649 ns (0 allocations)
```

mcabbott (commented):
This seems surely harmless.

Maybe it would be worth adding some tests, using JLArrays so that they can run on githun?

src/gradient.jl
marius311: Thats for the case where `result::Matrix` since `partials::AbstractVector`, otherwise its a size-mismatch error for those cases. That's exactly the allocations I was talking about above.

mcabbott: Oh I see. Will it work if it's `Base.ReshapedArray((rand(3,3)), (9,), ())  .= 0`?

devmotion: Or maybe
```suggestion
    vec(result) .= partials(T, dual)
```
or
```suggestion
    copyto!(result, partials(T, dual))
```
?

devmotion: Or
```suggestion
    fill!(result, partials(T, dual))
```
?

mcabbott: `vec(result) .=` is the same as `results[:] .=`, and `fill!(rand(3), (1,2,3))` won't work.

devmotion: What about
```suggestion
    copyto!(result, partials(T, dual))
```
?

marius311: That's exactly what it was before that doesn't work with CUDA :) 

I guess you could say its CUDA's fault they don't have `copyto!(::CuArray, ::AbstractArray)` or something, JLArrays doesn't seem to have it either fwiw. This generic version might enable more array types we're not thinking of right now. 

Here's three ideas: 

1. Keep PR as is and eat the allocation (only for `result::Matrix`, not `result::Vector`) in the name of simpler more generic code. 
2. Bring back old versions and use them for `::Array` and use my changes for `::AbstractArray`
3. Bring back old versions and use them for `::AbstractArray` and use my changes for `::AbstractGPUArray` (and then depend on that package)

Up to you guys.

marius311:
Added a barebones set of JLArray tests.

src/gradient.jl
devmotion: > I guess you could say its CUDA's fault they don't have copyto!(::CuArray, ::AbstractArray) or something, 

Yes, I would have assumed that CUDA defines `copyto!`, possibly via the broadcasting fallback. I think adding it to CUDA would be the optimal solution - it would not require any workarounds here and could be useful more generally.

marius311:
Ah, sorry, all the discussion above is moot, I just realized that even though 
```julia
julia> @btime $(rand(3,3))[:] .= 0;
  min 27.973 ns, mean 35.812 ns (2 allocations, 80 bytes)
```
in our actual case the RHS is a Partials and that broadcasts differently and without allocations both for chunk and non-chunk codepaths, as evidenced by the new allocations tests I just added. 

So overall theres no allocation regression here for any shaped Array result. Ready?

src/gradient.jl
marius311: Sorry, this is all moot point, see my comment below.

devmotion: As an additional comment:

> Keep PR as is and eat the allocation (only for result::Matrix, not result::Vector) in the name of simpler more generic code.

IMO arguably `copyto!` is the simplest and most generic code. In contrast to the broadcasting "hack" it is concise and a function that is supposed to do exactly what we want to achieve here. It should work regardless of whether the array supports scalar indexing or not. (Also broadcasting tends to cause increased compilation times.)

marius311: My problem is which packages' responsibility is `copyto!(::CuArray, ::Partials)` and if anything I'd argue its ForwardDiff's since Partials is the more limited-used array type (see also [this](https://github.com/JuliaGPU/GPUArrays.jl/issues/422#issuecomment-1207164226))

But maybe its a moot point, since I discovered this works without any allocations and for CuArrays: 
```julia
result[:] = partials(T, dual)
```

I'm still searching for something that works for the chunk mode though.

devmotion: > My problem is which packages' responsibility is copyto!(::CuArray, ::Partials) and if anything I'd argue its ForwardDiff's since Partials is the more limited-use array type (see also https://github.com/JuliaGPU/GPUArrays.jl/issues/422#issuecomment-1207164226)

IMO the natural thing to do would be to use `copyto!(result, partials(T, dual).values)` since in that case the second argument is just a `Tuple` which should work more generally.

src/gradient.jl
```diff
@@ -80,13 +80,13 @@ function extract_gradient!(::Type{T}, result::DiffResult, dual::Dual) where {T}
...
+    return result
+end
 
 function extract_gradient_chunk!(::Type{T}, result, dual, index, chunksize) where {T}
-    offset = index - 1
-    for i in 1:chunksize
-        result[i + offset] = partials(T, dual, i)
-    end
+    result[index:index+chunksize-1] .= partials.(T, dual, 1:chunksize)
```
devmotion, line 89: Similarly, maybe one approach here would be to use something like
```suggestion
    copyto!(result, index, partials(T, dual).values, 1, chunksize)
```
Again that would require that CUDA/GPUArrays defines `copyto!` for `Tuple`s.

marius311: JLArrays seems to have it but my GPU cluster is down atm to double check with CUDA, but this is promising!

src/gradient.jl
devmotion: > is just a `Tuple` which should work more generally.

It's even a `NTuple`, so support for heterogeneous tuples is not needed.

marius311:
Ok, found the magic code which gives no allocation or speed regressions vs. current master for Arrays of any dimensions and which works with CuArrays, its this:
```julia
map!(view(Base.ReshapedArray(result, (length(result),), ()), index:index+chunksize-1), 1:chunksize) do i
    @inbounds partials(T, dual, i)
end
```
Trick is indeed as @mcabbott you suggested to use a ReshapedArray which unlike `reshape` is non-allocating for Arrays (TIL). And CUDA.jl is smart enough to dispatch on `view(ReshapedArray(CuArray))` and do its thing. Hoping this approach works for jacobians too, which I'd like to tackle next.

Squashed into a single commit. Ready on my end.

src/gradient.jl
marius311: Thanks, yea, I do agree `copyto!` makes alot of logical sense for the operation we want, but I've got a solution below now which only relies on an even simper interface (basically get/set index). I can ask about the prospects of a `copyto!(::CuArray, ::Tuple)` though since it does seem to offer a good way to get stack-allocated memory onto GPU. If that ever happens we could then switch to `copyto!`.

marius311:
Don't want to have this lost after the holidays, could this get looked at again?

devmotion (commented):
To me this seems still mainly a CUDA issue. Changing the definitions to `copyto!` operating on `Tuple`s seems the most direct and natural implementation to me, and one that I would expect to work for any array. Whereas to me the code in this PR seems to be a bit non-intuitive and solely due to CUDA. So I would suggest
- using `copyto!` with Tuples (as suggested in more detail above)
- adding definitions of `copyto!` with `Tuple` to CUDA

marius311:
> adding definitions of copyto! with Tuple to CUDA

That seems like a strech to me with CUDA tbh. Although these things aren't specified perfectly, `copyto!` seems aimed to be an array interface, not really for tuples. I would expect such a PR not to be accepted in CUDA, but we can try. 

In the meantime, this PR has no regressions but enables GPU, so can it still be merged? And we can revist later if we get the CUDA interface?

marius311:
Here's a fun problem with the `copyto!` option. How do you implement `extract_gradient_chunk!(::Type{T}, result::AbstractArray, y::Real, index, chunksize)` so thats its inferred and allocation free for Arrays? Closest I can come up with is:

```julia
copyto!(result, index, ntuple(i->0, Val(chunksize)), 1, chunksize)
```

which is _not_ inferred, although the Julia compiler must work some magic bc its allocation free. But its 2X slower than my solution in this PR.

So in lieu of a solution, it seems even if we get `copyto!(::CuArray, ::Tuple)` (which fwiw was [suggested](https://julialang.slack.com/archives/C689Y34LE/p1674547552254489) on Slack that it might be possible), its not enough to write the code here generically.

marius311:
I should also that one nice thing about the `copyto!` option is that it would allow the specialized `copyto!(::CuArray, ::Tuple)` to pass the tuple data into the kernel by-reference as opposed by-value (which has a 4K memory limit, as pointed out on Slack). I'm not sure if that's a real worry except for chunk sizes that are so large theyd be unrealistic, but wanted to point that out. 

But in any case, I can't fiure out how to do it with only `copyto!` while avoiding regressions, c.f. the comment just above.

devmotion:
> How do you implement extract_gradient_chunk!(::Type{T}, result::AbstractArray, y::Real, index, chunksize) so thats its inferred and allocation free for Arrays?

With `fill!` as done in the definition of `extract_gradient!` for `y::Real`, I would assume?

marius311:
Thanks, `fill!` doesn't have a variant with an offset / chunk size though unless I'm mistaken?

devmotion:
> Thanks, fill! doesn't have a variant with an offset / chunk size though unless I'm mistaken?

No, not that I know of. But couldn't you `fill!` a `view`?

I still think that it would be nice if we wouldn't have to use slightly unorthodox implementations, just to be able to add implicit support for another package.

To mention the obvious, on Julia >= 1.9 yet another alternative would be to add e.g. GPUArraysCore (I guess that might be sufficient?) as a weak dependency and add a special method for GPU arrays there. The disadvantages would be that it would not fix Julia < 1.9.

marius311:
> No, not that I know of. But couldn't you fill! a view?

No, thats allocating for multidimensional Arrays, which is the entire issue that led to the solution currently in this PR.

marius311:
> I still think that it would be nice if we wouldn't have to use slightly unorthodox implementations, just to be able to add implicit support for another package.

Yea, I'm sympathetic to that, also bc the by-value thing, but ultimately although the code looks a little wierd, in the case of Array its purely relying on the get/set index interface, and in the case of CuArrays, on the `map!` interface, which are definitely very orthodox.

devmotion:
>  the map! interface, which are definitely very orthodox.

Oh, with unorthodox I was referring rather to the fact that currently the PR removes a simple `copyto!` statement and adds a workaround based on `ReshapedArray`.

marius311:
No I get that. All I'm saying is its unorthodox in terms of the usage of ReshapedArray sure, but in terms of the underlying native code, for Arrays, its compiled to nearly identical code as before this PR, since either this or the old for-loop all ultimately end up at the `setindex(::Array, ..)`.

devmotion:
> No, thats allocating for multidimensional Arrays, which is the entire issue that led to the solution currently in this PR.

We could use `fill!` + `Base.ReshapedArray` :smile: That seems to be faster than the current `for`-loop:
```julia
julia> function h2(::Type{T}, result, x, index, chunksize) where {T}
           fill!(view(Base.ReshapedArray(result, (length(result),), ()), index:(index + chunksize - 1)), zero(x))
           return result
       end
h2 (generic function with 1 method)

julia> @btime h2(Float64, $(rand(100, 2)), 0f0, 20, 15);
  2.570 ns (0 allocations: 0 bytes)

julia> function h3(::Type{T}, result, x, index, chunksize) where {T}
           offset = index - 1
           for i in 1:chunksize
               result[i + offset] = zero(x)
           end
           return result
       end
h3 (generic function with 1 method)

julia> @btime h3(Float64, $(rand(100, 2)), 0f0, 20, 15);
  5.338 ns (0 allocations: 0 bytes)
```

marius311:
Sorry I'm confused, if you're ok with ReshapedArray there, whats the issue with the current PR thought?

devmotion:
Well, I still think that 1) replacing the `copyto!` and dispatching to the chunk size-based implementation and 2) the `map!`+`ReshapedArray` are both very unnatural workarounds for a very specific package and use case. But of course I wanted to benchmark `Base.ReshapedArray` for the example. BTW, as a comparison the code in the PR seems to perform worse (than both examples):
```julia
julia> function h4(::Type{T}, result, x, index, chunksize) where {T}
           map!(view(Base.ReshapedArray(result, (length(result),), ()), index:(index + chunksize - 1)), 1:chunksize) do _
               return zero(x)
           end
           return result
       end
h4 (generic function with 1 method)

julia> @btime h4(Float64, $(rand(100, 2)), 0f0, 20, 15);
  10.412 ns (0 allocations: 0 bytes)
```

devmotion:
So it seems the PR actually introduces a regression in its current form?

marius311:
I had benchmarked some full gradient calls, and there I saw essentially no difference (presumably extracting the gradient is not a dominant part of the computation except for trivial functions). But you're right it looks like the `extract_gradient_chunk!` code alone is a little slower for `y::Real`, but also try the case `y::Dual`, on my machine its only like 30% slower (do the exact code in the PR):

```
# PR
julia> @btime ForwardDiff.extract_gradient_chunk!(Nothing, $(rand(100, 2)), $(ForwardDiff.Dual(rand(16)...)), 20, 15);
  10.052 ns (0 allocations: 0 bytes)

# master
julia> @btime ForwardDiff.extract_gradient_chunk!(Nothing, $(rand(100, 2)), $(ForwardDiff.Dual(rand(16)...)), 20, 15);
  8.530 ns (0 allocations: 0 bytes)
```

However, `copyto!` is like 5X slower than that on my machine, so thats not actually an alternative. 

So at the moment I guess it boils down to a small speed regression in `extract_gradient_chunk` which should negligibly impact realistic gradient calls in exchange for having this package work on GPU.

src/gradient.jl
```diff
@@ -80,12 +80,12 @@ function extract_gradient!(::Type{T}, result::DiffResult, dual::Dual) where {T}
 end
 
 extract_gradient!(::Type{T}, result::AbstractArray, y::Real) where {T} = fill!(result, zero(y))
-extract_gradient!(::Type{T}, result::AbstractArray, dual::Dual) where {T}= copyto!(result, partials(T, dual))
+extract_gradient!(::Type{T}, result::AbstractArray, dual::Dual) where {T} =
+    extract_gradient_chunk!(T, result, dual, 1, npartials(dual))
```
devmotion, line 84: I'm still somewhat unhappy about this change, intuitively it feels that it should be easier for base and packages to optimize `copyto!(::AbstractArray, ::Tuple)` than the chunk-based version. But if benchmarks don't show any regression, I guess it seems OK - and we don't have to wait for CUDA to add the implementation.

devmotion (commented):
OK, so what do you think about adding an implementation of `extract_gradient_chunk!(::Type{T}, result, dual::Real, index, chunksize)` using `fill!` with `ReshapedArray` which seems to be fast? The PR already uses `ReshapedArray` anyway, so if we go with it, we might as well just avoid the regression and use `ReshapedArray` once more.

And afterwards maybe check that there are really no regressions left?

vpuri3:
@marius311 is this still being worked on?

vpuri3:
Was this not merged earlier due to a regression on presently working cases? If so, then we can move this implementation to an extension package `ForwardDiffGPUArraysExt` which defines `ForwardDiff.extract_gradient_chunk!(::Type{T}, ::AbstractGPUArray, ...)`

marius311:
My summary from what I remember is what I said [above](https://github.com/JuliaDiff/ForwardDiff.jl/pull/619#issuecomment-1404248199):
> at the moment I guess it boils down to a small speed regression in extract_gradient_chunk which should negligibly impact realistic gradient calls in exchange for having this package work on GPU

I unfortunately ran out of time to dedicate to responding to the requests here but anyone is absolutely welcome to finish it off or close it and do your extension package solution.

vpuri3:
would that be a workable solution in your opinion, @devmotion?

vpuri3:
@devmotion

vpuri3:
@mcabbott what do you think about moving this change to a GPUArrays ext?

PR #632, opened 2023-03-07
Add a copy method for GradientConfig.

tpapp:
This is motivated by https://github.com/tpapp/LogDensityProblemsAD.jl/pull/8, where the user can supply a `GradientConfig`, but the unified API for thread-safe invocation would require that the resulting object can be `copy`ed.

Instead of relying on unexposed internals of this package, I thought it would be better to put this functionality here.


KristofferC:
Makes sense but shouldn't all configs have this method define then for consistency?

tpapp:
I just realized that for `::AbstractConfig` there is a `copy` that falls back to `deepcopy`, which should be sufficient. Closing.

PR #634, opened 2023-03-09
add back the iteration based seeding for `Array`

KristofferC:
These were removed in https://github.com/JuliaDiff/ForwardDiff.jl/pull/472 and while I was a bit uncomfortable with it I did not protest. However, now after doing some benchmarks I see that this has a runtime cost that is very much nonsignificant.

Below is a profile and the big peaks are from `seed!`. 

gnome-shell-screenshot-57lfk9

If I use the iteration-based seed function they completely disappear. In order to not break anything, I added these back but with a restriction to `Array` so that e.g. GPU arrays will keep using the broadcast based seeding.

### DIFF
--- a/src/apiutils.jl
+++ b/src/apiutils.jl
@@ -46,10 +46,26 @@ function seed!(duals::AbstractArray{Dual{T,V,N}}, x,
     return duals
 end
 
+function seed!(duals::Array{Dual{T,V,N}}, x,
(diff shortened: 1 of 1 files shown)


devmotion:
I wonder if one could use `map!` instead of broadcasting? I would assume that this would still be GPU-compatible but possibly there is (almost) no runtime cost for `Array`?

Issue #639, opened 2023-04-17
Type instability in jacobian! filling JacobianResult when using StaticArrays


mapclyps:
Running the following code
```julia
using ForwardDiff
using StaticArrays

f(x)=SA[x[1]^2+x[2]^2, x[2]^2+x[3]^2]

x=SA[1.0,2,3]

y=f(x)

imdr=DiffResults.JacobianResult(y,x)

@code_warntype ForwardDiff.jacobian!(imdr,f,x) 
```
shows up as not type stable for me in Julia 1.8.5. 
Maybe also related is this older topic on discourse https://discourse.julialang.org/t/am-i-using-diffresults-jl-correctly/35894

I think it is coming from
https://github.com/JuliaDiff/ForwardDiff.jl/blob/5bd71168f3ecbb542e5c700cd2f3d67051648f18/ext/ForwardDiffStaticArraysExt.jl#L108

Replacing the anonymous function `d -> value(T,d)` with `Base.Fix1(value,T)` fixes it for me.
I can make a PR in case this is the desired solution.

mcabbott:
> `Base.Fix1(value,T)` fixes it for me.

This seems completely safe, a PR would be great.

ojwoodford:
[Possibly related](https://discourse.julialang.org/t/change-in-behaviour-of-forwarddiff-in-julia-1-9/98616).

mapclyps:
I was trying to write a test for my proposed fix, but it did not have an inference issue in the tests. It seems like it depends on whether something else is executed before.
```julia
using ForwardDiff
using StaticArrays
using Test

sx = @SArray rand(3,3)
addone(A) = A.+one(eltype(A))
sy = addone(sx)

sresult = DiffResults.JacobianResult(sy, sx)
ff(A)=sin.(A)

@inferred ForwardDiff.jacobian!(sresult, ff, sx) #throws the error (type instability) as reported
@inferred ForwardDiff.jacobian!(sresult, addone, sx) #no error 
```
Swapping the last two lines, interestingly, removes the error entirely.
It seems that there is some hidden state or something?

KristofferC:
All of this seems to work now

PR #640, opened 2023-04-19
improve type stability with StaticArrays

mapclyps:
Replace an anonymous function with Base.Fix1 for type stability in some cases, see #639

### DIFF
--- a/ext/ForwardDiffStaticArraysExt.jl
+++ b/ext/ForwardDiffStaticArraysExt.jl
@@ -105,7 +105,7 @@ end
     T = typeof(Tag(f, eltype(x)))
     ydual = static_dual_eval(T, f, x)
     result = DiffResults.jacobian!(result, extract_jacobian(T, ydual, x))
-    result = DiffResults.value!(d -> value(T,d), result, ydual)
+    result = DiffResults.value!(Base.Fix1(value,T), result, ydual)
     return result
 end
 


devmotion (commented):
Thank you! Could you add a test for https://github.com/JuliaDiff/ForwardDiff.jl/issues/639?

mapclyps:
I tried to replicate the type instability in the tests and put a test at the end of JacobianTest.jl like 
```julia
ff(x) = sin.(x)
ssx = SA[1.0, 2.0]
ssy = ff(ssx)
result_store = DiffResults.JacobianResult(ssx)
@test begin
    result = (@inferred ForwardDiff.jacobian!(result_store, ff, ssx)) #the actual test
    jac = ForwardDiff.jacobian(ff, ssx)
    expected = DiffResults.ImmutableDiffResult(ssy, (jac,)) 
    result == expected
end
```
But to my surprise, there was no type instability with this test also on master.
 
However, just spinning up a REPL and executing this throws the error as expected and reported in #639.
Then I just moved this test to the top of the JacobianTest.jl file, and then it also threw the error on master (so it was working correctly as a test for the fix for #639). But it made me wonder what is actually going on there... and where this "hidden state" comes from. 

Or is there a better, more stable way to test it?

devmotion:
Fixed (and extended) in #735.

PR #644, opened 2023-04-25
Allow failures on nightly

andreasnoack:
While it's good to test on nightly, we shouldn't allow it to cause overall failure for the PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,11 +18,15 @@ jobs:
         version:
           - '1.6'
           - '1'
-          - 'nightly'
         os:
           - ubuntu-latest
         arch:
           - x64
+        include:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -18,11 +18,15 @@ jobs:
...
-          - 'nightly'
         os:
           - ubuntu-latest
         arch:
           - x64
+        include:
+          - version: 'nightly'
+            os: ubuntu-latest
+            arch: x64
+            continue-on-error: true
```
devmotion, line 29: Isn't that already set above in line 14?

andreasnoack: Yeah. I thought this would avoid the overall status to become red but that is apparently not possible

PR #650, opened 2023-06-07
Add test of second-order derivative of `t -> abs(t)^2`

devmotion:
This PR adds the example in https://github.com/JuliaDiff/DiffRules.jl/pull/98#issuecomment-1574420052 as a test, to avoid that upstream reintroduces this bug (DiffRules runs downstream tests of ForwardDiff).

### DIFF
--- a/test/MiscTest.jl
+++ b/test/MiscTest.jl
@@ -158,4 +158,7 @@ end
 @test ForwardDiff.derivative(x -> rem2pi(x, RoundUp), rand()) == 1
 @test ForwardDiff.derivative(x -> rem2pi(x, RoundDown), rand()) == 1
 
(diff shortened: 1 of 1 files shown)


andreasnoack:
Any idea why we'd see these test failures?

devmotion:
Not completely sure but I assume on Julia 1.9 it might just be due to different random numbers as only https://github.com/JuliaDiff/ForwardDiff.jl/actions/runs/5198679278/jobs/9375117936?pr=650#step:6:615 is failing. I don't know what's going on with nightly and the invalidations action.

devmotion:
I guess the test failures were indeed random, tests on Julia 1.9 passed successfully after rerunning them.

PR #652, opened 2023-06-07
use `oneunit(T)` in `construct_seeds`

### LINKED ISSUE #651: `construct_seeds` for types where `typeof(one(T)) !=T` is broken

longemen3000:
should fix #651 ?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -38,9 +38,10 @@ julia = "1.6"
 Calculus = "49dc2e85-a5d0-5ad3-a950-438e2897f1b9"
 DiffTests = "de460e47-3fe3-5279-bb4a-814414816d5d"
 InteractiveUtils = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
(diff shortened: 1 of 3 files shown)


longemen3000:
the invalidations tests has an unrelated error

mcabbott:
This seems fine. Is it easy to add a test, e.g. loading the smallest / slowest-changing package which defines a number for which one != oneunit?

longemen3000:
Measurements? and an unary function. (binary functions require defining some interactions between `ForwardDiff.Dual` and `Measurements.Measurement`, i had a package that defines some of those in https://github.com/longemen3000/ForwardDiffOverMeasurements.jl, but that is not necessary for the test).

longemen3000:
I added Measurements.jl as an example package that has `oneunit(T) != one(T)`.

The invalidation tests fails with the `stat_rep(::Nothing)` method error, and there are errors on nightly, that i don't know if are unrelated.

longemen3000:
replaced by #659

PR #570, opened 2021-12-13
WIP: use SIMD.jl for partials

KristofferC:
This needs quite well testing for downstream users to ensure regressions haven't crept in due to the implementation in SIMD.jl.

PR #655, opened 2023-07-06
Merge master, inline retval

chriselrod:
Adding the `@inline`s
```julia
julia> @benchmark ForEach(expm!, $B, $As)()
BenchmarkTools.Trial: 8692 samples with 1 evaluation.
 Range (min … max):   88.607 μs …   1.523 ms  ┊ GC (min … max):  0.00% … 87.83%
 Time  (median):      95.456 μs               ┊ GC (median):     0.00%
 Time  (mean ± σ):   110.611 μs ± 120.857 μs  ┊ GC (mean ± σ):  11.04% ±  9.27%

  █      ▁                                                      ▁
  █▇▄▁▄▄▅█▆▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▄▄▆ █
  88.6 μs       Histogram: log(frequency) by time       1.12 ms <

 Memory estimate: 545.41 KiB, allocs estimate: 119.
```
Prior to that, performance was around 215 microseconds.


chriselrod:
@KristofferC ?

chriselrod:
Running some internal proprietary benchmark, the latest ForwardDiff release gives me
```julia
 18.568555 seconds (24.68 M allocations: 1.288 GiB, 2.00% gc time, 2.42% compilation time)
  7.463995 seconds (24.35 M allocations: 1.268 GiB, 2.79% gc time, 0.14% compilation time)
 18.064215 seconds (24.40 M allocations: 1.270 GiB, 1.72% gc time)
  7.450395 seconds (24.35 M allocations: 1.268 GiB, 2.67% gc time)
```
versus this PR
```julia
 13.992152 seconds (24.72 M allocations: 1.295 GiB, 2.51% gc time, 3.18% compilation time)
  6.132790 seconds (24.45 M allocations: 1.277 GiB, 4.11% gc time, 0.18% compilation time)
 13.443605 seconds (24.44 M allocations: 1.277 GiB, 2.11% gc time)
  6.045382 seconds (24.45 M allocations: 1.277 GiB, 3.19% gc time)
```
The difference is partially because my computer has AVX512.

If I start Julia with `-C'native,-prefer-256-bit'` to allow the autovectorizer to use 512 bit vectors, the latest ForwardDiff release gives me
```julia
 15.560895 seconds (24.68 M allocations: 1.288 GiB, 2.31% gc time, 2.70% compilation time)
  6.524974 seconds (24.35 M allocations: 1.268 GiB, 3.69% gc time, 0.17% compilation time)
 15.263915 seconds (24.40 M allocations: 1.270 GiB, 2.00% gc time)
  6.641746 seconds (24.35 M allocations: 1.268 GiB, 3.04% gc time)
```
while this PR is unchanged (as it isn't relying on the autovectorizer).

Compile times appear unchanged.
This PR is an easy free performance win.

I am not sure if retval not being inlined caused the problem we observed before the initial reverting, but I would like to see this move forward.

Or I'm just going to give up and create my own ForwardDiff, since duplicated effort tends to be smaller than collaboration's cost.

KristofferC:
I missed this. Wouldn't the "correct" way to do this be to rebase the original branch on master, force push that commit and then add this above. As it is right now, it is kind of unreviewable.

KristofferC:
> Or I'm just going to give up and create my own ForwardDiff,

Hehe, yeah, replicating ForwardDiff is pretty fast. I tried something with Hessians using SIMD.jl in https://github.com/KristofferC/HyperHessians.jl with some decent perf results.

src/dual.jl
```diff
@@ -195,6 +195,38 @@ macro define_ternary_dual_op(f, xyz_body, xy_body, xz_body, yz_body, x_body, y_b
...
+    reval, imval = reim(val)
+    if deriv isa Real
+        p = deriv * partial
+        return Complex(Dual{T}(reval, p), Dual{T}(imval, zero(p)))
+    else
+        rederiv, imderiv = reim(deriv)
+        return Complex(Dual{T}(reval, rederiv * partial), Dual{T}(imval, imderiv * partial))
+    end
+end
+@inline function dual_definition_retval(::Val{T}, val::Complex, deriv1::Union{Real,Complex}, partial1::Partials, deriv2::Union{Real,Complex}, partial2::Partials) where {T}
```
chriselrod, line 215: @KristofferC adding `@inline` here should be the only actual change.
I can't rebase your original PR. Mind doing so?

Perhaps it'd be better to rebase on the 0.10 branch than on master (although you could do master if you'd prefer having that be a breaking release for safety reasons).

You're then free to add these 4 `@inline`s directly, or seeing if this diff looks cleaner afterwards.

KristofferC: I updated the `kc/simd` branch, please check it out.

chriselrod:
Closing because this PR is no longer necessary (#570 incorporates the change).

PR #657, opened 2023-08-06
example in docs

cossio:
The ForwardDiff docs are a bit strange: Introduction has no usage example, first entry of "User Documentation" is Limitations, then API reference, and then an Advanced section. Would be nice to start instead with a simple usage example.

I just copy/pasted the example from the README. I think this will be helpful for users that land on the docs page of the package without having seen the readme.


gdalle:
Thanks for the contribution!

PR #659, opened 2023-08-09
use `oneunit(T)` in `construct_seeds`

### LINKED ISSUE #651: `construct_seeds` for types where `typeof(one(T)) !=T` is broken

longemen3000:
replaces #652 , fixes #651

### DIFF
--- a/src/dual.jl
+++ b/src/dual.jl
@@ -360,6 +360,9 @@ end
 @inline Base.one(d::Dual) = one(typeof(d))
 @inline Base.one(::Type{Dual{T,V,N}}) where {T,V,N} = Dual{T}(one(V), zero(Partials{N,V}))
 
+@inline Base.oneunit(d::Dual) = oneunit(typeof(d))
+@inline Base.oneunit(::Type{Dual{T,V,N}}) where {T,V,N} = Dual{T}(oneunit(V), zero(Partials{N,V}))
+
 @inline function Base.Int(d::Dual)
     all(iszero, partials(d)) || throw(InexactError(:Int, Int, d))
     Int(value(d))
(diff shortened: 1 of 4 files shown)


src/dual.jl
```diff
@@ -360,6 +360,9 @@ end
 @inline Base.one(d::Dual) = one(typeof(d))
 @inline Base.one(::Type{Dual{T,V,N}}) where {T,V,N} = Dual{T}(one(V), zero(Partials{N,V}))
 
+@inline Base.oneunit(d::Dual) = oneunit(typeof(d))
```
devmotion, line 363: I'm not sure if there are any examples of numeric types where `oneunit` is only defined for instances but not for types (could imagine unitful numbers where the unit is not a type parameter). In such cases falling back on the definition for `Dual` here would fail. Therefore I think the more correct implementation would be
```suggestion
@inline Base.oneunit(d::Dual{T}) where {T} = Dual{T}(oneunit(primal(d)), zero(partials(d)))
```
(IMO the same reasoning applies also to other functions already defined in dual.jl)

longemen3000: Fair enough

longemen3000: There is exactly this in DynamicQuantities.jl:
https://github.com/SymbolicML/DynamicQuantities.jl/blob/1d62cec8a7df1867b240f8c5d9183063c0a00a41/src/utils.jl#L107

longemen3000: Hmm, the suggestion fails with `UndervarError: primal not defined`, maybe is `value` instead of `primal`?

devmotion: Ah sorry, you're probably right, I guess I misremembered the function name.

longemen3000:
The error in Julia latest seems unrelated?

longemen3000:
bump?

RomeoV:
Could use a rebase onto master. That makes the errors go away for me.

longemen3000:
superseded by #750

PR #662, opened 2023-09-05
Add Aqua tests

charleskawczynski:
This PR adds Aqua tests.

### DIFF
--- a/test/AquaTest.jl
+++ b/test/AquaTest.jl
@@ -1,0 +1,30 @@
+using Test
+using ForwardDiff
+using Aqua
+
+@testset "Aqua tests (performance)" begin
+    # This tests that we don't accidentally run into
+    # https://github.com/JuliaLang/julia/issues/29393
+    ua = Aqua.detect_unbound_args_recursively(ForwardDiff)
+    @test length(ua) == 6
+
+    # See: https://github.com/SciML/OrdinaryDiffEq.jl/issues/1750
+    # Test that we're not introducing method ambiguities across deps
+    ambs = Aqua.detect_ambiguities(ForwardDiff; recursive = true)
+    pkg_match(pkgname, pkdir::Nothing) = false
+    pkg_match(pkgname, pkdir::AbstractString) = occursin(pkgname, pkdir)
+    filter!(x -> pkg_match("ForwardDiff", pkgdir(last(x).module)), ambs)
+
+    @test length(ambs) == 0
+end
+
+@testset "Aqua tests (additional)" begin
+    Aqua.test_undefined_exports(ForwardDiff)
+    Aqua.test_stale_deps(ForwardDiff)
+    Aqua.test_deps_compat(ForwardDiff)
+    Aqua.test_project_extras(ForwardDiff)
+    Aqua.test_project_toml_formatting(ForwardDiff)
+    Aqua.test_piracy(ForwardDiff)
+end
+
+nothing
--- a/Project.toml
+++ b/Project.toml
@@ -35,6 +35,7 @@ StaticArrays = "1.5"
 julia = "1.6"
 
 [extras]
+Aqua = "4c88cf16-eb10-579e-8560-4a9242c79595"
 Calculus = "49dc2e85-a5d0-5ad3-a950-438e2897f1b9"
 DiffTests = "de460e47-3fe3-5279-bb4a-814414816d5d"
 InteractiveUtils = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
@@ -43,4 +44,4 @@ StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["Calculus", "DiffTests", "SparseArrays", "StaticArrays", "Test", "InteractiveUtils"]
+test = ["Aqua", "Calculus", "DiffTests", "SparseArrays", "StaticArrays", "Test", "InteractiveUtils"]
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -51,5 +51,10 @@ Random.seed!(SEED)
         t = @elapsed include("AllocationsTest.jl")
         println("##### done (took $t seconds).")
     end
+    @testset "Aqua" begin
+        println("##### Testing aqua...")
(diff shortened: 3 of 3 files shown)


test/AquaTest.jl
```diff
@@ -0,0 +1,30 @@
...
+    @test length(ambs) == 0
+end
+
+@testset "Aqua tests (additional)" begin
+    Aqua.test_undefined_exports(ForwardDiff)
+    Aqua.test_stale_deps(ForwardDiff)
+    Aqua.test_deps_compat(ForwardDiff)
+    Aqua.test_project_extras(ForwardDiff)
+    Aqua.test_project_toml_formatting(ForwardDiff)
+    Aqua.test_piracy(ForwardDiff)
```
hyrodium, line 27:
```suggestion
    Aqua.test_piracies(ForwardDiff)
```
The function was renamed by https://github.com/JuliaTesting/Aqua.jl/pull/230.
It would be better to have `[comapt]` for Aqua.jl.

test/AquaTest.jl
```diff
@@ -0,0 +1,30 @@
...
+    @test length(ambs) == 0
+end
+
+@testset "Aqua tests (additional)" begin
+    Aqua.test_undefined_exports(ForwardDiff)
+    Aqua.test_stale_deps(ForwardDiff)
+    Aqua.test_deps_compat(ForwardDiff)
+    Aqua.test_project_extras(ForwardDiff)
+    Aqua.test_project_toml_formatting(ForwardDiff)
+    Aqua.test_piracies(ForwardDiff)
```
devmotion, line 27: Can we just run
```suggestion
    Aqua.test_all(ForwardDiff; ambiguities = false)
```
?

charleskawczynski: Yes, that looks good to me.

test/AquaTest.jl
```diff
@@ -0,0 +1,27 @@
...
+    filter!(x -> pkg_match("ForwardDiff", pkgdir(last(x).module)), ambs)
+
+    @test length(ambs) == 0
+end
+
+@testset "Aqua tests - remaining" begin
+    Aqua.test_all(ForwardDiff; ambiguities = false, unbound_args = false)
+end
+
+nothing
```
hyrodium, line 27: I personally don't much prefer it, but it seems better to surround the code with `module` for consistency with other test julia scripts. https://github.com/JuliaDiff/ForwardDiff.jl/blob/master/test/PartialsTest.jl

hyrodium (commented):
Could you please add the following badge to README.md?
```md
[![Aqua QA](https://raw.githubusercontent.com/JuliaTesting/Aqua.jl/master/badge.svg)](https://github.com/JuliaTesting/Aqua.jl)
```

[Aqua QA](https://github.com/JuliaTesting/Aqua.jl)

charleskawczynski:
I'd appreciate feedback on what to do with the tests, here. From what I can see, it seems like CI is broken on master

test/AquaTest.jl
```diff
@@ -0,0 +1,29 @@
...
+using Test
+using ForwardDiff
+using Aqua
+
+@testset "Aqua tests - unbound_args" begin
+    # This tests that we don't accidentally run into
+    # https://github.com/JuliaLang/julia/issues/29393
+    ua = Aqua.detect_unbound_args_recursively(ForwardDiff)
+    @test length(ua) == 6
+end
```
devmotion, line 12: Ideally the unbound arguments should be fixed, which would also allow us to just run the default `Aqua.test_all` test with `unbound_args = true` below instead of this custom test.

test/AquaTest.jl
```diff
@@ -0,0 +1,29 @@
...
+
+@testset "Aqua tests - ambiguities" begin
+    # See: https://github.com/SciML/OrdinaryDiffEq.jl/issues/1750
+    # Test that we're not introducing method ambiguities across deps
+    ambs = Aqua.detect_ambiguities(ForwardDiff; recursive = true)
+    pkg_match(pkgname, pkdir::Nothing) = false
+    pkg_match(pkgname, pkdir::AbstractString) = occursin(pkgname, pkdir)
+    filter!(x -> pkg_match("ForwardDiff", pkgdir(last(x).module)), ambs)
+
+    @test length(ambs) == 0
```
devmotion, line 22: Seems this test fails: https://github.com/JuliaDiff/ForwardDiff.jl/actions/runs/7412737730/job/20170158384?pr=662#step:6:1288

test/AquaTest.jl
charleskawczynski: These ambiguities seem to be upstream issues. See https://github.com/JuliaTesting/Aqua.jl/issues/86 and https://github.com/JuliaLang/julia/issues/28086.

We can "fix" this (I'm not 100% sure if the issue is here or upstream, even based on discussions) by changing e.g.,

```julia
@generated function one_tuple(::Type{NTuple{N,V}}) where {N, V}
```
to
```julia
@generated function one_tuple(::Type{Tuple{V,Vararg{R1,N}}) where {N, V}
```
Should I move forward with those changes?

test/AquaTest.jl
charleskawczynski: I wasn't able to reproduce this locally for 1.10.

One of the ambiguities that I'm seeing on 1.6, `@inline Base.convert(::Type{Dual{T,V,N}}, x) where {T,V,N} = Dual{T}(V(x), zero(Partials{N,V}))`, can be eliminated, but we would need to depend on ChainRulesCore, and define `@inline Base.convert(::Type{Dual{T,V,N}}, x::ChainRulesCore.AbstractZero) where {T,V,N} = Dual{T}(V(x), zero(Partials{N,V}))` in order to do so.

I'm not sure that adding the dependency is worth it? I suppose we could add it as an extension, but I'd rather address that in a followup PR.

devmotion: Just run the Aqua tests first, before StaticArrays etc are loaded (I think StaticArrays or some other test dependency might pull in ChainRulesCore?). If only ForwardDiff, Aqua, and Test are loaded such indirect issues should not show up.

charleskawczynski: Done, let's see what happens. Hmm, I'm getting the same thing locally (6 ambiguities).

charleskawczynski:
bump

charleskawczynski:
bump

PR #664, opened 2023-09-18
make one(::Dual) return underlying Real type

### LINKED ISSUE #624: one(::type(Dual))
Currently, `one` returns a `Dual` number.  It seems like it could simply return the underlying value type instead.  i.e. change [these lines](https://github.com/JuliaDiff/ForwardDiff.jl/blob/4b143a199f541e7c1248dc5bc29b397be938e81d/src/dual.jl#L356-L357) to
```
Base.one(::Type{Dual{T,V,N}}) where {T,V,N} = one(V)
Base.oneunit(::Type{Dual{T,V,N}}) where {T,V,N} = Dual{T}(oneunit(V), zero(Partials{N,V}))
```
since this is still a multiplicative identity for `Dual`, and type promotion takes care of the rest.   In general, `one(x)` may not not return the same type as `x` — if you want the same type, you use `oneunit`, which is why I added a method.

The advantage of this is that code that is designed to "strip the units" from numbers will then work with `Dual` numbers as well.  For example, [this code in QuadGK.jl](https://github.com/JuliaMath/QuadGK.jl/blob/b4803e50f03fa3c0db69b1fc73d7a38a00a0f652/src/gausskronrod.jl#L264-L265) will work, and you will suddenly be able to differentiate numerical integrals:
```
julia> ForwardDiff.derivative(x -> quadgk(y -> y^2, 0, x)[1], 4)
16.0
```

stevengj:
Fixes #624

### DIFF
--- a/src/dual.jl
+++ b/src/dual.jl
@@ -353,7 +353,9 @@ end
 @inline Base.zero(::Type{Dual{T,V,N}}) where {T,V,N} = Dual{T}(zero(V), zero(Partials{N,V}))
 
 @inline Base.one(d::Dual) = one(typeof(d))
-@inline Base.one(::Type{Dual{T,V,N}}) where {T,V,N} = Dual{T}(one(V), zero(Partials{N,V}))
+@inline Base.oneunit(d::Dual) = oneunit(typeof(d))
+@inline Base.one(::Type{Dual{T,V,N}}) where {T,V,N} = one(V)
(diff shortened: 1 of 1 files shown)


stevengj:
Closing based on discussion in https://github.com/SymbolicML/DynamicQuantities.jl/issues/40

ajwheeler:
@stevengj did you mean to close this?

PR #665, opened 2023-09-18
use oneunit, not one, to preserve type

### LINKED ISSUE #651: `construct_seeds` for types where `typeof(one(T)) !=T` is broken

stevengj:
Fixes #651

### DIFF
--- a/src/partials.jl
+++ b/src/partials.jl
@@ -7,7 +7,7 @@ end
 ##############################
 
 @generated function single_seed(::Type{Partials{N,V}}, ::Val{i}) where {N,V,i}
-    ex = Expr(:tuple, [ifelse(i === j, :(one(V)), :(zero(V))) for j in 1:N]...)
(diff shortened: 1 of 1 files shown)


giordano:
Bump

devmotion (approved):
Looks good to me. If possible, it would be good to add a test.

giordano:
@longemen3000 maybe you can help coming up with a simple test?

longemen3000:
This is an stripped down version of the `Furlong` number type used in [julia tests](https://github.com/JuliaLang/julia/blob/0e5d6aefe27afc14d76647700b6c650ff4657d03/test/testhelpers/Furlongs.jl) ?

longemen3000:
i could add another PR with a test, just including a Furlong.jl file (the same as the one used in the julia tests) and then testing:

```
furlong = Furlongs.Furlong{2}(1.0)
f(x) = exp(x) + 4*sin(x)*oneunit(x)
@test ForwardDiff.derivative(f,furlong )= exp(furlong) + 4*cos(furlong)
```
does that seem alright?

stemann:
@devmotion ?

PR #667, opened 2023-09-28
Add flexibility in dispatch for `iszero_tuple`

timholy:
IntervalArithmetic.jl may abandon support for `==` among intervals (https://github.com/JuliaIntervals/IntervalArithmetic.jl/pull/571). To support specialization for specific Number subtypes, this makes `iszero_tuple` into a "trait"-dispatched function, first unwrapping all the way down to the elementary numeric type and then jointly dispatching on that type and the actual tuple. This makes it possible to create an extension in IntervalArithmetic that specializes the implementation to use the new comparison operator `≛`. The use of recursive unwrapping enables support for higher-order derivatives.

This doesn't pass tests locally, but neither does `master`, and the failures look identical.

### DIFF
--- a/src/dual.jl
+++ b/src/dual.jl
@@ -25,6 +25,9 @@ end
 ##########
 Base.ArithmeticStyle(::Type{<:Dual{T,V}}) where {T,V} = Base.ArithmeticStyle(V)
 
+unwrap_dual(::Type{Dual{T,V,N}}) where {T,V,N} = unwrap_dual(V)
+unwrap_dual(::Type{V}) where V = V
+
 ##############
 # Exceptions #
 ##############
--- a/src/partials.jl
+++ b/src/partials.jl
@@ -163,7 +163,11 @@ end
 @inline rand_tuple(::AbstractRNG, ::Type{Tuple{}}) = tuple()
 @inline rand_tuple(::Type{Tuple{}}) = tuple()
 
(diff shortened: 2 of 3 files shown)


OlivierHnt:
Is there someone specific we can ping to revive this PR?
Context: the bottom line is that we want a small hook/helper function that specifically checks unambiguously whether a quantity is exactly zero (similarly to how it is in SparseArrays). I did not look at this PR specifically, but the changes should be very minimal.

EDIT: after looking at it, this PR should be updated to include other zero checks, such as https://github.com/JuliaDiff/ForwardDiff.jl/blob/090ddbbb7ac10f5a0cc7decfce6cc331bd1df7b3/src/dual.jl#L566-L575.

OlivierHnt:
Not sure who to ping about the internal of ForwardDiff..

@devmotion , @KristofferC , would you consider merging a PR creating a helper function `_iszero(x) = iszero(x)` that can be overloaded by other numerical types (in our case, `iszero` can return an error)?
Basically mimicking what was done in SparseArrays:
https://github.com/JuliaSparse/SparseArrays.jl/blob/ce9fe7818cdfa3e4391cc1cb2587a058dbdba1d8/src/SparseArrays.jl#L42-L50

Side note: maybe in this repo a `_check_equality` helper function will also be needed.

Issue #671, opened 2023-11-15
Correctly forming nested dual numbers.


vpuri3:
MWE
```julia
using ForwardDiff
f = x -> x .^ 2                                                                               
x = [1.0, 2.0, 3.0, 4.0]                                                                               
v = ones(4)                                                                     
```

```julia
# 1st order (works)                                                                     
y = ForwardDiff.Dual.(x, v)                                                     
fy = f(y)                                                                       
@assert ForwardDiff.value.(fy) == [1.0, 4.0, 9.0, 16.0]                         
@assert ForwardDiff.partials.(fy, 1) == [2.0, 4.0, 6.0, 8.0]
```

```julia
# 2nd order (incorrect 2nd derv)
using ForwardDiff: Dual, partials, value
z = Dual.(Dual.(x, v), Dual.(v, v))
fz = f(z)
```
```julia
4-element Vector{Dual{Nothing, Dual{Nothing, Float64, 1}, 1}}:
  Dual{Nothing}(Dual{Nothing}(1.0,2.0),Dual{Nothing}(2.0,4.0))
  Dual{Nothing}(Dual{Nothing}(4.0,4.0),Dual{Nothing}(4.0,6.0))
  Dual{Nothing}(Dual{Nothing}(9.0,6.0),Dual{Nothing}(6.0,8.0))
 Dual{Nothing}(Dual{Nothing}(16.0,8.0),Dual{Nothing}(8.0,10.0))
```
Notice that the 2nd derivative (which should be [2.0, 2.0, 2.0, 2.0]
```julia
julia> partials.(partials.(fz, 1), 1)
4-element Vector{Float64}:
  4.0
  6.0
  8.0
 10.0
```
is off exactly by the first derivative.
```julia
julia> partials.(partials.(fz, 1), 1) - partials.(value.(fz), 1)
4-element Vector{Float64}:
 2.0
 2.0
 2.0
 2.0
```

I noticed this pattern in another function: `f = x -> exp.(x)`:
```julia
4-element Vector{Dual{Nothing, Dual{Nothing, Float64, 1}, 1}}:
  Dual{Nothing}(Dual{Nothing}(2.718281828459045,2.718281828459045),Dual{Nothing}(2.718281828459045,5.43656365691809))
  Dual{Nothing}(Dual{Nothing}(7.38905609893065,7.38905609893065),Dual{Nothing}(7.38905609893065,14.7781121978613))
[...]

PR #672, opened 2023-11-15
Update how_it_works.md

vpuri3:
Add code for forming nested duals in the docs so folks don't make the mistakes I made in https://github.com/JuliaDiff/ForwardDiff.jl/issues/671

### DIFF
--- a/docs/src/dev/how_it_works.md
+++ b/docs/src/dev/how_it_works.md
@@ -56,8 +56,9 @@ learn about [hyper-dual numbers](https://arc.aiaa.org/doi/abs/10.2514/6.2011-886
 which extend dual numbers to higher orders by introducing extra ``\epsilon`` terms that can
 cross-multiply. ForwardDiff's `Dual` number implementation naturally supports hyper-dual
 numbers without additional code by allowing instances of the `Dual` type to nest within each
-other. For example, a second-order hyper-dual number has the type `Dual{T,Dual{S,V,M},N}`, a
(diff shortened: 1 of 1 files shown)


docs/src/dev/how_it_works.md
```diff
@@ -56,8 +56,9 @@ learn about [hyper-dual numbers](https://arc.aiaa.org/doi/abs/10.2514/6.2011-886
 which extend dual numbers to higher orders by introducing extra ``\epsilon`` terms that can
 cross-multiply. ForwardDiff's `Dual` number implementation naturally supports hyper-dual
 numbers without additional code by allowing instances of the `Dual` type to nest within each
-other. For example, a second-order hyper-dual number has the type `Dual{T,Dual{S,V,M},N}`, a
-third-order hyper-dual number has the type `Dual{T,Dual{S,Dual{R,V,K},M},N}`, and so on.
+other. For example, a second-order hyper-dual number has the type `Dual{T,Dual{S,V,M},N}`,
+and can be formed as follows: `Dual(Dual(x, one(eltype(x))), one(eltype(x)))`. Athird-order
```
hyrodium, line 60: Typo.
```suggestion
and can be formed as follows: `Dual(Dual(x, one(eltype(x))), one(eltype(x)))`. A third-order
```

I prefer putting `\n` at the ends of sentences for easy to see the diff, but that will not be consistent with the whole document.
```suggestion
and can be formed as follows: `Dual(Dual(x, one(eltype(x))), one(eltype(x)))`.
A third-order hyper-dual number has the type `Dual{T,Dual{S,Dual{R,V,K},M},N}`, and so on.
```

PR #678, opened 2023-12-19
Add single-pass value_and_derivative(s) functions

### LINKED ISSUE #610: Is this code a supported use? (single-pass value and derivative)
Considering these two methods that compute the value and first derivative of a scalar function in a single pass:

```jl
import ForwardDiff
import DiffResults

@inline function value_and_derivative(f, Y::Type, x::Real)
    diffresult = ForwardDiff.derivative!(DiffResults.DiffResult(zero(Y), zero(Y)), f, x)
    return DiffResults.value(diffresult), DiffResults.derivative(diffresult)
end

@inline function value_and_derivative(f, x::Real)
    T = typeof(ForwardDiff.Tag(f, typeof(x)))
    y = f(ForwardDiff.Dual{T}(x, one(x)))
    return y.value, ForwardDiff.extract_derivative(T, y)
end
```

* The first method seems to be the officially recommended way to do this. However, (1) it introduces the `DiffResults` dependency just for this, and (2) it needlessly requires the caller to specify `f`'s return type ahead of the call. Neither are dealbreakers, but IMO they add friction for something that looks like it shouldn't have it (intuition says that the value comes for free when computing a derivative with `ForwardDiff`)

* The second method uses the `Tag` and `Dual` types as well as the `extract_derivative` function, which are not listed in the "Differentiation API" in the docs, so I'm not sure if they're considered part of the stable public API

Both methods run equally fast (and significantly faster than a naive two-pass implementation), so my question is: **does the second method constitute a supported use of `ForwardDiff`'s public API?**

If such an use isn't supported (but maybe even if it is so—both method implementations appear too convoluted for a pretty common use case IMO), I'd like to suggest adding a `value_and_derivative` function with the second method to either this or one of the other packages in `JuliaDiff` (I'm willing to write a PR).

Related: https://github.com/JuliaDiff/ForwardDiff.jl/issues/401

gerlero:
Closes #610.

## Proposal
I propose adding a `value_and_derivative` function for easy single-pass computation of value and derivative that has the same signature as `derivative` but returns a plain tuple. I also propose a `value_and_derivatives` function that also includes the second derivative. I'm not proposing methods for even higher order derivatives (honestly because I don't have a need for those), but if needed these could be added in the future via an optional argument (e.g. a`Val{n}`) in `value_and_derivatives`.

## Rationale
The rationale is the one I presented in #610. To mention it here, using the `DiffResults` API (which is intended for these use cases) for very simple scalar derivatives is somewhat convoluted, crucially requiring knowledge of the return types of the differentiated function ahead of time instead of just being able to let the compiler deal with the types.

This has caused implementations of functions like `value_and_derivative` to appear in other packages (including mine); but these have to rely on internals of `ForwardDiff` (i.e., `Dual` and related functions), which is not ideal.

## Naming
I considered the names `value_and_derivative` and `value_derivative`. I decided on the former as that's the [name that was chosen in the `AbstractDifferentiation` package](https://github.com/JuliaDiff/AbstractDifferentiation.jl/blob/211b67528c5ed91971bb524c57adb63837163367/src/AbstractDifferentiation.jl#L106C16-L106C16).

## Background
Initially I wanted to register a wrapper package with this functionality, but Registry maintainers asked me to consider a PR here instead. I agree and I hope these changes can be merged here.

### DIFF
--- a/docs/src/user/api.md
+++ b/docs/src/user/api.md
@@ -9,6 +9,8 @@ CurrentModule = ForwardDiff
 ```@docs
 ForwardDiff.derivative
 ForwardDiff.derivative!
+ForwardDiff.value_and_derivative
+ForwardDiff.value_and_derivatives
 ```
 
 ## Gradients of `f(x::AbstractArray)::Real`
--- a/src/derivative.jl
+++ b/src/derivative.jl
@@ -75,6 +75,32 @@ end
 derivative(f, x::AbstractArray) = throw(DimensionMismatch("derivative(f, x) expects that x is a real number. Perhaps you meant gradient(f, x)?"))
 derivative(f, x::Complex) = throw(DimensionMismatch("derivative(f, x) expects that x is a real number (does not support Wirtinger derivatives). Separate real and imaginary parts of the input."))
 
+"""
+    ForwardDiff.value_and_derivative(f, x::Real)
+
+Return `f(x)` and `df/dx` evaluated at `x` in a single pass, assuming `f` is called as `f(x)`.
+
+This method assumes that `isa(f(x), Union{Real,AbstractArray})`.
+"""
+@inline function value_and_derivative(f::F, x::R) where {F,R<:Real}
+    T = typeof(Tag(f, R))
+    ydual = f(Dual{T}(x, one(x)))
+    return value(T, ydual), extract_derivative(T, ydual)
+end
+
+"""
+    ForwardDiff.value_and_derivatives(f, x::Real)
+
+Return `f(x)` and its first and second derivatives evaluated at `x` in a single pass, assuming `f` is called as `f(x)`.
+
+This method assumes that `isa(f(x), Union{Real,AbstractArray})`.
+"""
+@inline function value_and_derivatives(f::F, x::R) where {F,R<:Real}
+    T = typeof(Tag(f, typeof(x)))
+    ydual, ddual = value_and_derivative(f, Dual{T}(x, one(x)))
+    return value(T, ydual), value(T, ddual), extract_derivative(T, ddual)
+end
+
 #####################
 # result extraction #
 #####################
--- a/test/DerivativeTest.jl
+++ b/test/DerivativeTest.jl
@@ -22,11 +22,22 @@ for f in DiffTests.NUMBER_TO_NUMBER_FUNCS
     v = f(x)
     d = ForwardDiff.derivative(f, x)
     @test isapprox(d, Calculus.derivative(f, x), atol=FINITEDIFF_ERROR)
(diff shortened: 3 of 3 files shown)


gerlero:
CI failures seem to be unrelated to these changes

devmotion:
> I decided on the former as that's the [name that was chosen in the `AbstractDifferentiation` package](https://github.com/JuliaDiff/AbstractDifferentiation.jl/blob/211b67528c5ed91971bb524c57adb63837163367/src/AbstractDifferentiation.jl#L106C16-L106C16).

I wonder if `AbstractDifferentation.value_and_derivative` should be implemented in this way (which might make this PR obsolete?).

gerlero:
> I wonder if `AbstractDifferentation.value_and_derivative` should be implemented in this way (which might make this PR obsolete?).

If it's acceptable for that package to rely on these ForwardDiff internals, I don't mind making the change in there. Just, can we add a higher-order `value_and_derivatives` function to that package? AFAICT there's no way to get values and two derivatives in a single pass using only documented functions.

gerlero:
> Just, can we add a higher-order value_and_derivatives function to that package? AFAICT there's no way to get values and two derivatives in a single pass using only documented functions.

Sorry, I failed to notice that `AsbtractDifferentiation` already defines an interface for multiple derivatives (the signature is `value_and_derivative(ab::AD.AbstractBackend, f, xs::Number...)`, BTW can this be type-stable?). So, this part could be ignored as ideally it would only require changes in the backend.

gerlero:
I had a few minutes to play around with making these changes in `AbstractDifferentiation` instead (to complete the interface I also added `second_derivative` and `value_and_second_derivative; which I guess I could add to this PR too).

While implementing the changes in `AbstractDifferentiation` is obviously viable, I have to say that that package's interface doesn't feel really comfortable when working with simple scalar functions (the nested tuple returns make debugging and getting it right in the first try more difficult). For this reason plus the already mentioned fact that the implementation relies on what are ForwardDiff internals, I'd really prefer that these functions be added here in `ForwardDiff` if at all possible.

gerlero:
> For this reason plus the already mentioned fact that the implementation relies on what are ForwardDiff internals, I'd really prefer that these functions be added here in ForwardDiff if at all possible.

Gentle bump @devmotion; are you willing to consider adding these functions to this package? (regardless of whether equivalent functions go into AbstractDifferentiation too)

devmotion:
IMO there is no need for duplicate definitions, in particular that the long-term goal of AbstractDifferentation is to move the extensions to the relevant packages (i.e., Zygote, Tracker, ForwardDiff etc.).

Issue #328, opened 2018-07-02
Unitful support


rafaqz:
Unitful and ForwardDiff don't play well together. But it would be good if they did! stripping the units out of a model to run ForwardDiff for a jacobian etc. isn't always feasible, in my case it would mean adding a lot of manual conversions and tests.

Loosening some Real to Number would get some of the way, but there are deeper problems where ForwardDiff re-wraps a Unitful.Quantity of ForwardDiff.Dual.

rafaqz:
Ok there seems to be an easy solution. We just need to strip the units before returning the results vector, which could happen in userland manually by dispatching on Dual number inputs. There are some problems with this but its workable.

But to get to that point a lot of Real Needs to be changed to Number:

```
@inline partials(x::Number) = Partials{0,typeof(x)}(tuple())
@inline partials(d::Dual) = d.partials
@inline partials(x::Number, i...) = zero(x)
```

I'll put together a pull request if you think it's worth having. Maybe all the Real should be Number to avoid this issue in future.

cstjean:
+1. To provide a concrete example of the failure:

```julia
julia> using Unitful, ForwardDiff

julia> f(x::Vector) = sum(sin, x) + prod(tan, x) * sum(sqrt, x);

julia> x = [1.0u"m", 2.0u"m^2"]
2-element Array{Quantity{Float64,D,U} where U where D,1}:
   1.0 m
 2.0 m^2

julia> g = x -> ForwardDiff.gradient(f, x); # g = ∇f

julia> g(x)
ERROR: MethodError: no method matching ForwardDiff.Partials(::Tuple{Float64,Quantity{Float64,Unitful.Dimensions{()},Unitful.FreeUnits{(),Unitful.Dimensions{()}}}})
Closest candidates are:
  ForwardDiff.Partials(::Tuple{Vararg{V,N}}) where {N, V} at /home/cst-jean/.julia/packages/ForwardDiff/hnKaN/src/partials.jl:2
Stacktrace:
 [1] macro expansion at /home/cst-jean/.julia/packages/ForwardDiff/hnKaN/src/partials.jl:0 [inlined]
 [2] single_seed(::Type{ForwardDiff.Partials{2,Quantity{Float64,D,U} where U where D}}, ::Val{1}) at /home/cst-jean/.julia/packages/ForwardDiff/hnKaN/src/partials.jl:10
[...]

PR #680, opened 2023-12-21
Support derivatives of functions with units via Unitful extension

gerlero:
Adds support for taking derivatives of functions with `Unitful` units on their input and/or output by means of an extension (weakdep). The extension works by ensuring that `Dual` always wraps a `Real` number—as `ForwardDiff` expects—even when `Unitful.Quantity`s are involved, stripping and adding back the units as appropriate.

Relevant issue is https://github.com/JuliaDiff/ForwardDiff.jl/issues/328.

With this PR only `derivative` has support for units, yet the same behavior could be later applied to the other functions as well.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -14,12 +14,15 @@ Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 SpecialFunctions = "276daf66-3868-5448-9aa4-cd146d93841b"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
+Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
 
 [weakdeps]
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
+Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
 
 [extensions]
 ForwardDiffStaticArraysExt = "StaticArrays"
+ForwardDiffUnitfulExt = "Unitful"
 
 [compat]
 Calculus = "0.5"
@@ -32,6 +35,7 @@ NaNMath = "1"
 Preferences = "1"
 SpecialFunctions = "1, 2"
 StaticArrays = "1.5"
+Unitful = "1"
 julia = "1.6"
 
 [extras]
@@ -41,6 +45,7 @@ InteractiveUtils = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
 
 [targets]
-test = ["Calculus", "DiffTests", "SparseArrays", "StaticArrays", "Test", "InteractiveUtils"]
+test = ["Calculus", "DiffTests", "SparseArrays", "StaticArrays", "Test", "Unitful", "InteractiveUtils"]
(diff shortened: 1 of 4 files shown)


gdalle:
Thank you for the contribution! Can I ask why you would prefer this code to be in ForwardDiff?
My justification for submitting a Unitful extension is that ForwardDiff is the more "generic" / "mainstream" package, and so Unitful should strive to adapt to it rather than the other way around. Does that make sense?

gerlero:
@gdalle Sure enough. While the same code would also work as an extension for `Unitful`, my reasoning for submitting it here is that it is using `ForwardDiff` internals (i.e., the `Dual` API, which is not documented).

Alternative implementations of the extension that only extend the public API (e.g., adding a new method to `derivative`) that I could come up with can't cover all bases, as a derived function with units on its output (which the extension needs to strip from the `Dual` return before `ForwardDiff` is able to process it) may have accept dimensionless (`Real`) input, and therefore we can't rely on dispatch for that.

gerlero:
Soft bump to the maintainers; interested in knowing if you're able to take these changes.

gdalle:
While I am a member of the JuliaDiff organization, I have never really maintained this repo so I probably won't be the one to merge this.
In my mind conceptually this belongs in Unitful. I agree that the names `value`, `derivative`, `extract_derivative`, `Dual` and `Tag` are not documented, but I'm willing to bet many packages actually rely on those internals too. So maybe the right question to ask would be https://github.com/JuliaDiff/ForwardDiff.jl/issues/682

jariji:
What are the assumptions underneath this? For example does it require that the function and its derivative have the same units at a point?

jariji:
I think this does the wrong thing on some functions like `sin` for [DimensionfulAngles.jl](https://github.com/cmichelenstrofer/DimensionfulAngles.jl/). Is it making certain assumptions about the derivative?

For reference I am looking at ["On the dimension of angles and their units"](https://iopscience.iop.org/article/10.1088/1681-7575/ac7bc2/pdf) which derives a good meaning of `sin` for unitful arguments.

Demonstration follows.

Just loosening the `derivative` bound from `Real` to `Number` and defining `can_dual`:

```jl
using ForwardDiff, Unitful, DimensionfulAngles

# Define a cycle unit equal to 360 degrees.
@unit cyc "cyc" Cycle 360u"°ᵃ" false


ForwardDiff.can_dual(::Type{Quantity{Float64}}) = true
function deriv(f::F, x::R) where {F,R<:Number}
    T = typeof(ForwardDiff.Tag(f, R))
    return ForwardDiff.extract_derivative(T, f(ForwardDiff.Dual{T}(x, one(x))))
end

julia> deriv(sin, 2π)
1.0

julia> deriv(sin, 1.0cyc)
1.0

julia> deriv(sin, 2π*u"rad")
1.0

julia> deriv(sin, 360.0*u"°ᵃ")
1.0
```

Versus this PR:

```jl
julia> ForwardDiff.derivative(sin, 0)
1.0

julia> ForwardDiff.derivative(sin, 1.0cyc)
6.283185307179586 cyc^-1

julia> ForwardDiff.derivative(sin, 2π*u"rad")
1.0 rad^-1

julia> ForwardDiff.derivative(sin, 360.0*u"°ᵃ")
0.017453292519943295 °^-1
```

gerlero:
@jariji My point with this PR was to support `Unitful` (which is what I use). I'm not familiar with `DimensionfulAngles`, but if small changes can be made here to support it then it should be done; otherwise I guess that can be left to a different PR.

jariji:
DimensionalAngles.jl just defines some more dimensions and units for Unitful.jl and their behavior in Base functions.

It's possible I'm misunderstanding the correct behavior -- I'm not an expert in quantities or autodiff. However, Imho a function that gives right answers for some units and wrong answers for some units is probably not safe to merge because that kind of decision is the source of Julia's many [correctness problems](https://yuri.is/not-julia/).

Ickaser:
Worth mentioning that as of https://github.com/JuliaPhysics/Unitful.jl/pull/765 earlier this year, Unitful has a ForwardDiff extension (which is currently used only to fix a very specific bug, but it's there). So insofar as the (IMO very valid) concerns about correctness can be resolved, I agree that this fits more naturally in Unitful than in ForwardDiff.

Issue #682, opened 2024-01-03
Document internals?


gdalle:
Is there a reason why types like `Dual` and `Tag` or functions like `value` and `derivative` are left undocumented?

gerlero:
Related issue: #610

PR #683, opened 2024-01-04
Update for Documenter.jl v1

hyrodium:


docs/src/user/api.md
```diff
@@ -60,3 +60,9 @@ ForwardDiff.GradientConfig
 ForwardDiff.JacobianConfig
 ForwardDiff.HessianConfig
 ```
+
+## Other API
+```@docs
+ForwardDiff.can_dual
+ForwardDiff.:≺
+```
```
devmotion, line 68: The docstring for `can_dual` is not useful at all, and both functions are not exported. IMO including them in the docs should be discussed and proposed in separate issues (#682) and PRs, but not be part of upgrading to Documenter v1.

devmotion (commented):
The doc build fails currently because `docs/Project.toml` is missing a compat entry for Documenter. Could you add one to avoid that a future breaking update of Documenter breaks the docs again?

hyrodium:
>The doc build fails currently because `docs/Project.toml` is missing a compat entry for Documenter. 

The [doc build](https://github.com/JuliaDiff/ForwardDiff.jl/actions/runs/7412439719/job/20169237582?pr=683) is not failing, but I agree with adding `[compat]` for future breaking changes in Documenter, and I have added the `[compat]`.

docs/src/user/api.md
hyrodium: Documenter.jl v1 has `warnonly=false` keyword argument in `makedocs`. (https://github.com/JuliaDocs/Documenter.jl/pull/2194), so `makedocs` will fail without adding the docs or setting `warnonly=true`.

These functions are already documented as docstrings, and I thought it would be okay to add them to the documentation. I understand these functions are not stable.
Should I replace "Other API" with "Internal API"?

devmotion: AFAIK you can set `checkdocs = :exports`.

hyrodium: Yeah, that's right.
My point is, that the ignored docstrings may confuse other users. The functions are documented as docstrings, so someone may think these functions are stable. I thought emphasizing that these functions are not stable would help that kind of people.

devmotion: But they are not exported, are they?

hyrodium: `ForwardDiff.gradient`, `ForwardDiff.hessian`, `ForwardDiff.derivative`, etc. are also not exported. It seems the API stability is not related to exported or not.

devmotion: To clarify, I think these functions should not be included in the documentation for the time being and only added if they are considered important enough and their docstrings are more useful.

hyrodium: Sorry, I don't quite understand. 

Documentation should be written for both package users and developers, and including Internal API documentation will be useful for future package developers. Of course, if these docstrings are incorrect or meaningless, they should be removed, but that is not the case.

There are more functions and types that need to be written (https://github.com/JuliaDiff/ForwardDiff.jl/issues/682), but I believe this is simply a matter of insufficient documentation, and not a reason to eliminate existing documentation.

hyrodium: While I am not the maintainer of the repository, I am open to following your guidance as the maintainer. If you believe that removing these documents is the best course of action, I am willing to proceed with that.

devmotion: AFAICT neither users nor developers should have to deal with `can_dual` and `≺`, so even though there is a docstring I don't think it should be included in the documentation (to me it seems these are just more detailed comments). `value`, `Dual`, etc. are more relevant IMO: #682 (but this should be discussed and addressed separately)

hyrodium: Okay, I'll update this PR to remove them from the documentation.

I will write down additional my concerns for future reference:

* If we use `checkdocs=:exports`, that will be as harmful as `warnonly=true` because this package provides non-exported stable API.
* I agree that the docstrings are something like detailed comments. But it would be very hard to discuss whether new docstrings should be in the docs or not every time. The easiest way would be to document the all docstrings.

docs/make.jl
```diff
@@ -1,18 +1,25 @@
...
+            "Advanced Usage Guide" => "user/advanced.md",
+            "Upgrading from Older Versions" => "user/upgrade.md"
+        ],
+        "Developer Documentation" => [
+            "How ForwardDiff Works" => "dev/how_it_works.md",
+            "How to Contribute" => "dev/contributing.md"
+        ]
+    ],
+    checkdocs=:exports,
+)
```
devmotion, line 21: I suggest keeping the changes minimal and following the current style:
```suggestion
makedocs(modules=[ForwardDiff],
         doctest = false,
         sitename = "ForwardDiff",
         pages = ["Introduction" => "index.md",
                  "User Documentation" => [
                    "Limitations of ForwardDiff" => "user/limitations.md",
                    "Differentiation API" => "user/api.md",
                    "Advanced Usage Guide" => "user/advanced.md",
                    "Upgrading from Older Versions" => "user/upgrade.md"],
                  "Developer Documentation" => [
                    "How ForwardDiff Works" => "dev/how_it_works.md",
                    "How to Contribute" => "dev/contributing.md"]],
         checkdocs = :exports)
```

docs/make.jl
```diff
@@ -1,18 +1,25 @@
...
 
 deploydocs(
-    repo = "github.com/JuliaDiff/ForwardDiff.jl.git"
+    repo="github.com/JuliaDiff/ForwardDiff.jl.git"
```
devmotion, line 24:
```suggestion
    repo = "github.com/JuliaDiff/ForwardDiff.jl.git"
```

devmotion (approved):

Issue #686, opened 2024-03-01
promote_rule ambiguity with AbstractIrrational and ForwardDiff.Dual


ElOceanografo:
Just ran into this error: 

```julia
ERROR: MethodError: promote_rule(::Type{IrrationalConstants.Twoπ}, ::Type{ForwardDiff.Dual{ForwardDiff.Tag{OptimizationReverseDiffExt.OptimizationReverseDiffTag, Float64}, Float64, 12}}) is ambiguous.

Candidates:
  promote_rule(::Type{S}, ::Type{T}) where {S<:AbstractIrrational, T<:Number}
    @ Base irrationals.jl:45
  promote_rule(::Type{<:AbstractIrrational}, ::Type{T}) where T<:Real
    @ Base irrationals.jl:44
  promote_rule(::Type{R}, ::Type{ForwardDiff.Dual{T, V, N}}) where {R<:Real, T, V, N}
    @ ForwardDiff C:\Users\...\.julia\packages\ForwardDiff\PcZ48\src\dual.jl:427

Possible fix, define
  promote_rule(::Type{R}, ::Type{ForwardDiff.Dual{T, V, N}}) where {R<:AbstractIrrational, T, V, N}
```
Changing `Irrational` to `AbstractIrrational` in [this line](https://github.com/JuliaDiff/ForwardDiff.jl/blob/5e9240ae63e5d4217e1dc17cc14deca1d74c1ad7/src/dual.jl#L437) resolves the problem by defining the missing method. Unless there is a reason not to broaden the typing here, I'll go ahead and open a PR...

devmotion:
I don't see a reason to not switch from `Irrational` to `AbstractIrrational`. Very likely there just didn't exist any other implementations of `AbstractIrrational` when this code was added.

devmotion:
Fixed by #687.

PR #687, opened 2024-03-03
Broaden promote_rule for Dual to include AbstractIrrational

ElOceanografo:
Fix for method ambiguity in https://github.com/JuliaDiff/ForwardDiff.jl/issues/686.

Couldn't tell if/how tests should be updated to reflect this change, happy to do that too if someone more knowledgeable can give me a pointer.


devmotion (approved):
LGTM, thank you!

PR #688, opened 2024-03-07
Restrict StaticArrays to compatible versions

andreasnoack:
It looks like version 1.7 broke the eigen definition here so until it has been fixed, the bounds would have to be restricted.


Project.toml
```diff
@@ -31,7 +31,7 @@ LogExpFunctions = "0.3"
 NaNMath = "1"
 Preferences = "1"
 SpecialFunctions = "1, 2"
-StaticArrays = "1.5"
+StaticArrays = "< 1.7"
```
devmotion, line 34: This will re-enable StaticArrays support for all versions < 1.5. I think it would be safer to only allow
```suggestion
StaticArrays = "1.5, 1.6"
```

devmotion (approved):
LGTM (I'm fine with fixing the remaining allocation tests in a separate PR).

andreasnoack:
@devmotion I've also disabled testing on nightly. There is a failure similarly to the one that was fixed in 1.10 by capping StaticArrays. I've also marked two of the allocation tests as broken.

devmotion (approved):

Issue #377, opened 2018-11-16
Problem with `abs`


briochemc:
Hi there,

I'm not sure this is intentional or if it is for numerical efficiency reasons, and I may be out of my league discussing this, but I figured I should point this out. When trying to copy the DualNumbers.jl structure into the HyperDualNumbers.jl structure, I noticed that there is a problem with how `abs` is dealt with. I believe the problem is that `abs(x)` returns `x` if `x >= 0` (in DualNumbers.jl, HyperDualNumbers.jl, and I believe in ForwardDiff.jl's dual implementation).

Here is a small example where this causes problems. I use `ForwardDiff.hessian` to evaluate the second derivative of `abs(-x^2)`, which should be the same as the second derivative of `x^2`, which is always `2`:
```julia
julia> using ForwardDiff

julia> ForwardDiff.hessian(x -> abs(-x[1]^2), [0.0])
1×1 Array{Float64,2}:
 -2.0
```
(I used `hessian` and a single element `Vector` because I did not know how to get to second derivatives of scalars using ForwardDiff.) 

I come with a suggestion: What about dealing with `abs` in the "physical" way, that is thinking of the absolute value of some `x` plus or minus some very small term `ε`. In other words, treating it recursively via something like:
```julia
function myabs(d::Dual)
    x, y = value(d), epsilon(d)
    if x > 0
        return d
    elseif x < 0
        return -d
    else
        if y ≥ 0
            return d
        else
            return -d
        end
    end
end
```

For hyper duals, something like
```julia
function myabs(h::Hyper)
    x, y, z, w = value(h), epsilon1(h), epsilon2(h), epsilon12(h)
    if x > 0
        return h
    elseif x < 0
        return -h
    else
        if y+z > 0
            return h
        elseif y+z < 0
            return -h
        else
            if z ≥ 0
                return h
            else z < 0
                return -h
            end
        end
    end
end
```
Now I know this is not DualNumbers.jl or HyperDualNumbers.jl, to which the above example suggestions apply, but I haven't really figured out how ForwardDiff.jl implements hyperduals so this was the best way I could convey my suggestions. I also believe that it is very likely that one can code the suggested `abs` in a much more efficient way than my `if` statements allow. But regardless of that, is that something worse paying some attention?

KristofferC:
```
julia> using ForwardDiff

julia> ForwardDiff.hessian(x -> abs(-x[1]^2), [0.0])
1×1 Matrix{Float64}:
[...]

PR #695, opened 2024-03-13
Make `isless`, `<`, `>`, `<=`, and `>=` consistent with `==` and `isequal`

devmotion:
Currently, the inequality operators for `Dual`s are inconsistent with the definitions of `==` and `isequal`. In particular, `isless` violates the first requirement listed in its docstring:
```markdown
If `isless(x, y)` is defined, then so is `isless(y, x)` and `isequal(x, y)`, and exactly one of those three yields true.
```

The PR fixes these inconsistencies.

### DIFF
--- a/src/dual.jl
+++ b/src/dual.jl
@@ -384,16 +384,30 @@ end
 # Before PR#481 this loop ran over this list:
 # BINARY_PREDICATES = Symbol[:isequal, :isless, :<, :>, :(==), :(!=), :(<=), :(>=)]
 # Not a minimal set, as Base defines some in terms of others.
-for pred in [:isless, :<, :>, :(<=), :(>=)]
+for pred in [:<, :>]
+    predeq = Symbol(pred, :(=))
     @eval begin
         @define_binary_dual_op(
             Base.$(pred),
-            $(pred)(value(x), value(y)),
-            $(pred)(value(x), y),
-            $(pred)(x, value(y)),
+            $(pred)(value(x), value(y)) || (value(x) == value(y) && $(pred)(partials(x), partials(y))),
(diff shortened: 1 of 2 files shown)


mcabbott:
To summarise what this does, tagged version is like so:
```julia

julia> pr(x, op, y) = println("x ", op, " y = ", op(x,y));

julia> pr.(Dual(1,1), (>,>=,==,<=,<), 1);
x > y = false
x >= y = true
x == y = true
x <= y = true
x < y = false

(@v1.12) pkg> st ForwardDiff
Status `~/.julia/environments/v1.12/Project.toml`
  [f6369f11] ForwardDiff v0.10.36
```
master has instead this:
```julia
julia> pr.(Dual(1,1), (>,>=,==,<=,<), 1);
x > y = false
x >= y = true
x == y = false
x <= y = true
x < y = false

(jl_I2PrM4) pkg> st ForwardDiff
Status `/private/var/folders/yq/4p2zwd614y59gszh7y9ypyhh0000gn/T/jl_I2PrM4/Project.toml`
  [f6369f11] ForwardDiff v0.11.0-DEV `~/.julia/dev/ForwardDiff`
```
With this PR,
```julia
julia> pr.(Dual(1,1), (>,>=,==,<=,<), 1);  # like pr.(1.01, (>,>=,==,<=,<), 1.0);
x > y = true
x >= y = true
x == y = false
x <= y = false
x < y = false

julia> pr.(Dual(1,-1), (>,>=,==,<=,<), 1);  # like pr.(-1.01, (>,>=,==,<=,<), 1.0);
x > y = false
x >= y = false
x == y = false
x <= y = true
x < y = true
```
A similar change was proposed in https://github.com/JuliaDiff/ForwardDiff.jl/issues/377, and also discussed in #609. Probably this PR closes both of those.

mcabbott (approved):

PR #697, opened 2024-03-25
fix: Improve formula presentation

junyixu:


andreasnoack:
Would be good to show the before and after in a comment

junyixu:
Before:
image


After:
Screenshot_20240325_193632

docs/src/dev/how_it_works.md
```diff
@@ -102,24 +102,28 @@ two calls to ``f`` to evaluate the gradient:
...
+```
+1st call:
+```math
+\to \vec{x}_{\epsilon} = \begin{bmatrix}
                         x_1 + \epsilon_1 \\
                         x_2 + \epsilon_2 \\
                         x_3 \\
                         x_4
                      \end{bmatrix}
 \to
```
mcabbott, line 114: Minor point, but can I suggest fixing the arrows?
```suggestion
\vec{x}_{\epsilon} = \begin{bmatrix}
                        x_1 + \epsilon_1 \\
                        x_2 + \epsilon_2 \\
                        x_3 \\
                        x_4
                     \end{bmatrix}
\mapsto
```
IMO it's not ideal to use `\to` with two different nonstandard meanings.

junyixu:
I'm not sure. Represent a function $f$ that maps an element $x$ in a set to another element $y$:
$$f: x \mapsto y.$$
But in the original doc page,
image
it seems the author intended to use `\to` to represent a type of change.

mcabbott:
Sorry just went by your pictures, where after used it for two kinds of change. x to x_eps, and x_eps to f(x_eps). It seems the original commits the same sin, could fix that too?

junyixu:
> Sorry just went by your pictures, where after used it for two kinds of change. x to x_eps, and x_eps to f(x_eps). It seems the original commits the same sin, could fix that too?

I force-pushed a new commit. What do you think?

Before:
image

After:
image

PR #703, opened 2024-05-28
Add support for StaticArrays >= 1.7

devmotion:
In StaticArrays 1.7 broadcasting with static arrays was improved: https://github.com/JuliaArrays/StaticArrays.jl/pull/1220 This fix implies that in some cases in newer versions now a static array is returned where in older versions incorrectly an `Array` was returned.

In ForwardDiff this breaks the `eigen` definition which uses an in-place method `_lyap_div!` to compute `A ./ (l' .- l)` (with diagonal entries `A[i, i]`). This PR fixes the problem by replacing it with a `_lyap_div!!` method that, following the bang-bang notation, falls back to an out-of-place implementation and only uses an in-place method for arrays that are known to be mutable such as `Matrix` or `StaticArrays.MMatrix`.

Fixes #690.
Fixes #702.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -32,7 +32,7 @@ LogExpFunctions = "0.3"
 NaNMath = "1"
 Preferences = "1"
 SpecialFunctions = "1, 2"
-StaticArrays = "1.5 - 1.6"
+StaticArrays = "1.5"
 julia = "1.6"
 
 [extras]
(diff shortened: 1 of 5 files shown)


mcabbott (approved):
Sounds good, haven't read every line, but am happy if tests pass.

Issue #716, opened 2024-11-01
Implementation of Dual for `NaNMath.pow` does not use `NaNMath.log`


jClugstor:
I would expect this: 
```julia
using ForwardDiff
using NaNMath

function new_pow(x)
    NaNMath.pow(x[1],x[2])
end

ForwardDiff.gradient(new_pow, [-1.0, 1.0])
```
to return a `NaN` I think? 

The issue is right here,
```julia
for f in (:(Base.:^), :(NaNMath.pow))
    @eval begin
        @define_binary_dual_op(
            $f,
            begin
                vx, vy = value(x), value(y)
                expv = ($f)(vx, vy)
                powval = vy * ($f)(vx, vy - 1)
                if isconstant(y)
                    logval = one(expv)
                elseif iszero(vx) && vy > 0
                    logval = zero(vx)
                else
                    logval = expv * log(vx)
                end
                new_partials = _mul_partials(partials(x), partials(y), powval, logval)
                return Dual{Txy}(expv, new_partials)
            end,
```

in `logval = expv*log(vx)` , I think `NaNMath.log` should be used if `f` is `NaNMath.pow`.

KristofferC:
Makes sense to me.

PR #717, opened 2024-11-01
Fix exponentiation for `NaNMath.pow`

jClugstor:
Currently an example like this:
```julia
using ForwardDiff
using NaNMath

function new_pow(x)
    NaNMath.pow(x[1],x[2])
end

ForwardDiff.gradient(new_pow, [-1.0, 2.0])
```
errors. I'd like to make it so that it returns a `NaN` instead. 

Fixes https://github.com/JuliaDiff/ForwardDiff.jl/issues/716

### DIFF
--- a/src/dual.jl
+++ b/src/dual.jl
@@ -552,7 +552,7 @@ end
 # exponentiation #
 #----------------#
 
-for (f, log) in ((:(Base.:^), :(NaNMath.pow)), (:(NaNMath.pow), :(NaNMath.log)))
+for (f, log) in ((:(Base.:^), :(Base.log)), (:(NaNMath.pow), :(NaNMath.log)))
     @eval begin
         @define_binary_dual_op(
             $f,
--- a/src/dual.jl
+++ b/src/dual.jl
@@ -552,43 +552,73 @@ end
 # exponentiation #
 #----------------#
 
-for f in (:(Base.:^), :(NaNMath.pow))
-    @eval begin
-        @define_binary_dual_op(
-            $f,
-            begin
-                vx, vy = value(x), value(y)
-                expv = ($f)(vx, vy)
-                powval = vy * ($f)(vx, vy - 1)
-                if isconstant(y)
-                    logval = one(expv)
-                elseif iszero(vx) && vy > 0
-                    logval = zero(vx)
-                else
-                    logval = expv * log(vx)
-                end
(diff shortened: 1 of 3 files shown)


mcabbott:
Can you add a few cases to the tests? Ideally NaN/not/negative through both paths.

src/dual.jl
```diff
@@ -552,43 +552,73 @@ end
 # exponentiation #
 #----------------#
 
-for f in (:(Base.:^), :(NaNMath.pow))
```
devmotion, line 555: Could we avoid the code duplication and just change this to
```julia
for (f, invf) in ((:(Base.:^), :(Base.log)), (:(NaNMath.pow), :(NaNMath.log)))
```
and use `$invf` instead of `log` below?

jClugstor: Yes

src/dual.jl
```diff
@@ -583,13 +583,16 @@ for f in (:(Base.:^), :(NaNMath.pow))
...
+                deriv = (iszero(x) && v > 0) ? zero(expv) : expv*($log)(x)
                 return Dual{Ty}(expv, deriv * partials(y))
             end
         )
     end
 end
 
+
+
+
```
devmotion, line 595:
```suggestion
```

test/GradientTest.jl
```diff
@@ -200,6 +201,18 @@ end
...
+    @test isnan(ForwardDiff.gradient(x -> NaNMath.pow(x[1],x[1]), [NaN, 1.0])[1])
+    @test ForwardDiff.gradient(x -> NaNMath.pow(x[1], x[2]), [1.0, 1.0]) == [1.0, 0.0]
+    @test isnan(ForwardDiff.gradient((x) -> NaNMath.pow(x[1], x[2]), [-1.0, 0.5])[1])
+
+    @test isnan(ForwardDiff.gradient(x -> x[1]^x[2], [NaN, 1.0])[1])
+    @test ForwardDiff.gradient(x -> x[1]^x[2], [1.0, 1.0]) == [1.0, 0.0]
+    @test_throws DomainError ForwardDiff.gradient(x -> x[1]^x[2], [-1.0, 0.5])
+
+end
+
```
devmotion, line 214:
```suggestion
end
```

devmotion (approved):

PR #722, opened 2024-11-15
Remove CI check for invalidations

devmotion:
The PR removes the invalidation check, since it has been failing due to a JET error for a while, is IMO typically ignored, doesn't flag increased numbers of invalidations correctly (https://github.com/julia-actions/julia-invalidations/pull/20) and might be misguided (https://github.com/julia-actions/julia-invalidations/issues/19).


charleskawczynski:
Can we instead change this to be a soft-fail? IMO, the invalidations table is still useful and provides actionable information. Testing on the single number of invalidations seems very impractical to me.

devmotion:
The action is broken currently: https://github.com/JuliaDiff/ForwardDiff.jl/actions/runs/11857687121/job/33046631013?pr=721 Of course, we could try to fix it, but due to the points mentioned above (no impact, incorrect reports, and arguably conceptually questionable) IMO it would be better to remove it.

KristofferC:
Yew, it's worthless.

KristofferC (approved):

Issue #726, opened 2024-12-04
Modification of equality checking between two Duals not synchronized between master and last release


clguillot:
Hi everyone,

It seems that the modifications required by issue #481, which should have been in the code starting from release 0.10.33 (according to the release notes) are indeed included in the branch master, but are not present in the last release 0.10.38.
In particular, the following
https://github.com/JuliaDiff/ForwardDiff.jl/blob/0a35a807e9a39a37d99d9928ea7d61247d8e2882/src/dual.jl#L392-L401
should have been replaced for the operator == by
https://github.com/JuliaDiff/ForwardDiff.jl/blob/3acc8a64088564deec85103ed2dcc7cf6c0c2406/src/dual.jl#L391-L407
As a consequence of this modification, equality between 2 Dual with the same tags should check both the equality of the value and the partials, hence avoiding some branching that would have been taken otherwise.
In particular, without this fix, we obtain
```julia
julia> f(x) = exp(complex(0, x))
f (generic function with 1 method)

julia> df(x) = ForwardDiff.derivative(f, x)
df (generic function with 1 method)

julia> d2f(x) = ForwardDiff.derivative(df, x)
d2f (generic function with 1 method)

julia> d2f(0.0)
0.0 + 0.0im
```
which is obviously wrong, but
```julia
julia> d2f(1e-15)
-1.0 - 1.0e-15im
```
which is right.
This can be explained by looking at the code of exp in complex.jl
```julia
function exp(z::Complex)
    zr, zi = reim(z)
    if isnan(zr)
        Complex(zr, zi==0 ? zi : zr)
    elseif !isfinite(zi)
        if zr == Inf
            Complex(-zr, oftype(zr,NaN))
        elseif zr == -Inf
            Complex(-zero(zr), copysign(zero(zi), zi))
        else
            Complex(oftype(zr,NaN), oftype(zi,NaN))
        end
    else
        er = exp(zr)
        if iszero(zi)
            Complex(er, zi)
        else
            s, c = sincos(zi)
            Complex(er * c, er * s)
        end
    end
end
```
Here, the branching if iszero(zi) is taken because iszero will only check that value(zi) is zero when it should also check that partials(zi) is zero (which won't be the case in this example).

Is there a reason why the modification that should have been included from release 0.10.33 onward can be found in the master branch but not in the last release ?

mcabbott:
> should have been in the code starting from release 0.10.33 (according to the release notes) 

[...]

Issue #653, opened 2023-06-09
incorrect 2nd derivative of complex exponential


stevengj:
This, [from discourse](https://discourse.julialang.org/t/bug-in-forward-diff-for-complex-valued-function/100119), looks like a bug:
```jl
julia> p(s) = exp(1im*(2π*s - π/2));

julia> p′(s) = ForwardDiff.derivative(p, s);

julia> p″(s) = ForwardDiff.derivative(p′, s);

julia> p″(0.2499999999999)
-39.47841760435743 + 2.4807694081270598e-11im

julia> p″(0.25)
0.0 + 0.0im
```
I don't think mere floating-point error can explain why `p″(0.25)` is suddenly zero — looks like a misplaced conditional or something.  Note that the correct answer for `p″(0.25)` is exactly `-(2π)² ≈ -39.47841760435743`.

(The first derivative `p′` looks okay.)

mikmoore:
Reduced example:
```
julia> p(s) = exp(1im*s);

julia> p′(s) = ForwardDiff.derivative(p, s);

julia> p″(s) = ForwardDiff.derivative(p′, s);

julia> p″(eps(Float64))
-1.0 - 2.220446049250313e-16im

julia> p″(0.0)
0.0 + 0.0im
```

mcabbott:
I believe this is one more case fixed by #481:
```
julia> p″(0.2499999999999)
-39.47841760435743 + 2.4807694081270598e-11im

julia> p″(0.25)
-39.47841760435743 + 0.0im

(jl_aRIFuW) pkg> st ForwardDiff
Status `/private/var/folders/yq/4p2zwd614y59gszh7y9ypyhh0000gn/T/jl_aRIFuW/Project.toml`
  [f6369f11] ForwardDiff v0.11.0-DEV `https://github.com/JuliaDiff/ForwardDiff.jl.git#master`
```

mcabbott:
Fixed on 1.0

Issue #514, opened 2021-04-06
Complex derivative of power fails


jagot:
I'm trying to calculate complex derivatives of analytic functions, using [a trick](https://discourse.julialang.org/t/automatic-differentiation-of-complex-valued-functions/30263/3) by @tpapp:

```julia
using ForwardDiff

∂(f, x::Real) = ForwardDiff.derivative(f, x)
function ∂(f, z::Complex)
    # https://discourse.julialang.org/t/automatic-differentiation-of-complex-valued-functions/30263/3
    ff = ((x,y),) -> begin
        fz = f(complex(x,y))
        [real(fz), imag(fz)]
    end
    J = ForwardDiff.jacobian(ff, [real(z), imag(z)])
    # Complex derivative for an analytic function
    #   f(x + im*y) = u(x,y) + im*v(x,y)
    # is given by
    #   f′(x + im*y) = uₓ + im*vₓ = v_y - im*u_y
    J[1,1] + im*J[2,1]
end
```

This works for some simple functions:
```julia
julia> ∂(sin, 0.1), ∂(sin, 0.1+0im), ∂(sin, 0.1+0.3im)
(0.9950041652780258, 0.9950041652780258 + 0.0im, 1.0401161756837587 - 0.030401301333122962im)

julia> ∂(t -> exp(-t), 0.1+0.3im)
-0.8644242021759518 + 0.26739774077289963im
```
But fails for functions involving powers:
```julia
julia> f = z -> z^2
#4019 (generic function with 1 method)

julia> ∂(f, 0.1)
0.2

julia> ∂(f, 0.1+0.0im)
ERROR: MethodError: no method matching Int64(::ForwardDiff.Dual{ForwardDiff.Tag{var"#4015#4016"{var"#4019#4020"}, Float64}, Float64, 2})
Closest candidates are:
  (::Type{T})(::T) where T<:Number at boot.jl:760
  (::Type{T})(::AbstractChar) where T<:Union{AbstractChar, Number} at char.jl:50
  (::Type{T})(::BigInt) where T<:Union{Int128, Int16, Int32, Int64, Int8} at gmp.jl:356
  ...
Stacktrace:
  [1] convert(#unused#::Type{Int64}, x::ForwardDiff.Dual{ForwardDiff.Tag{var"#4015#4016"{var"#4019#4020"}, Float64}, Float64, 2})
    @ Base ./number.jl:7
[...]

PR #727, opened 2024-12-04
Explicit derivatives for complex analytic functions

clguillot:
Hi everyone,

As documented in #726, #514, #653  ForwardDiff encounters some difficulty when it comes to differentiating complex functions. The issue is similar to the one in #481 . This is a very serious issue which has been encountered by several people, including myself, and can be quite difficult to identify the first time, especially when using ForwardDiff as a "black box". Julia and ForwardDiff are widely used by mathematicians in the field of quantum chemistry, who manipulate complex number daily, and encountering this issue can result in a huge loss of time.
Up to a few exceptions (mainly the functions abs and abs2), all the rules defined in DiffRules can be extended to the complex domain with no modification. I worked a bit on the issue and came up with a start of fix by modifying unary_dual_definition as follows:
https://github.com/clguillot/ForwardDiff.jl/blob/114cfe90755df8591488c7d71bd3109be5325fb9/src/dual.jl#L234-L265
I simply define a version of the derivative for Complex{Dual} defined with the expression found in DiffRule and returning a new Complex{Dual}.
With this fix, I get the correct result when computing the order 2 derivative of exp(ix):
```julia
julia> f(x) = exp(1im*x);
julia> df(x) = ForwardDiff.derivative(f, x);
julia> ForwardDiff.derivative(df, 0.0)
-1.0 + 0.0im
```
Without the fix, the same computation (unless the modification in #481 is implemented) returns 0.0 + 0.0im, which is obviously wrong.
I also implemented sin and cos for Complex by hand
https://github.com/clguillot/ForwardDiff.jl/blob/114cfe90755df8591488c7d71bd3109be5325fb9/src/dual.jl#L747-L771
but avoiding sincos since I was lazy.

I believe having explicit derivatives in those cases will mostly free ForwardDiff from having to worry about how the functions are implemented in the libraries, since it never needs to actually go through this code with a Dual type. Moreover, I don't think it to be too harmful for the performances either.

One issue that I see with this pull request is the manual exclusion of some function. It would probably be more elegant to modify DiffRules to indicate which function can see its derivative extended to the complex domain, for example by defining a macro @define_analytic_diffrule which would make a call to @define_diffrules and put the function into a some kind of list indicating that it is analytic. Until something like this can be pulled up, the code above should at least provide a basis that returns the right answer in most cases.
It would also be nice to provide a similar fix for functions of several variables, but all in good time.

### DIFF
--- a/src/dual.jl
+++ b/src/dual.jl
@@ -726,6 +726,22 @@ end
     Dual{Tz}(muladd(x, y, value(z)), partials(z))      # z_body
 )
 
+# inv(Complex) #
+#--------#
+
+function Base.inv(d::Complex{<:Dual{T}}) where{T}
+    FD = ForwardDiff
+    x = complex(FD.value(real(d)), FD.value(imag(d)))
+    val = inv(x)
+    deriv = - val * val
+    re_deriv, im_deriv = reim(deriv)
+    re_partials = FD.partials(real(d))
+    im_partials = FD.partials(imag(d))
+    re_retval = FD.dual_definition_retval(Val{T}(), real(val), re_deriv, re_partials, -im_deriv, im_partials)
+    im_retval = FD.dual_definition_retval(Val{T}(), imag(val), im_deriv, re_partials, re_deriv, im_partials)
+    return complex(re_retval, im_retval)
+end
+
 # sin/cos #
 #--------#
 
--- a/src/dual.jl
+++ b/src/dual.jl
@@ -245,11 +245,11 @@ function unary_dual_definition(M, f)
             return $FD.dual_definition_retval(Val{T}(), val, deriv, $FD.partials(d))
         end
     end
-    if (M, f) in ((:Base, :abs), (:Base, :abs2))
+    if (M, f) in ((:Base, :abs), (:Base, :abs2), (:Base, :inv))
         return real_diff_exp
     else
         complex_diff_expr = quote
-            @inline function $M.$f(d::Complex{$FD.Dual{T, V, N}}) where{T, V, N}
+            @inline function $M.$f(d::Complex{<:$FD.Dual{T}}) where{T}
                 x = complex($FD.value(real(d)), $FD.value(imag(d)))
                 $work
                 re_deriv, im_deriv = reim(deriv)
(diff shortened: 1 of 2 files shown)


mcabbott:
Am I correct to think that #514 and #653 are fixed on master, as is #486 (similar but not mentioned above)?

And are there other cases where Complex{Dual} goes wrong, which aren't fixed?

Certainly the tests in this PR pass:
```julia
julia> @testset "analytic functions" begin
           dexp(x) = ForwardDiff.derivative(y -> exp(complex(0, y)), x)
           @test ForwardDiff.derivative(dexp, 0.0) ≈ -1
           @test ForwardDiff.derivative(x -> exp(1im*x), 0.7) ≈ im * cis(0.7)
           @test ForwardDiff.derivative(x -> sqrt(im + (1+im) * x), 1.23) ≈ (1+im) / (2 * sqrt(im + (1+im)*1.23))
       end
Test Summary:      | Pass  Total  Time
analytic functions |    3      3  0.5s
Test.DefaultTestSet("analytic functions", Any[], 3, false, false, true, 1.733352822760275e9, 1.733352823277592e9, false, "REPL[29]")
```

That's not to say we shouldn't add complex methods, but if they don't have correctness justifications, then they need something else. Are these more accurate, or faster, or do they handle some edge cases better?

clguillot:
The problem is effectively fixed on master, but not in the release. From a user point of view, it is equivalent to an unresolved issue.

I see several benefits from handling the complex derivatives this way:

- The question of having or not ForwardDiff compatible code inside complex elementary functions would no longer be a preoccupation as long as a correct diffrule is defined.
- This would fix the issue in a way that seems less breaking than #481 , since I don't see why someone would rely on the fact that we actually go through the code of elementary functions with a Dual type.
- Regarding the performance, that's a good points, it would probably require some careful testing. I am not expecting too much improvement on this side anyway.
- Another benefit worth mentioning concerns the support of functions defined in SpecialFunctions. Since a lot of those are defined only for ComplexF64, the current version of ForwardDiff will throw an error when calling it with Complex{Dual}. Using the diffrule instead, we get the correct answer.
```julia
julia> (1+im) * digamma((1+im)*1.2)
-0.7029212687502131 + 1.3421243356860844im
julia> ForwardDiff.derivative(x -> loggamma((1+im)*x), 1.2) # With the diffrule
-0.7029212687502131 + 1.3421243356860844im
julia> ForwardDiff.derivative(x -> loggamma((1+im)*x), 1.2) # Without the diffrule
ERROR: MethodError: no method matching _loggamma(::Complex{ForwardDiff.Dual{ForwardDiff.Tag{var"#9#10", Float64}, Float64, 1}})
The function `_loggamma` exists, but no method is defined for this combination of argument types.
```
In any case, if this modification does not seem appropriate, a corrected version of exp(Complex{Dual}) should be defined as a temporary fix in the release, until the discussions about v1.0 are concluded. If you prefer this way I can make a pull request in this sense, and maybe leave the question of complex derivatives for another time.

PR #728, opened 2024-12-05
v1.0

dlfivefifty:
I also updated the CI to test on lts and bumped the julia version

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1.6'
+          - 'lts'
           - '1'
           # - 'nightly'
         os:
--- a/Project.toml
+++ b/Project.toml
@@ -13,7 +13,6 @@ Preferences = "21216c6a-2e73-6563-6e65-726566657250"
 Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 SpecialFunctions = "276daf66-3868-5448-9aa4-cd146d93841b"
-StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 
 [weakdeps]
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1.6'
+          - 'lts'
```
devmotion, line 20:
```suggestion
          - 'min'
          - 'lts'
```
(They are not identical)

Project.toml
```diff
@@ -33,7 +33,7 @@ NaNMath = "1"
 Preferences = "1"
 SpecialFunctions = "1, 2"
 StaticArrays = "1.5"
-julia = "1.6"
+julia = "1.10"
```
devmotion, line 36: Why?

dlfivefifty: No, but if we aren't testing on Julia v1.6 it's best to explicitly drop support. It also allows us to remove weakdeps as deps and use any features from v1.10+.

.github/workflows/ci.yml
dlfivefifty: What's min?

devmotion: It tests the minimum compatible version.

Project.toml
devmotion: My point is that the package is tested successfully on 1.6, it's that the PR in its current form dropped tests on 1.6 (but adding `min` would fix that). Releasing 1.0 doesn't require making any changes to the package or its dependencies.

devmotion: To be clear, I'm not generally opposed to making such changes but I think they should be done separately. Releasing 1.0 should just involve updating the version number + removing deprecations (if existent) IMO. Otherwise it's also easy to forget to remove VERSION-specific code, old workarounds, etc.

dlfivefifty: I don't really care, push any changes you want to this PR

devmotion: No, I don't want to. I want to keep them in separate PRs. And I don't think any of the changes in this PR apart from the 1.0 bump are necessary for releasing 1.0.

dlfivefifty: Then make your own PR

devmotion: https://github.com/JuliaDiff/ForwardDiff.jl/pull/729

devmotion: https://github.com/JuliaDiff/ForwardDiff.jl/pull/730

PR #730, opened 2024-12-06
Update CI

devmotion:
The PR updates the CI setup a bit:
- Tests are run with `min`, `lts`, and `1` Julia versions
- Codecov action is updated to v5 (fixes #723)
- The julia-actions/cache action is used instead of manual caching

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,5 @@
 name: CI
+
 on:
   pull_request:
     branches:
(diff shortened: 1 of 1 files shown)


devmotion:
Test failures were caused by a bug in NaNMath which was fixed in https://github.com/JuliaMath/NaNMath.jl/pull/92.

devmotion:
@gdalle Thanks for merging but I generally prefer the ColPrac standard

> PRs by people with merge rights should not be merged by people other than the author (just approved).

gdalle:
I was aware of it but there is no ColPack badge on this repo, should we add it?

devmotion:
The repo doesn't follow ColPrac officially, and I don't think it should be enforced. As I said, it's my general preference but it might not be shared by other contributors.

gdalle:
Judging by #749 it seems you're right! I'll keep it it mind for your PRs then

PR #731, opened 2024-12-07
Require Julia v1.10+

dlfivefifty:
I think with v1.0 it makes sense to drop the support for earlier versions of Julia before the LTS. The benefits are:

1. We can drop StaticArrays as a dependency
2. Makes it less restricted in terms of testing compatibility and adding features
3. Removal of a small amount of special cases in the tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1.6'
+          - 'lts'
           - '1'
           # - 'nightly'
         os:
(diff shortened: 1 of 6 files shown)


devmotion:
> We can drop StaticArrays as a dependency

StaticArrays is only a regular dependency on Julia < 1.9, on Julia >= 1.9 it is only a weak dependency.

mcabbott:
Maybe it's ideal to do this at 1.1? Then in case anyone still wants 1.6, there will be a 1.0 release they can install. I wouldn't put much work into supporting 1.6 now, but delaying slightly is easy.

dlfivefifty:
I can’t imagine anyone cares either way. Any Julia v1.6 users can still use ForwardDiff v0.x, if they are happy with old versions of Julia they are surely happy with old versions of ForwardDiff

the only point is to maintain developer sanity. Why support old versions for fictional users that probably don’t exist? Who is on Julia v1,6 but is upset about being on ForwardDiff v0.8??? Literally no one! 

So why not just drop support and we can all move on with our lives.

mcabbott:
My fictitious user is someone trying to update bounds so that their CI works! Not creating tighter coupling than we have to seems polite, when it's really easy. Agree real end-use isn't the issue.

dlfivefifty:
Yes I agree that’s a serious problem. Often trying to use outdated code fails even with manifests because it can figure out the right versionszzz

But I think that’s a Julia Pkg problem…

PR #732, opened 2024-12-12
Implement partials/value for complex duals

dlfivefifty:
At the moment values/partials give useless (and probably "wrong") answers for complex-duals:

```julia
julia> ForwardDiff.value(ForwardDiff.Dual(1,2) + im*ForwardDiff.Dual(3,4))
Dual{Nothing}(1,2) + Dual{Nothing}(3,4)*im


julia> ForwardDiff.value(ForwardDiff.Dual(1,2) + im*ForwardDiff.Dual(3,4))
Dual{Nothing}(1,2) + Dual{Nothing}(3,4)*im

julia>  ForwardDiff.partials(ForwardDiff.Dual(1,2) + im*ForwardDiff.Dual(3,4))
0-element ForwardDiff.Partials{0, Complex{ForwardDiff.Dual{Nothing, Int64, 1}}}

julia>  ForwardDiff.partials(ForwardDiff.Dual(1,2) + im*ForwardDiff.Dual(3,4),1)
Dual{Nothing}(0,0) + Dual{Nothing}(0,0)*im
```

This PR implements correct versions of these functions.

### DIFF
--- a/src/dual.jl
+++ b/src/dual.jl
@@ -88,6 +88,7 @@ Dual{T,V,N}(x::Base.TwicePrecision) where {T,V,N} =
 
 @inline value(x) = x
 @inline value(d::Dual) = d.value
+@inline value(d::Complex{<:Dual}) = complex(value(real(d)), value(imag(d)))
 
 @inline value(::Type{T}, x) where T = x
 @inline value(::Type{T}, d::Dual{T}) where T = value(d)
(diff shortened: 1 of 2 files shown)


mcabbott:
Are these the desired return types?
```julia
julia> ForwardDiff.partials(ForwardDiff.Dual(1,2) + im*ForwardDiff.Dual(3,4))
1-element Vector{Complex{Int64}}:
 2 + 4im

julia> ForwardDiff.partials(1 + im)
0-element ForwardDiff.Partials{0, Complex{Int64}}
```

dlfivefifty:
Ah yes you are right, it is better to return a `Partials`. It's fixed now

mcabbott:
I don't swear this is better, it just needs some thought. Are there any paths by which `Partials{3, ComplexF64}` can result in a Dual{Complex} instead of Complex{Dual}? That would be bad.

dlfivefifty:
Given that nothing is currently calling `partials(::Complex)` I don't see how this change can result in that happening, but yes there is a question about what `Partials` can represent.

PR #743, opened 2025-04-07
Fix seeding with undefined elements

### LINKED ISSUE #436: Jacobian of in-place functions in presence of undefined references
The method
```
ForwardDiff.jacobian(f!, y::AbstractArray, x::AbstractArray)
```
throws `UndefRefError` when `y` contains undefined references.
For example,
```
using ForwardDiff
f!(y, x) = copy!(y, x)
x0 = BigFloat[1.0, 1.0]
ForwardDiff.jacobian(f!, similar(x0), x0)
```
In particular, this affects the `NLsolve` package, e.g.,
```
using NLsolve
nlsolve(f!, x0, autodiff=:forward)
```
Since I think the reasonable expectation is that only the type and shape of the `y` array are going to be used in `ForwardDiff.jacobian`, I am reporting this here. If you do not think that the undefined case should be taken in account, I will post an issue for `NLsolve` instead.

### LINKED ISSUE #740: Why is seeding `ydual` necessary?
There are some lines in the code that I don't understand, essentially all the ones that show up here:
https://github.com/search?q=repo%3AJuliaDiff%2FForwardDiff.jl%20seed!(y&type=code

Why would we need to set the dual numbers inside `ydual` to a specific value, before overwriting them with `f!(ydual, xdual)`? Can we just get rid of these steps?

This is causing issues when `y` has non-bitstype elements, because they may not be initialized the first time that `jacobian` is called.

Related:

- https://github.com/JuliaNLSolvers/NLSolversBase.jl/pull/160
- https://github.com/JuliaNLSolvers/LsqFit.jl/pull/263

devmotion:
Fixes #436. Fixes #740.


gdalle:
I didn't know about `isassigned`, this seems appropriate. Could this cause a slowdown in the vast majority of cases where `similar` actually assigns stuff?

test/JacobianTest.jl
```diff
@@ -279,4 +279,18 @@ end
     end
 end
 
+@testset "BigFloat" begin
```
gdalle, line 282: Does this test cover all the new branches? It seems CodeCov is sleeping at the wheel.
I think we may also want to add cases where only some of the values are uninitialized?

src/apiutils.jl
```diff
@@ -72,16 +72,36 @@ end
...
+    if isbitstype(V)
+        for idx in structural_eachindex(duals, x)
+            duals[idx] = Dual{T,V,N}(x[idx], seed)
+        end
+    else
+        for idx in structural_eachindex(duals, x)
+            if isassigned(x, idx)
+                duals[idx] = Dual{T,V,N}(x[idx], seed)
+            else
+                Base._unsetindex!(duals, idx)
```
KristofferC, line 84: :/, this is quite unfortunate, is this really the only way to fix this?

gdalle: I think the idea is that if `x[idx]` is `#undef`, we want `duals[idx]` to be the same

devmotion: Yeah, I wasn't 100% percent happy but it was the best I could think of and what apparently is used in base. IMO we really want to ensure here that we get the same behaviour as `copyto!` on the primal part, to avoid that `yduals` contains any surprising values, possibly from previous chunks or differentations.

devmotion:
```julia
julia> x = BigFloat[1, 2];

julia> y = similar(x); y[2] = 1;

julia> x
2-element Vector{BigFloat}:
 1.0
 2.0

julia> y
2-element Vector{BigFloat}:
 #undef
   1.0

julia> copyto!(x, y);

julia> x
2-element Vector{BigFloat}:
 #undef
   1.0
```

gdalle:
What's the word on this one? Do we want it?

pkofod:
gentle bump from me as well

pkofod:
Bump again

test/JacobianTest.jl
devmotion: Codecov confirms that all modified lines are covered by the tests: https://app.codecov.io/github/JuliaDiff/ForwardDiff.jl/commit/ec3e483747159b46e0db5328c60c3543c13478a6/indirect-changes

KristofferC:
I opened https://github.com/JuliaLang/julia/issues/59394 regarding `unsetindex!`.

KristofferC (approved):

KristofferC (approved):

PR #744, opened 2025-04-07
Add `public` declarations

mcabbott:
First commit adds only those in `api.md`, but perhaps we should add more... like `Dual`? 

```julia
help?> ForwardDiff.Dual
  │ Warning
  │
  │  The following bindings may be internal; they may change or be
  │  removed in future versions:
  │
  │    •  ForwardDiff.Dual
```

(`eval` is following this https://github.com/JuliaLang/julia/pull/55097/files )

### DIFF
--- a/src/ForwardDiff.jl
+++ b/src/ForwardDiff.jl
@@ -22,6 +22,13 @@ include("gradient.jl")
 include("jacobian.jl")
 include("hessian.jl")
 
+if VERSION >= v"1.11.0-DEV.469"
+    # This list matches /docs/src/user/api.md
+    eval(Meta.parse(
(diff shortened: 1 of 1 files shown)


gdalle:
This is a very welcome initiative to guarantee API stability! So many people have been using ForwardDiff internals that they're not really internals anymore at this point.
For instance, here is the list of names that DI uses:

https://github.com/JuliaDiff/DifferentiationInterface.jl/blob/ba8a2fa9e7558e65655aeedbc36158e41ea5e324/DifferentiationInterface/ext/DifferentiationInterfaceForwardDiffExt/DifferentiationInterfaceForwardDiffExt.jl#L7-L28

gdalle:
Gentle bump on this one

gdalle (changes requested):

PR #746, opened 2025-04-11
Add support for floatmax(value) and friends

mancellin:
Fix https://github.com/JuliaMath/SpecialFunctions.jl/issues/475

The SpecialFunction `expint` is using `floatmax` on a value and thus the following currently fails:
```
ForwardDiff.derivative(x -> expint(im*x), 1.0)
```

I believe it makes sense to support both `floatmax(::T)` and `floatmax(::Type{T})` here since Base implements both.

### DIFF
--- a/src/dual.jl
+++ b/src/dual.jl
@@ -830,6 +830,9 @@ function Base.show(io::IO, d::Dual{T,V,N}) where {T,V,N}
 end
 
 for op in (:(Base.typemin), :(Base.typemax), :(Base.floatmin), :(Base.floatmax))
+    @eval function $op(::ForwardDiff.Dual{T,V,N}) where {T,V,N}
+        ForwardDiff.Dual{T,V,N}($op(V))
+    end
     @eval function $op(::Type{ForwardDiff.Dual{T,V,N}}) where {T,V,N}
         ForwardDiff.Dual{T,V,N}($op(V))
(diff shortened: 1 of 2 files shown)


devmotion (commented):
I just checked, and the docstring of `typemin`, `typemax`, `floatmin`, and `floatmax` only documents that these function can be called with types. A non-`Type` argument does not seem to be documented, so maybe it's not intended to be used with values? For comparison, the docstring for `eps` documents both the version with values and types.

The Julia docs even discourage the definition of functions for both value and type arguments: https://docs.julialang.org/en/v1/manual/style-guide/#Avoid-confusion-about-whether-something-is-an-instance-or-a-type

mancellin:
> Given the documentation of these functions, I'd think the better solution would be to change SpecialFunctions

I'm fine with both approaches.

@ararslan suggested in https://github.com/JuliaMath/SpecialFunctions.jl/issues/475 to implement it in ForwardDiff, since other might be using the undocumented behavior from Base.

devmotion:
My feeling is that it would be preferable if it would be documented that this function is expected to work on values before adding these definitions in ForwardDiff.

devmotion:
Closed in favour of https://github.com/JuliaMath/SpecialFunctions.jl/pull/494.

PR #749, opened 2025-05-12
Add ColPrac badge

gdalle:
@devmotion sorry for the confusion, this should help prevent unwanted merges in the future


KristofferC:
I don't think this is useful. AFAIU this has not ever been "followed" in this repo.

gdalle:
Fair enough

Issue #303, opened 2018-02-02
Incorrect computation of Hessian of squared euclidean norm


jomi2801:
The computation of the Hessian of the squared euclidean norm results in an interesting output:
```
using ForwardDiff;
f(x::Vector) = .5*norm(x, 2)^2;
x = zeros(2);
ForwardDiff.hessian(f, x)
```
which is
```2×2 Array{Float64,2}:
 0.0  0.0
 0.0  1.0
```

However, the codes
```
f(x::Vector) = .5*x[1]^2 + .5*x[2]^2;
ForwardDiff.hessian(f, x)
```
and
```
f(x::Vector) = .5*dot(x, x);
ForwardDiff.hessian(f, x)
```
provide the correct result. Why is that?

jrevels:
Possibly similar to https://github.com/JuliaDiff/ForwardDiff.jl/issues/197 i.e. Base's linear algebra code might be optimized in an "AD unsafe" way.

JoshuaLampert:
Is there any chance this gets fixed? Unfortunately, in my case it is not really an option to replace the `norm(x)^2` by `sum(x.^2)` since I have a function calling `norm(x)` and later another function squaring the result and because the concatenation gives `NaN`:
```julia
julia> using ForwardDiff

julia> f(x) = 0.5*sqrt(sum(x.^2))^2
f (generic function with 1 method)

julia> ForwardDiff.hessian(f, [0.0, 0.0])
2×2 Matrix{Float64}:
 NaN  NaN
 NaN  NaN
```
So this would require to rewrite my code base, which I would really like to avoid.

JoshuaLampert:
This issue happened to me again when trying to upgrade to ForwardDiff.jl v1. Even worse: Upgrading to ForwardDiff.jl@1, this also concerns `gradient`.
On ForwardDiff.jl v1.0.1:
```julia
julia> using ForwardDiff, LinearAlgebra

julia> f(x) = 0.5 * norm(x)^2
f (generic function with 1 method)

julia> ForwardDiff.hessian(f, [0.0, 0.0])
2×2 Matrix{Float64}:
 NaN  NaN
 NaN  NaN

julia> ForwardDiff.gradient(f, [0.0, 0.0])
2-element Vector{Float64}:
 NaN
 NaN
```
On ForwardDiff.jl v0.10.38:
```julia
julia> using ForwardDiff, using LinearAlgebra

[...]

PR #753, opened 2025-06-17
Add section in docs about limitation for functions composed of non-differentiable functions

JoshuaLampert:
As discussed in #303, this PR adds a new bullet point in the limitations section explaining possible issues, when a (differentiable) function is composed of non-differentiable functions.


andreasnoack:
The test failure might go away with https://github.com/JuliaDiff/ForwardDiff.jl/pull/754

PR #757, opened 2025-07-24
Fix implementations of `eigen` and `eigvals`

### LINKED ISSUE #756: Regression in 1.0, asking for Hessian of a complex matrix function results in infinite loop
I've encountered this issue when running the tests of my package [ConicQKD](https://github.com/araujoms/ConicQKD.jl), which started to hang when I switched from ForwardDiff 0.10 to 1.0.

It was a lot of work to reduce it to a minimum working example. Apparently the problem appears when you try to compute the Hessian of a function of a complex matrix that sets several elements to zero and asks for its eigenvalues. If I just ask for the eigenvalues without setting anything to zero then it's fine. Real matrices work without any problems, even when setting elements to zero.

To reproduce, call `test_hessian(ComplexF64)`. You can also check the real case `test_hessian(Float64)`, it works fine.

```julia
using LinearAlgebra
import ForwardDiff
import GenericLinearAlgebra.eigen
import Random

function test_hessian(::Type{T}) where {T}
    Random.seed!(1)

    d = 3

    function barrier(point)
        d2 = div(d, 2)
        M = mat(point, T)
        M[1:d2, d2+1:end] .= 0
        return sum(eigvals(Hermitian(M)))
    end

    x = randn(T, d, d)
    point = rvec(x * x')
    fd_hess = ForwardDiff.hessian(barrier, point)

    return fd_hess
end

function rvec(M::AbstractMatrix{T}) where {T}
    if T <: Real
        return vec(M)
    else
        return [real(vec(M)); imag(vec(M))]
    end
end

function mat(v::AbstractVector, ::Type{T}) where {T}
    if T <: Real
        n = isqrt(length(v))
        return reshape(v, n, n)
    else
        l2 = div(length(v), 2)
        n = isqrt(l2)
        return reshape(v[1:l2] + im*v[l2+1:end], n, n)
    end
end        
```

devmotion:
The PR adds missing definitions of `eigvals` and `eigen` for `Hermitian{<:Dual}` and of `eigen` for `Hermitian{<:Complex{<:Dual}}`.
Additionally, the PR adds more tests of `eigen` and `eigvals`, unifies the existing implementations, and removes duplicate calculations in the definitions of `eigen`.

Fixes #756 without GenericLinearAlgebra.

---

For the example in #756, I get on ForwardDiff@0.10 without GenericLinearAlgebra:

```julia
julia> test_hessian(Float64)
ERROR: MethodError: no method matching eigvals!(::Hermitian{ForwardDiff.Dual{…}, Matrix{…}}; sortby::Nothing)
The function `eigvals!` exists, but no method is defined for this combination of argument types.
...

julia> test_hessian(ComplexF64)
ERROR: MethodError: no method matching eigen!(::Hermitian{Complex{ForwardDiff.Dual{…}}, Matrix{Complex{…}}}; sortby::Nothing)
The function `eigen!` exists, but no method is defined for this combination of argument types.
...
```

And on ForwardDiff@0.10 with `import GenericLinearAlgebra: eigen`:

```julia
julia> @time test_hessian(Float64)
  0.000071 seconds (38 allocations: 33.750 KiB)
9×9 Matrix{Float64}:
 0.0  0.0  0.0  0.0  0.0          0.0  0.0  0.0          0.0
 0.0  0.0  0.0  0.0  0.0          0.0  0.0  0.0          0.0
 0.0  0.0  0.0  0.0  0.0          0.0  0.0  0.0          0.0
 0.0  0.0  0.0  0.0  0.0          0.0  0.0  0.0          0.0
 0.0  0.0  0.0  0.0  0.0          0.0  0.0  4.44089e-16  0.0
 0.0  0.0  0.0  0.0  0.0          0.0  0.0  0.0          0.0
 0.0  0.0  0.0  0.0  0.0          0.0  0.0  0.0          0.0
 0.0  0.0  0.0  0.0  4.44089e-16  0.0  0.0  2.66454e-15  0.0
 0.0  0.0  0.0  0.0  0.0          0.0  0.0  0.0          0.0

julia> @time test_hessian(ComplexF64)
  0.000556 seconds (1.97 k allocations: 1.387 MiB)
18×18 Matrix{Float64}:
 0.0  0.0  0.0  0.0   0.0          0.0  …  0.0  0.0  0.0   0.0          0.0
 0.0  0.0  0.0  0.0   0.0          0.0     0.0  0.0  0.0   0.0          0.0
 0.0  0.0  0.0  0.0   0.0          0.0     0.0  0.0  0.0   0.0          0.0
 0.0  0.0  0.0  0.0   0.0          0.0     0.0  0.0  0.0   0.0          0.0
 0.0  0.0  0.0  0.0  -6.93889e-18  0.0     0.0  0.0  0.0  -1.38778e-16  0.0
 0.0  0.0  0.0  0.0   0.0          0.0  …  0.0  0.0  0.0   0.0          0.0
 0.0  0.0  0.0  0.0   0.0          0.0     0.0  0.0  0.0   0.0          0.0
 0.0  0.0  0.0  0.0   4.33681e-19  0.0     0.0  0.0  0.0   0.0          0.0
 0.0  0.0  0.0  0.0   0.0          0.0     0.0  0.0  0.0   0.0          0.0
 0.0  0.0  0.0  0.0   0.0          0.0     0.0  0.0  0.0   0.0          0.0
 0.0  0.0  0.0  0.0   0.0          0.0  …  0.0  0.0  0.0   0.0          0.0
 0.0  0.0  0.0  0.0   0.0          0.0     0.0  0.0  0.0   0.0          0.0
 0.0  0.0  0.0  0.0   0.0          0.0     0.0  0.0  0.0   0.0          0.0
 0.0  0.0  0.0  0.0   0.0          0.0     0.0  0.0  0.0   0.0          0.0
 0.0  0.0  0.0  0.0   0.0          0.0     0.0  0.0  0.0   0.0          0.0
 0.0  0.0  0.0  0.0   0.0          0.0  …  0.0  0.0  0.0   0.0          0.0
 0.0  0.0  0.0  0.0   1.38778e-17  0.0     0.0  0.0  0.0   2.77556e-17  0.0
 0.0  0.0  0.0  0.0   0.0          0.0     0.0  0.0  0.0   0.0          0.0
```

On this PR I get **without GenericLinearAlgebra**:

```julia
julia> @time test_hessian(Float64)
  0.000071 seconds (215 allocations: 55.547 KiB)
9×9 Matrix{Float64}:
 0.0  0.0  0.0  0.0  0.0          0.0  0.0  0.0           0.0
 0.0  0.0  0.0  0.0  0.0          0.0  0.0  0.0           0.0
 0.0  0.0  0.0  0.0  0.0          0.0  0.0  0.0           0.0
 0.0  0.0  0.0  0.0  0.0          0.0  0.0  0.0           0.0
 0.0  0.0  0.0  0.0  0.0          0.0  0.0  2.77556e-17   0.0
 0.0  0.0  0.0  0.0  0.0          0.0  0.0  0.0           0.0
 0.0  0.0  0.0  0.0  0.0          0.0  0.0  0.0           0.0
 0.0  0.0  0.0  0.0  2.77556e-17  0.0  0.0  0.0          -2.77556e-17
 0.0  0.0  0.0  0.0  0.0          0.0  0.0  2.77556e-17   0.0

julia> @time test_hessian(ComplexF64)
  0.000491 seconds (1.00 k allocations: 477.500 KiB)
18×18 Matrix{Float64}:
 0.0  0.0  0.0  0.0   0.0          0.0  …  0.0  0.0  0.0  0.0   0.0          0.0
 0.0  0.0  0.0  0.0   0.0          0.0     0.0  0.0  0.0  0.0   0.0          0.0
 0.0  0.0  0.0  0.0   0.0          0.0     0.0  0.0  0.0  0.0   0.0          0.0
 0.0  0.0  0.0  0.0   0.0          0.0     0.0  0.0  0.0  0.0   0.0          0.0
 0.0  0.0  0.0  0.0   6.93889e-18  0.0     0.0  0.0  0.0  0.0  -1.38778e-17  0.0
 0.0  0.0  0.0  0.0   0.0          0.0  …  0.0  0.0  0.0  0.0   0.0          0.0
 0.0  0.0  0.0  0.0   0.0          0.0     0.0  0.0  0.0  0.0   0.0          0.0
 0.0  0.0  0.0  0.0   0.0          0.0     0.0  0.0  0.0  0.0   0.0          0.0
 0.0  0.0  0.0  0.0  -6.93889e-18  0.0     0.0  0.0  0.0  0.0   0.0          0.0
 0.0  0.0  0.0  0.0   0.0          0.0     0.0  0.0  0.0  0.0   0.0          0.0
 0.0  0.0  0.0  0.0   0.0          0.0  …  0.0  0.0  0.0  0.0   0.0          0.0
 0.0  0.0  0.0  0.0   0.0          0.0     0.0  0.0  0.0  0.0   0.0          0.0
 0.0  0.0  0.0  0.0   0.0          0.0     0.0  0.0  0.0  0.0   0.0          0.0
 0.0  0.0  0.0  0.0   0.0          0.0     0.0  0.0  0.0  0.0   0.0          0.0
 0.0  0.0  0.0  0.0   0.0          0.0     0.0  0.0  0.0  0.0   0.0          0.0
 0.0  0.0  0.0  0.0   0.0          0.0  …  0.0  0.0  0.0  0.0   0.0          0.0
 0.0  0.0  0.0  0.0   0.0          0.0     0.0  0.0  0.0  0.0   2.77556e-17  0.0
 0.0  0.0  0.0  0.0   0.0          0.0     0.0  0.0  0.0  0.0   0.0          0.0
```

---

Some benchmarks against master:

## master

```julia
julia> using ForwardDiff, Chairmarks, LinearAlgebra

## Symmetric{<:Real}
julia> @b rand(10) ForwardDiff.gradient(x -> sum(eigvals(Symmetric(x * x'))), _)
11.312 μs (110 allocs: 49.297 KiB)

julia> @b rand(10) ForwardDiff.gradient(x -> sum(eigen(Symmetric(x * x')).values), _)
29.083 μs (273 allocs: 115.469 KiB)

## Hermitian{<:Complex}
julia> @b rand(10) ForwardDiff.gradient(x -> sum(eigvals(Hermitian(complex.(x * x', x' * x)))), _)
71.625 μs (4136 allocs: 542.594 KiB)

julia> @b rand(10) ForwardDiff.gradient(x -> sum(eigen(Hermitian(complex.(x * x', x' * x))).values), _)
ERROR: MethodError: no method matching eigen!(::Hermitian{Complex{ForwardDiff.Dual{…}}, Matrix{Complex{…}}}; sortby::Nothing)
...

## SymTridiagonal{<:Real}
julia> @b rand(10) ForwardDiff.gradient(x -> sum(eigvals(SymTridiagonal(x, x[begin:(end-1)]))), _)
14.500 μs (128 allocs: 37.625 KiB)

julia> @b rand(10) ForwardDiff.gradient(x -> sum(eigen(SymTridiagonal(x, x[begin:(end-1)])).values), _)
32.625 μs (314 allocs: 101.422 KiB)

## Hermitian{<:Real}
julia> @b rand(10) ForwardDiff.gradient(x -> sum(eigvals(Hermitian(x * x'))), _)
ERROR: MethodError: no method matching eigvals!(::Hermitian{ForwardDiff.Dual{ForwardDiff.Tag{…}, Float64, 10}, Matrix{ForwardDiff.Dual{…}}}; sortby::Nothing)
The function `eigvals!` exists, but no method is defined for this combination of argument types.
...

julia> @b rand(10) ForwardDiff.gradient(x -> sum(eigen(Hermitian(x * x')).values), _)
ERROR: MethodError: no method matching eigen!(::Hermitian{ForwardDiff.Dual{ForwardDiff.Tag{…}, Float64, 10}, Matrix{ForwardDiff.Dual{…}}}; sortby::Nothing)
The function `eigen!` exists, but no method is defined for this combination of argument types.
...
```

## This PR

```julia
julia> using ForwardDiff, Chairmarks, LinearAlgebra

## Symmetric{<:Real}
julia> @b rand(10) ForwardDiff.gradient(x -> sum(eigvals(Symmetric(x * x'))), _)
15.834 μs (110 allocs: 49.297 KiB)

julia> @b rand(10) ForwardDiff.gradient(x -> sum(eigen(Symmetric(x * x')).values), _)
19.500 μs (133 allocs: 67.594 KiB)

## Hermitian{<:Complex}
julia> @b rand(10) ForwardDiff.gradient(x -> sum(eigvals(Hermitian(complex.(x * x', x' * x)))), _)
32.625 μs (136 allocs: 98.062 KiB)

julia> @b rand(10) ForwardDiff.gradient(x -> sum(eigen(Hermitian(complex.(x * x', x' * x))).values), _)
39.875 μs (159 allocs: 132.047 KiB)

## Symtridiagonal{<:Real}
julia> @b rand(10) ForwardDiff.gradient(x -> sum(eigvals(SymTridiagonal(x, x[begin:(end-1)]))), _)
10.938 μs (148 allocs: 31.062 KiB)

julia> @b rand(10) ForwardDiff.gradient(x -> sum(eigen(SymTridiagonal(x, x[begin:(end-1)])).values), _)
14.125 μs (171 allocs: 49.359 KiB)

## Hermitian{<:Real}
julia> @b rand(10) ForwardDiff.gradient(x -> sum(eigvals(Hermitian(x * x'))), _)
15.917 μs (110 allocs: 49.297 KiB)

julia> @b rand(10) ForwardDiff.gradient(x -> sum(eigen(Hermitian(x * x')).values), _)
19.250 μs (133 allocs: 67.594 KiB)
```

### DIFF
--- a/test/JacobianTest.jl
+++ b/test/JacobianTest.jl
@@ -238,22 +238,74 @@ end
 end
 
 @testset "eigen" begin
-    @test ForwardDiff.jacobian(x -> eigvals(SymTridiagonal(x, x[1:end-1])), [1.,2.]) ≈ [(1 - 3/sqrt(5))/2 (1 - 1/sqrt(5))/2 ; (1 + 3/sqrt(5))/2 (1 + 1/sqrt(5))/2]
-    @test ForwardDiff.jacobian(x -> eigvals(Symmetric(x*x')), [1.,2.]) ≈ [0 0; 2 4]
+    eigvals_symreal(x) = eigvals(Symmetric(x*x'))
+    eigvals_hermreal(x) = eigvals(Hermitian(x*x'))
+    eigvals_hermcomplex(x) = map(abs2, eigvals(Hermitian(complex.(x*x', x'*x))))
+    eigvals_symtridiag(x) = eigvals(SymTridiagonal(x, x[begin:(end - 1)]))
 
-    x0 = [1.0, 2.0];
-    ev1(x) = eigen(Symmetric(x*x')).vectors[:,1]
-    @test ForwardDiff.jacobian(ev1, x0) ≈ Calculus.finite_difference_jacobian(ev1, x0)
-    ev2(x) = eigen(SymTridiagonal(x, x[1:end-1])).vectors[:,1]
-    @test ForwardDiff.jacobian(ev2, x0) ≈ Calculus.finite_difference_jacobian(ev2, x0)
+    eigen_vals_symreal(x) = eigen(Symmetric(x*x')).values
+    eigen_vals_hermreal(x) = eigen(Hermitian(x*x')).values
+    eigen_vals_hermcomplex(x) = map(abs2, eigen(Hermitian(complex.(x*x', x'*x))).values)
+    eigen_vals_symtridiag(x) = eigen(SymTridiagonal(x, x[begin:(end - 1)])).values
 
-    x0_svector = SVector{2}(x0)
-    @test ForwardDiff.jacobian(ev1, x0_svector) isa SMatrix{2, 2}
-    @test ForwardDiff.jacobian(ev1, x0_svector) ≈ Calculus.finite_difference_jacobian(ev1, x0)
+    eigen_vec1_symreal(x) = eigen(Symmetric(x*x')).vectors[:,1]
+    eigen_vec1_hermreal(x) = eigen(Hermitian(x*x')).vectors[:,1]
+    eigen_vec1_hermcomplex(x) = map(abs2, eigen(Hermitian(complex.(x*x', x'*x))).vectors[:,1])
+    eigen_vec1_symtridiag(x) = eigen(SymTridiagonal(x, x[begin:(end - 1)])).vectors[:,1]
+    
+    # Note: SymTridiagonal is not supported for StaticArrays
+    for T in (Int, Float32, Float64), A in (Array, SArray, MArray)
+        if A <: StaticArrays.StaticArray
+            x = A{Tuple{2},T}(x0)
+            JT = A{Tuple{2,2},float(T)}
+        else
+            x = A{T,1}(x0)
+            JT = A{float(T),2}
+        end
 
-    x0_mvector = MVector{2}(x0)
-    @test ForwardDiff.jacobian(ev1, x0_mvector) isa MMatrix{2, 2}
-    @test ForwardDiff.jacobian(ev1, x0_mvector) ≈ Calculus.finite_difference_jacobian(ev1, x0)
+        # analytic solutions
+        @test ForwardDiff.jacobian(eigvals_symreal, x) ≈ [0 0; 2 4]
+        @test ForwardDiff.jacobian(eigvals_hermreal, x) ≈ [0 0; 2 4]
+        if !(x isa StaticArrays.StaticArray)
+           @test ForwardDiff.jacobian(eigvals_symtridiag, x) ≈ [(1 - 3/sqrt(5))/2 (1 - 1/sqrt(5))/2 ; (1 + 3/sqrt(5))/2 (1 + 1/sqrt(5))/2]
+        end
+
+        # eigen + eigvals
+        for ev in (
+            eigvals_symreal, eigvals_hermreal, eigvals_hermcomplex, eigvals_symtridiag,
+            eigen_vals_symreal, eigen_vals_hermreal, eigen_vals_hermcomplex, eigen_vals_symtridiag,            
+            eigen_vec1_symreal, eigen_vec1_hermreal, eigen_vec1_hermcomplex, eigen_vec1_symtridiag,
+        )
+            if x isa StaticArrays.StaticArray &&
+                (ev === eigvals_symtridiag || ev === eigen_vals_symtridiag || ev === eigen_vec1_symtridiag)
+                continue
+            end
+
+            # Chunk size can only be inferred for static arrays
+            if x isa StaticArrays.StaticArray
+                @test @inferred(ForwardDiff.jacobian(ev, x)) isa JT
+            else
+                @test ForwardDiff.jacobian(ev, x) isa JT
+            end
+            cfg = ForwardDiff.JacobianConfig(ev, x)
+            @test @inferred(ForwardDiff.jacobian(ev, x, cfg)) isa JT
+
+            @test ForwardDiff.jacobian(ev, x) ≈ Calculus.finite_difference_jacobian(ev, float.(x0))
+        end
+
+        # consistency of eigen and eigvals
+        for (eigvals, eigen_vals) in (
+            (eigvals_symreal, eigen_vals_symreal),
+            (eigvals_hermreal, eigen_vals_hermreal),
+            (eigvals_hermcomplex, eigen_vals_hermcomplex),
+            (eigvals_symtridiag, eigen_vals_symtridiag),
+        )
+            if x isa StaticArrays.StaticArray && eigvals === eigvals_symtridiag
+                continue
+            end
+            @test ForwardDiff.jacobian(eigvals, x) ≈ ForwardDiff.jacobian(eigen_vals, x)
+        end
+    end
 end
 
 @testset "type stability" begin
--- a/test/JacobianTest.jl
+++ b/test/JacobianTest.jl
@@ -254,6 +254,7 @@ end
     eigen_vec1_symtridiag(x) = eigen(SymTridiagonal(x, x[begin:(end - 1)])).vectors[:,1]
     
     # Note: SymTridiagonal is not supported for StaticArrays
+    x0 = [1.0, 2.0]
     for T in (Int, Float32, Float64), A in (Array, SArray, MArray)
         if A <: StaticArrays.StaticArray
             x = A{Tuple{2},T}(x0)
--- a/test/JacobianTest.jl
+++ b/test/JacobianTest.jl
@@ -312,6 +312,29 @@ end
             @test ForwardDiff.jacobian(eigvals, x) ≈ ForwardDiff.jacobian(eigen_vals, x)
         end
     end
+
+    # The matrices above are all of the form `x*x'` and hence symmetric, so `:U` and `:L` wrap the
+    # same matrix. Here the raw storage is deliberately not symmetric/Hermitian: the values and the
+    # partials both have to be read from the triangle that `uplo` selects.
+    @testset "uplo = :$uplo" for uplo in (:U, :L)
+        # `2*x[1]+x[2]^2` rather than something proportional to the off-diagonal entry, so that the
+        # eigenvector direction actually depends on `x` and the eigenvector test is not vacuous
+        raw_real(x) = [x[1] x[2]; 3*x[2] 2*x[1]+x[2]^2]
+        raw_complex(x) = complex.(raw_real(x), [0 x[1]; -2*x[2] 0])
+        x0 = [1.0, 2.0]
+
+        @testset "$name" for (name, wrap) in (
+            ("Symmetric{<:Real}", x -> Symmetric(raw_real(x), uplo)),
+            ("Hermitian{<:Real}", x -> Hermitian(raw_real(x), uplo)),
+            ("Hermitian{<:Complex}", x -> Hermitian(raw_complex(x), uplo)),
+        )
+            for ev in (x -> eigvals(wrap(x)),
+                       x -> eigen(wrap(x)).values,
+                       x -> map(abs2, eigen(wrap(x)).vectors[:, 1]))
+                @test ForwardDiff.jacobian(ev, x0) ≈ Calculus.finite_difference_jacobian(ev, x0)
+            end
+        end
+    end
 end
 
 @testset "type stability" begin
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ForwardDiff"
 uuid = "f6369f11-7733-5829-9624-2563aa707210"
-version = "1.4.5"
+version = "1.4.6"
 
 [deps]
 CommonSubexpressions = "bbf7d656-a473-5ed7-a52c-81e309532950"
--- a/ext/ForwardDiffStaticArraysExt.jl
+++ b/ext/ForwardDiffStaticArraysExt.jl
@@ -24,15 +24,28 @@ end
 @inline static_dual_eval(::Type{T}, f::F, x::StaticArray) where {T,F} = f(dualize(T, x))
 
 # To fix method ambiguity issues:
-function LinearAlgebra.eigvals(A::Symmetric{<:Dual{Tg,T,N}, <:StaticArrays.StaticMatrix}) where {Tg,T<:Real,N}
-    return ForwardDiff._eigvals(A)
+function LinearAlgebra.eigvals(A::Symmetric{Dual{T,V,N}, <:StaticArrays.StaticMatrix}) where {T,V<:Real,N}
+    return ForwardDiff._eigvals_hermitian(A)
 end
-function LinearAlgebra.eigen(A::Symmetric{<:Dual{Tg,T,N}, <:StaticArrays.StaticMatrix}) where {Tg,T<:Real,N}
-    return ForwardDiff._eigen(A)
+function LinearAlgebra.eigvals(A::Hermitian{Dual{T,V,N}, <:StaticArrays.StaticMatrix}) where {T,V<:Real,N}
+    return ForwardDiff._eigvals_hermitian(A)
+end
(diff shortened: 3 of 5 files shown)


test/JacobianTest.jl
```diff
@@ -238,22 +238,74 @@ end
...
+            x = A{T,1}(x0)
+            JT = A{float(T),2}
+        end
+
+        # analytic solutions
+        @test ForwardDiff.jacobian(eigvals_symreal, x) ≈ [0 0; 2 4]
+        @test ForwardDiff.jacobian(eigvals_hermreal, x) ≈ [0 0; 2 4]
+        if !(x isa StaticArrays.StaticArray)
+           @test ForwardDiff.jacobian(eigvals_symtridiag, x) ≈ [(1 - 3/sqrt(5))/2 (1 - 1/sqrt(5))/2 ; (1 + 3/sqrt(5))/2 (1 + 1/sqrt(5))/2]
+        end
```
mcabbott, line 271: I haven't looked very closely, but are there more tests somewhere that check the values?

In what's changed in this PR, I see one more line with `Calculus.finite_difference_jacobian(ev, float.(x0))` where I think `x0 = [1.0, 2.0]`.

devmotion: Yes, there are more in the lines below. Much more exhaustive than on master.

mcabbott: Sorry that wasn't clear. I see the many tests of different paths against each other, which is good. 

But what I meant is tests of any ForwardDiff path against something completely independent -- a known answer, or finite differences. To catch things like making the same logic error in writing both `_eigvals_hermitian` and `_eigen_hermitian` here.

devmotion: In the tests below, Jacobians of every test function with ForwardDiff are checked against finite differencing based Jacobians. The latter doesn't involve any ForwardDiff paths.

mcabbott: Sorry to be dense but where are you looking? Besides line 293, as mentioned:

> I see one more line with Calculus.finite_difference_jacobian(ev, float.(x0)) where I think x0 = [1.0, 2.0].

devmotion: It is line 293. Calculus doesn't support all vector types, hence the ForwardDiff results with `x` (the possibly static version of `x0`) is compared with the finite differencing result using the regular array `x0`.

mcabbott: Ok.

So the matrices being tested are these. Does it seem OK to test e.g. no negative eigenvalue for Symmetric{Real}, no zero eigenvalues for Hermitian{Complex}? Haven't thought hard but it just seemed a small sample.
```julia
julia> x
2-element Vector{Float64}:
 1.0
 2.0

julia> Symmetric(x*x')  # also wrapped in Hermitian
2×2 Symmetric{Float64, Matrix{Float64}}:
 1.0  2.0
 2.0  4.0

julia> eigvals(ans)
2-element Vector{Float64}:
 0.0
 5.0

julia> Hermitian(complex.(x*x', x'*x))
2×2 Hermitian{ComplexF64, Matrix{ComplexF64}}:
 1.0+0.0im  2.0+5.0im
 2.0-5.0im  4.0+0.0im

julia> eigvals(ans)
2-element Vector{Float64}:
 -3.0901699437494754
  8.090169943749475
```

mcabbott: (That said, when I tried various tests locally, I did not manage to break this.)

devmotion:
The PR fixes #780 as well: https://github.com/JuliaDiff/ForwardDiff.jl/issues/780#issuecomment-3373892904

devmotion:
Refreshed this branch: merged current `master`, added a regression test for the `uplo` handling, and fixed the one performance regression it had against `master`.

### Merged `master`

The base was ~10 months old. One trivial conflict: `master` inserted the `LinearAlgebra.givensAlgorithm` section (#783) immediately above the eigen section header this PR renames. The full suite passes locally; CI on this push will confirm.

### `uplo = :L` was fixed here but never tested

Every matrix in the `eigen` testset is of the form `x*x'` and hence symmetric, so `:U` and `:L` wrap the same matrix and the `uplo` handling in `_structured_value`/`_structured_partials` is never exercised. Added a case whose raw storage is deliberately not symmetric.

On `master` those fail, because `Symmetric(value.(parent(A)))` drops the `uplo` and so reads the values from the upper triangle while the partials are read through the wrapper and honour the actual one:

```julia
julia> M = [1.0 2.0; 7.0 4.0]; Ad = Dual{TestTag}.(M, [1.0 0.0; 0.0 0.0]);

julia> ForwardDiff.value.(eigvals(Symmetric(Ad, :L)))   # master
2-element Vector{Float64}:
 0.0
 5.0

julia> eigvals(Symmetric(M, :L))                        # what it should be
2-element Vector{Float64}:
 -4.658910531638176
  9.658910531638175
```

The Jacobian is off by 0.198 there; this branch matches finite differences to 1e-11 for `Symmetric{<:Real}`, `Hermitian{<:Real}` and `Hermitian{<:Complex}`, in both `:U` and `:L`, for `eigvals`, `eigen().values` and `eigen().vectors`.

One wrinkle worth recording: the eigenvector part of that test needs a matrix whose eigenvector direction actually depends on `x`. For `[x1 x2; x2 x1+x2]`, `tan(2θ) = 2x₂/(a-d) = -2` is constant, so the eigenvectors don't move, the derivative is identically zero, and the test would pass with or without the fix.

### Associating `Q' * (∂A * Q)` instead of `(Q' * ∂A) * Q`

Benchmarking against current `master` showed `eigvals(Symmetric)` had got slower — 11.67 μs / 127 allocs / 58.5 KiB against master's 10.62 μs / 107 allocs / 49.3 KiB — which was odd given `eigen(Symmetric)` got much faster.

The cause is that `(Q' * ∂A) * Q` starts with `Adjoint * Symmetric`, which has no BLAS specialization. It falls back to the generic `*(::AbstractMatrix, ::AbstractMatrix)`, which allocates an extra n² temporary per partial direction and skips `symm`/`hemm`:

```
Q' * M                     178.7 ns    2 allocs    944 bytes    # BLAS gemm
Q' * Symmetric(M)          273.5 ns    4 allocs    1.844 KiB    # generic fallback
mul!(C, Q', M)             168.8 ns    0 allocs
mul!(C, Q', Symmetric(M))  1003.0 ns   2 allocs                 # 6x slower
```

Re-associating fixes it without giving up the structure: `∂A * Q` is `Symmetric * Matrix`, which does dispatch to `symm`, and the remaining `Q' * _` is `Adjoint * Matrix`, i.e. `gemm`. Same result, one temporary fewer, and the halved flop count now actually materializes — at n = 100 the structured form finally beats a dense `∂A` (32.3 μs vs 52.0 μs), which it did not before (49.5 μs).

`SymTridiagonal` keeps its wrapper either way and also benefits (23.3 μs vs 31.7 μs at n = 100); there the structure makes the first product O(n²), so it is well worth preserving.

### Benchmarks against current `master`

`ForwardDiff.gradient(f, rand(10))`, minimum of `@b`:

| | `master` | this branch |
|---|---|---|
| `eigvals(Symmetric{<:Real})` | 10.62 μs, 107 allocs, 49.27 KiB | 10.54 μs, 107 allocs, 49.27 KiB |
| `eigen(Symmetric{<:Real}).values` | 25.54 μs, 267 allocs, 115.41 KiB | 13.58 μs, 130 allocs, 68.56 KiB |
| `eigvals(Hermitian{<:Real})` | `MethodError` | 10.83 μs, 107 allocs, 49.27 KiB |
| `eigen(Hermitian{<:Real}).values` | `MethodError` | 13.58 μs, 130 allocs, 68.56 KiB |
| `eigvals(Hermitian{<:Complex})` | 51.58 μs, 4133 allocs, 544.84 KiB | 25.75 μs, 133 allocs, 100.31 KiB |
| `eigen(Hermitian{<:Complex}).values` | `MethodError` | 33.38 μs, 156 allocs, 136.80 KiB |
| `eigvals(SymTridiagonal)` | 11.23 μs, 163 allocs, 31.64 KiB | 10.56 μs, 143 allocs, 31.02 KiB |
| `eigen(SymTridiagonal).values` | 26.38 μs, 380 allocs, 89.41 KiB | 14.00 μs, 166 allocs, 50.31 KiB |

No regressions left; the numbers in the PR description above predate current `master` and this change.

### Two things deliberately not changed

- `_to_duals`'s complex branch computes `tuple.(parts...)` twice, once for the real and once for the imaginary part. Hoisting it into a local looks like an obvious win but is a pessimization: inside the fused broadcast the tuples are never materialized, so hoisting forces an n²-element array of N-tuples into existence — 448 KiB against 240 KiB at n = 40, N = 8, to buy ~8% time. Left as is.
- The repeated-eigenvalue case still divides by a vanishing eigenvalue gap and yields `NaN`/`Inf` eigenvector derivatives, exactly as on `master` (and as in ChainRules, which tracks it as an open TODO). That is #788's territory, not this PR's.

JoshuaLampert:
What is the status of this PR? Would be nice to have this merged.

andreasnoack (approved):

PR #739, opened 2025-04-02
Do not seed structural zeros

devmotion:
Fixes #738.

Since I had made https://github.com/JuliaDiff/ForwardDiff.jl/pull/695 I felt somewhat responsible for #738. Defining `seed!` and `extract_gradient!` for triangular matrices seems to fix the example.

Probably this should b extended to additional matrix types.

PR #760, opened 2025-08-13
fix: regression in non-fast scalar indexing support

### LINKED ISSUE #759: "Thou shalt not iterate on GPU arrays" // #739 broke GPU compatibility
Consider the following minimal reproducible example:
```
using CUDA, ForwardDiff
x = cu(ones(3))
f(x) = x.^2
ForwardDiff.jacobian(f, x)
```

Result on ForwardDiff 1.0.0:
```
julia> ForwardDiff.jacobian(f, x)
3×3 CuArray{Float32, 2, CUDA.DeviceMemory}:
 2.0  0.0  0.0
 0.0  2.0  0.0
 0.0  0.0  2.0
```
which is obviously correct.

On ForwardDiff 1.0.1:
```
ERROR: Scalar indexing is disallowed.
Invocation of getindex resulted in scalar indexing of a GPU array.
This is typically caused by calling an iterating implementation of a method.
Such implementations *do not* execute on the GPU, but very slowly on the CPU,
and therefore should be avoided.

If you want to allow scalar iteration, use `allowscalar` or `@allowscalar`
to enable scalar iteration globally or for the operations in question.
```

Issue introduced in: https://github.com/JuliaDiff/ForwardDiff.jl/pull/739

avik-pal:
fixes #759

- [x] Jacobian support for GPU Arrays has been restored
- [x] `ForwardDiff.gradient` now supports GPU Arrays

cc @ChrisRackauckas @devmotion


avik-pal:
I have completely forgotten how to load extensions pre 1.9

Project.toml
```diff
@@ -1,11 +1,12 @@
 name = "ForwardDiff"
 uuid = "f6369f11-7733-5829-9624-2563aa707210"
-version = "1.0.1"
+version = "1.0.2"
 
 [deps]
 CommonSubexpressions = "bbf7d656-a473-5ed7-a52c-81e309532950"
 DiffResults = "163ba53b-c6d8-5494-b064-1a9d43ac40c5"
 DiffRules = "b552c78f-8df3-52c6-915a-8e097449b14b"
+GPUArraysCore = "46192b85-c4d5-4398-a991-12ede77f4527"
```
devmotion, line 9: IMO we should update the Julia compat to 1.10 instead of enforcing a new dependency on users of Julia < 1.10.
```suggestion
```

ext/ForwardDiffGPUArraysCoreExt.jl
```diff
@@ -0,0 +1,62 @@
+module ForwardDiffGPUArraysCoreExt
+
+using GPUArraysCore: AbstractGPUArray
+using ForwardDiff: ForwardDiff, Dual, Partials, npartials, partials
+
+function ForwardDiff.seed!(duals::AbstractGPUArray{Dual{T,V,N}}, x,
+                           seed::Partials{N,V} = zero(Partials{N,V})) where {T,V,N}
```
devmotion, line 7: It's not necessary to duplicate the default values here:
```suggestion
                           seed::Partials{N,V}) where {T,V,N}
```

ext/ForwardDiffGPUArraysCoreExt.jl
```diff
@@ -0,0 +1,62 @@
...
+    idxs = collect(ForwardDiff.structural_eachindex(duals, x))
+    duals[idxs] .= Dual{T,V,N}.(x[idxs], Ref(seed))
```
devmotion, line 9: I think we should use `view` here to reduce allocations, as in the old implementation (see https://github.com/JuliaDiff/ForwardDiff.jl/pull/472). And given #634, is there some overhead here from using broadcasting instead of `map!`?
```suggestion
    duals[idxs] .= Dual{T,V,N}.(view(x, idxs), Ref(seed))
```

ext/ForwardDiffGPUArraysCoreExt.jl
```diff
@@ -0,0 +1,62 @@
...
+    return duals
+end
+
+function ForwardDiff.seed!(duals::AbstractGPUArray{Dual{T,V,N}}, x,
+                           seeds::NTuple{N,Partials{N,V}}) where {T,V,N}
+    idxs = collect(ForwardDiff.structural_eachindex(duals, x))[1:N]
```
devmotion, line 15: Maybe just
```suggestion
    idxs = collect(Iterators.take(ForwardDiff.structural_eachindex(duals, x), N))
```
to reduce allocations?

ext/ForwardDiffGPUArraysCoreExt.jl
```diff
@@ -0,0 +1,62 @@
...
+    duals[idxs] .= Dual{T,V,N}.(x[idxs], seeds[1:N])
```
devmotion, line 16:
```suggestion
    duals[idxs] .= Dual{T,V,N}.(view(x, idxs), getindex.(Ref(seeds), 1:length(idxs)))
```

ext/ForwardDiffGPUArraysCoreExt.jl
```diff
@@ -0,0 +1,62 @@
...
+    return duals
+end
+
+function ForwardDiff.seed!(duals::AbstractGPUArray{Dual{T,V,N}}, x, index,
+                           seed::Partials{N,V} = zero(Partials{N,V})) where {T,V,N}
```
devmotion, line 21:
```suggestion
                           seed::Partials{N,V}) where {T,V,N}
```

ext/ForwardDiffGPUArraysCoreExt.jl
```diff
@@ -0,0 +1,62 @@
...
+    offset = index - 1
+    idxs = collect(Iterators.drop(ForwardDiff.structural_eachindex(duals, x), offset))
+    duals[idxs] .= Dual{T,V,N}.(x[idxs], Ref(seed))
```
devmotion, line 24:
```suggestion
    duals[idxs] .= Dual{T,V,N}.(view(x, idxs), Ref(seed))
```

ext/ForwardDiffGPUArraysCoreExt.jl
```diff
@@ -0,0 +1,62 @@
...
+    return duals
+end
+
+function ForwardDiff.seed!(duals::AbstractGPUArray{Dual{T,V,N}}, x, index,
+                           seeds::NTuple{N,Partials{N,V}}, chunksize = N) where {T,V,N}
```
devmotion, line 29:
```suggestion
                           seeds::NTuple{N,Partials{N,V}}, chunksize) where {T,V,N}
```

ext/ForwardDiffGPUArraysCoreExt.jl
devmotion: Another problem is actually that there could potentially be < N indices.

ext/ForwardDiffGPUArraysCoreExt.jl
```diff
@@ -0,0 +1,62 @@
...
+    offset = index - 1
+    idxs = collect(
+        Iterators.drop(ForwardDiff.structural_eachindex(duals, x), offset)
+    )[1:chunksize]
```
devmotion, line 33: Same as above:
```suggestion
    idxs = collect(
        Iterators.take(
            Iterators.drop(ForwardDiff.structural_eachindex(duals, x), offset),
            chunksize,
        ),
    )
```

ext/ForwardDiffGPUArraysCoreExt.jl
```diff
@@ -0,0 +1,62 @@
...
+    duals[idxs] .= Dual{T,V,N}.(x[idxs], seeds[1:chunksize])
```
devmotion, line 34:
```suggestion
    duals[idxs] .= Dual{T,V,N}.(view(x, idxs), getindex.(Ref(seeds), 1:length(idxs)))
```

ext/ForwardDiffGPUArraysCoreExt.jl
```diff
@@ -0,0 +1,62 @@
...
+    return duals
+end
+
+# gradient
+function ForwardDiff.extract_gradient!(::Type{T}, result::AbstractGPUArray,
+                                       dual::Dual) where {T}
+    # this closure is needed for gpu compilation
+    partial_fn(dual, i) = partials(T, dual, i)
```
devmotion, line 42: Since it appears multiple times and since you're closing over `dual` as well, I'd suggest defining it as a callable struct, something like
```julia
struct PartialsFn{T,D}
    dual::D
end
PartialsFn{T}(dual) where {T} = PartialsFn{T,typeof(d)}(d) 

(f::PartialsFn{T})(i) where {T} = partials(T, f.dual, i)
```
outside of this function.

ext/ForwardDiffGPUArraysCoreExt.jl
```diff
@@ -0,0 +1,62 @@
...
+
+    idxs = ForwardDiff.structural_eachindex(result)
+    result[idxs] .= partial_fn.(Ref(dual), 1:npartials(dual))
```
devmotion, line 45: To match the generic definition:
```suggestion
    idxs = collect(Iterators.take(ForwardDiff.structural_eachindex(result), npartials(dual)))
    result[idxs] .= partial_fn.(Ref(dual), 1:length(idxs))
```

ext/ForwardDiffGPUArraysCoreExt.jl
```diff
@@ -0,0 +1,62 @@
...
+
+function ForwardDiff.extract_gradient_chunk!(::Type{T}, result::AbstractGPUArray, dual,
+                                             index, chunksize) where {T}
+    # this closure is needed for gpu compilation
+    partial_fn(dual, i) = partials(T, dual, i)
+
+    offset = index - 1
+    idxs = collect(
+        Iterators.drop(ForwardDiff.structural_eachindex(result), offset)
+    )[1:chunksize]
```
devmotion, line 57:
```suggestion
    idxs = collect(
        Iterators.take(
                    Iterators.drop(ForwardDiff.structural_eachindex(result), offset),
            chunksize,
        )
    )
```

ext/ForwardDiffGPUArraysCoreExt.jl
```diff
@@ -0,0 +1,62 @@
...
+    result[idxs] .= partial_fn.(Ref(dual), 1:chunksize)
```
devmotion, line 58:
```suggestion
    result[idxs] .= partial_fn.(Ref(dual), 1:length(idxs))
```

ext/ForwardDiffGPUArraysCoreExt.jl
```diff
@@ -0,0 +1,67 @@
...
+    result[idxs] .= fn.(1:length(idxs))
+    return result
+end
+
+function ForwardDiff.extract_gradient_chunk!(::Type{T}, result::AbstractGPUArray, dual,
+                                             index, chunksize) where {T}
+    fn = PartialsFn{T}(dual)
+    offset = index - 1
+    idxs = collect(
+        _take(Iterators.drop(ForwardDiff.structural_eachindex(result), offset), chunksize)
```
devmotion, line 61:
```suggestion
        Iterators.take(Iterators.drop(ForwardDiff.structural_eachindex(result), offset), chunksize)
```

Project.toml
```diff
@@ -27,22 +28,25 @@ CommonSubexpressions = "0.3"
...
 DiffTests = "0.1"
+GPUArraysCore = "0.1, 0.2"
 IrrationalConstants = "0.1, 0.2"
 LogExpFunctions = "0.3"
 NaNMath = "1"
 Preferences = "1"
 SpecialFunctions = "1, 2"
 StaticArrays = "1.5"
-julia = "1.6"
+julia = "1.10"
```
devmotion, line 38: When performing this compat bump (IMO would be better in a separate PR but I'm OK with it being included in the PR), we have to do a few more clean ups, e.g., we have to go through the package and remove all unnecessary `VERSION` checks (such as https://github.com/JuliaDiff/ForwardDiff.jl/blob/d90a50581d3224730160fa02358d2546e8a0e18f/src/dual.jl#L303). Probably the reference to Julia 1.6 in https://github.com/JuliaDiff/ForwardDiff.jl/blob/d90a50581d3224730160fa02358d2546e8a0e18f/docs/src/user/advanced.md?plain=1#L89 should be removed as well.

devmotion (dismissed):

devmotion:
Yes, on the master branch seeding is (again) performed without broadcasting. Depending on the structural array type the set of indices are not readily available in an allocation-free broadcastable form (e.g. set of uppertriangular indices for `UpperTriangular`) and hence I reverted the code back to iterating over indices - without considering that it would break GPU compatibility.

If we want to avoid these allocations (and the broadcasting overhead) for non-GPU arrays, I don't immediately see how this issue could be solved by a generic implementation. Possibly the amount of code duplication could be reduced by introducing a helper function or branch that based on the type of the input array switches between broadcasting and iterating (presumably defaulting to iteration?), but even in this case it would be necessary to add an extension that ensures that GPU arrays use broadcasting.

Alternatively, we could default to using broadcasting (with the additional overhead of collecting the indices), and - as an additional optimization - only use iteration for a handful of selected base array types such as `Array`, `UpperTriangular{_,<:Matrix}`, etc. This wouldn't require an extension, but if we add such optimizations we would still require both an implementation with and one without broadcasting.

What are your thoughts @KristofferC?

src/utils.jl
```diff
@@ -0,0 +1,15 @@
+# overload for array types that
+@inline supports_fast_scalar_indexing(::Array) = true
```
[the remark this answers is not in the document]
avik-pal: StridedArray is too broad here

```
julia> SubArray{Float64, 2, JLArray{Float64, 2}, Tuple{UnitRange{Int64}, UnitRange{Int64}}, false} <: StridedArray
true
```

devmotion (commented):
I was hoping we could achieve it with a bit less code but maybe that's not possible.

Can you also add tests of https://github.com/JuliaDiff/ForwardDiff.jl/pull/739 and https://github.com/JuliaDiff/ForwardDiff.jl/pull/743 for these new branches?

ext/ForwardDiffStaticArraysExt.jl
```diff
@@ -10,6 +10,8 @@ using ForwardDiff: Dual, partials, npartials, Partials, GradientConfig, Jacobian
                    vector_mode_jacobian, vector_mode_jacobian!, valtype, value
 using DiffResults: DiffResult, ImmutableDiffResult, MutableDiffResult
 
+ForwardDiff.supports_fast_scalar_indexing(::StaticArray) = true
```
devmotion: Is this actually true? `StaticArray` is the abstract supertype, is it required that all subtypes support fast scalar indexing?

PR #762, opened 2025-08-14
Enable testing on prereleases

ChrisRackauckas:
Noticed while setting up the GPU infrastructure that ForwardDiff CI tests were missing pre.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,6 +26,7 @@ jobs:
           - 'min'
           - 'lts'
           - '1'
+          - 'pre'
         os:
           - ubuntu-latest
     steps:


ChrisRackauckas:
That was nightly, that's not very related to testing on prereleases which is much safer. Nightly can fail because Julia base itself is broken which makes it not necessarily a good signal for anything useful. Prereleases shouldn't be broken as they should already be pretty close to being a release. The allocation tests failing here means that they will be likely breaking when the release is made unless someone opens an issue, which is a pretty good thing to know. In fact, if someone doesn't open that MWE on Julia itself in like 1 month it will just be a regression 😅

ViralBShah:
In addition, `pre` just runs the regular release CI when you are not in a pre-release period and thus consuming a lot of wasteful CI cycles.

PR #771, opened 2025-09-08
prevent adding methods to the functions `>` and `>=`

nsajko:
As documented, the intended way to implement `>` is to add a method to `<`. Similarly with `>=`. A package should never add a method to either `>` or `>=`.

### DIFF
--- a/src/dual.jl
+++ b/src/dual.jl
@@ -383,7 +383,7 @@ end
 # Before PR#481 this loop ran over this list:
 # BINARY_PREDICATES = Symbol[:isequal, :isless, :<, :>, :(==), :(!=), :(<=), :(>=)]
 # Not a minimal set, as Base defines some in terms of others.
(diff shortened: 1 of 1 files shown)


nsajko:
This change decreases the unique invalidation count upon `using ForwardDiff` from `1504` to `784`.

Instructions to reproduce this result:

* Do `@snoop_invalidations using ForwardDiff` before loading any other package except for SnoopCompileCore. Not even the REPL or Pkg may be loaded before ForwardDiff, because they load StyledStrings, which does some piracy, causing invalidation, thus affecting the result.

* Use nightly Julia v1.13, commit JuliaLang/julia@865b8be.

Data on each invalidation:

* [before.json](https://github.com/user-attachments/files/22215610/before.json)

* [after.json](https://github.com/user-attachments/files/22215615/after.json)

PR #804, opened 2026-04-20
CI: Upgrade from setup-julia@v2 to v3

DilumAluthge:


DilumAluthge:
CI looks good on `min`.

I am seeing failures on `pre`, which seem to also occur on master (#805). @devmotion Are the `pre` failures a known issue?

devmotion (commented):
We don't want this PR, we want https://github.com/JuliaDiff/ForwardDiff.jl/pull/802.

Pre failures are known and caused by incorrect version bounds for JET in the registry IIRC.

Issue #798, opened 2026-03-25
Allocations when doing nested Jacobians with StaticArrays


albert-de-montserrat:
I am working a code where we do some nested Jacobians with StaticArrays, however I do not seem to manage to get it done without some allocations due to some internal type instability. 

The MWE below shows the allocations, and a weird behavior I have also seen while debugging our larger code, where type stability and allocations disappear if, after benchmarking (or simply calling) my function, I redefine `J_flat`:

```julia
using ForwardDiff, StaticArrays, Chairmarks

@inline f(x) = SVector(x[1]^2 * x[2], sin(x[1]) + x[2]^3)

function J_flat(x) 
    y = ForwardDiff.jacobian(f, x)
    SA[y[1], y[2]]
end

function nested_jacobian(x0::SVector{2, T}) where T
    ForwardDiff.jacobian(J_flat, x0)
end

x0 = SVector(1.0, 2.0)
@code_warntype nested_jacobian(x0) # shows a type instability
@b nested_jacobian($x0) # 131.447 ns (4 allocs: 176 bytes)

function J_flat(x) 
    y = ForwardDiff.jacobian(f, x)
    SA[y[1], y[2]]
end

@code_warntype nested_jacobian(x0) # type instability is gone
@b nested_jacobian($x0) # 26.594 ns
```
Redefining `J_flat` without benchmarking or calling `nested_jacobian` does not result in allocation-free call
```julia
using ForwardDiff, StaticArrays, Chairmarks

@inline f(x) = SVector(x[1]^2 * x[2], sin(x[1]) + x[2]^3)

function J_flat(x) 
    y = ForwardDiff.jacobian(f, x)
    SA[y[1], y[2]]
end

function nested_jacobian(x0::SVector{2, T}) where T
    ForwardDiff.jacobian(J_flat, x0)
end

x0 = SVector(1.0, 2.0)

function J_flat(x) 
    y = ForwardDiff.jacobian(f, x)
    SA[y[1], y[2]]
end

@code_warntype nested_jacobian(x0) # shows a type instability
@b nested_jacobian($x0) # 129.829 ns (4 allocs: 176 bytes)
```
Is there any workaround to avoid these allocations?

moble:
@albert-de-montserrat This can be closed now, right?

PR #810, opened 2026-05-31
Remove allocations of nested Jacobians of StaticArrays

albert-de-montserrat:
Addresses and fixes the issue raised in #798. 

The same MWE now yields:

```julia
using ForwardDiff, StaticArrays, Chairmarks

@inline f(x) = SVector(x[1]^2 * x[2], sin(x[1]) + x[2]^3)

function J_flat(x) 
    y = ForwardDiff.jacobian(f, x)
    SA[y[1], y[2]]
end

function nested_jacobian(x0::SVector{2, T}) where T
    ForwardDiff.jacobian(J_flat, x0)
end

x0 = SVector(1.0, 2.0)
```

```julia-repl
julia> @b nested_jacobian($x0)
19.031 ns
```

Looks like the compiler was giving up to inline `static_dual_eval(T, f, x)` in this case. Manual inlining seems to solve the problem.


devmotion (commented):
Can you add a test that fails on master but passes with these changes?

albert-de-montserrat:
Added the [following test](https://github.com/albert-de-montserrat/ForwardDiff.jl/blob/6ede2e6e429504e21c119161da8f244ffd7dc7ee/test/AllocationsTest.jl#L48), fails on master.

test/AllocationsTest.jl
```diff
@@ -44,4 +45,23 @@ end
     @test iszero(allocs_jacobian!())
 end
 
+@testset "allocation-free nested StaticArray jacobian" begin
+    # test that nested jacobians of StaticArrays do not allocate. 
+    # This is a regression test for issue #798, where the inner jacobian was allocating
+    @inline toy_f(x) = SVector(x[1]^2 * x[2], sin(x[1]) + x[2]^3)
```
devmotion: Can this be removed? I'd assume that it would be a more realistic use case:

```suggestion
    toy_f(x) = SVector(x[1]^2 * x[2], sin(x[1]) + x[2]^3)
```

albert-de-montserrat: Indeed, it's not strictly necessary.

devmotion:
It seems CI fails.

albert-de-montserrat:
Only `pre` fails now, but it seems `#master` is failing as well

devmotion (dismissed):
I can reproduce this locally: The test fails on master but passes with this PR. Can you update the version number?

devmotion (approved):

PR #807, opened 2026-05-13
Add tagdepth fast path to tag comparison

ChrisRackauckas:
Fixes https://github.com/SciML/OrdinaryDiffEq.jl/issues/3381 and superscedes https://github.com/SciML/OrdinaryDiffEq.jl/pull/3587 . Also fixes NonlinearSolve.jl master and superscedes https://github.com/SciML/NonlinearSolve.jl/pull/932

Superscedes https://github.com/JuliaDiff/ForwardDiff.jl/pull/724 and is a better solution to https://github.com/JuliaDiff/ForwardDiff.jl/issues/714.

The crux of the issue is that ForwardDiff.jl's tagging system is somewhat designed around the tag only being used once, i.e. the function is created, the derivative function is called, the tag is set for that derivative as a type of the function being differentiated, and therefore it's unique. Then this ends up working with nested differentiation because you call the inner function first, usually, before the outer function, or only do the combination, and so the tag ordering is set correctly.

Mixing tagging with precompilation then leads to this issue where it's possible for the outer tag to be precompiled before the inner tag. This makes the tag ordering the opposite, and what happens is then that the type promotion mechanism gets confused because it is tied to the tag ordering. This seems pretty fundamental because it's a useful property, it's the core property used to prevent perturbation confusion, but it means that this interaction between nested differentiation and precompilation ends up having odd bugs.

I tried working around this downstream (https://github.com/SciML/OrdinaryDiffEq.jl/pull/3587) but it was very nasty. Basically, you had to make sure you didn't have dual numbers automatically converting Float64s, as then sometimes it could convert to the inner type instead of the outer type, and it wouldn't do the normal conversion of first to the inner to then the wrapped outer type because doing so required the outer type to postdate the inner type.

But, this really then showcases that the bug truly only manifests with nested types. And if you have nested types, you know you don't have perturbation confusion if one tag is nested deeper than the other tag, because there are not the same number of partials. So in the case where the tag depths are not the same, you can do an alternative tag ordering since you will have already proven perturbations aren't confusing. And in that case, you can choose the deeper nested tags to just always be `<` the less deeper tags.

So added that and poof, tag nesting worked out in these cases with precompilation. So I think this captures the true crux of the problem and solves it at its core.

PR #815, opened 2026-06-11
Proposal: Tagging API

### LINKED ISSUE #813: tagging API?
Hello,

On #807, one option that i did not see considered is the posibility of defining an available (if experimental) tagging API, with `ForwardDiff.Tag` as it's only available implementation in ForwardDiff, via an `AbstractTag{F,V}` supertype.

For example, one package implementing `StaticTag{F,V} <: AbstractTag{F,V}` (in particular, https://github.com/JuliaDiff/ForwardDiff.jl/pull/807#issuecomment-4472479160 without the fallback)  could define their own static comparison mechanism, without compromising the current runtime capabilities that ForwardDiff already gives with their own `Tag` implementation, and erroring when there there is any ambiguity (closures of duals) instead of recurring to checking a global value.

Another example was an attempt done in #748. While the discussion on that PR was long and the PR was finally reverted, that PR inadvertly required following a currently private tagging API:

- `checktag`
- `≺`
- replacing most hardcoded Tag constructors by an overloadable `maketag` method

While i agree that the current `Tag` mechanism is adequate for most automatic differentiation workloads, new use cases, like completely compiled evaluations, merit thinking about the current invariants baked in the package, in my opinion. In this context, the implementation of a tagging API just shifts the burden from the ForwardDiff developer to the developer of the tagging mechanism.

longemen3000:
This PR proposes a Tagging API. the main objective is to provide external ForwardDiff users a way to control tag generation and comparison. The API is composed of:

- `AbstractTag{F,V}`: Abstract type for all tags following the Tagging API.
- `maketag(f::F,::Type{V})`: generates an instance of a tag, given an instance of a function `f` and a element type `V` will generate a tag that follows the tag API.

Some invariants:

- All tags are defined as a combination of a "callable" type `f` and a element type `V`.
- `V` should be a number (not enforced at the `AbstractTag` level, i don't know if any package does something weird here)
- dispatch for custom tags is done via dispatching on `f::F`, where `f` should behave according to the context it was used
- two tag types can be compared by the combination of the callable type and the element type. If they have the same callable type and the same element type, they are considered equal for the purposes of tag comparison.
  
Note that the invariants above do not imply that `maketagtype(f::F,::Type{V}) == MyTag{F,V}`. For example, `F` could have additional information about closed-over variables that we want to move over to `V`.

One application for this API is to create a `StaticTag` type that does comparisons via inspecting the fields of the closures (like what's proposed in #807), or for a `HashedTag` type (like what was tried in #748). More importantly, experimenting with new tag types can be done in external packages, instead of PRs to `ForwardDiff`.

Closes #813

Any comments are really appreciated.

### DIFF
--- a/ext/ForwardDiffStaticArraysExt.jl
+++ b/ext/ForwardDiffStaticArraysExt.jl
@@ -51,12 +51,12 @@ ForwardDiff._lyap_div!!(A::StaticArrays.MMatrix, λ::AbstractVector) = ForwardDi
 end
 
 @inline function ForwardDiff.vector_mode_gradient(f::F, x::StaticArray) where {F}
-    T = typeof(maketag(f,eltype(x)))
+    T = maketagtype(f,eltype(x))
     return extract_gradient(T, f(dualize(T, x)), x)
 end
 
 @inline function ForwardDiff.vector_mode_gradient!(result, f::F, x::StaticArray) where {F}
-    T = typeof(maketag(f,eltype(x)))
+    T = maketagtype(f,eltype(x))
     return extract_gradient!(T, result, f(dualize(T, x)))
 end
 
@@ -81,7 +81,7 @@ end
 end
 
 @inline function ForwardDiff.vector_mode_jacobian(f::F, x::StaticArray) where {F}
-    T = typeof(maketag(f,eltype(x)))
+    T = maketagtype(f,eltype(x))
     return extract_jacobian(T, f(dualize(T, x)), x)
 end
 
@@ -91,7 +91,7 @@ function extract_jacobian(::Type{T}, ydual::AbstractArray, x::StaticArray) where
 end
 
 @inline function ForwardDiff.vector_mode_jacobian!(result, f::F, x::StaticArray) where {F}
-    T = typeof(maketag(f,eltype(x)))
+    T = maketagtype(f,eltype(x))
     ydual = f(dualize(T, x))
     result = extract_jacobian!(T, result, ydual, length(x))
     result = extract_value!(T, result, ydual)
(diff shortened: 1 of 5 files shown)


devmotion:
> for a `HashedTag` type (like what was tried in #748)

Apparently that approach is completely unsound: https://github.com/JuliaDiff/ForwardDiff.jl/pull/748#issuecomment-3850028185

longemen3000:
Yes, that particular PR was unsound, but there are more options for tags defined at runtime, if someone wants to give it a try (one of the alternatives in #807 was via objectid)

longemen3000:
tests are finally passing (except for `pre`, because of JET version incompatibility)

devmotion:
I think the motivation for this change is missing and its design is unclear without clearly designed alternative tagging systems. Without such it's also not clear to me whether this would solve any actual problems and whether eg combinations of different tagging systems would create new problems.

longemen3000:
>I think the motivation for this change is missing and its design is unclear without clearly designed alternative tagging systems. 

The main motivation for this PR is to allow experimenting with alternative tag designs. #748 is closed and #807 is stalled in its current form. For a proof of concept, i made an alternative tagging system, based on https://github.com/JuliaDiff/ForwardDiff.jl/pull/807#issuecomment-4472479160 along with a closure workaround
https://github.com/ClapeyronThermo/Clapeyron.jl/tree/AD-custom-tags/lib/StaticForwardDiffTags

That tag system in particular takes the tradeoff of a more cumbersome function definition in exchange of faster compile times; while the package itself does not have tests, the branch is used as AD provider for Clapeyron.jl, where we use up to 4th order nested duals for some functions.
Most of the overloads of that package are just changing the hardcoded `Tag` in ForwardDiff functions for `maketagtype`.

> whether eg combinations of different tagging systems would create new problems.

On the interaction between two tagging systems, and in particular, the interaction of a external tagging system and the current `ForwardDiff.Tag`, I see no other option that just register new tags (with the same types) and fallback to `tagcount`. 
Another, more restrictive option would be to just block two tags systems from interacting. (probably useful if one wants to be sure that no code result is determined at runtime, like what `tagcount` does

(the extension code is broken at the moment if you want to load it as a standalone package, but the rest should work)

PR #820, opened 2026-07-03
Order nested Dual tags by provable containment (supersedes #807)

### LINKED ISSUE #714: Precompilation can lead to Dual tag ordering problems

ChrisRackauckas-Claude:
This is the updated version of #807, superseding it: same goal (fix #714's precompilation-inverted tag ordering, and with it [SciML/OrdinaryDiffEq.jl#3381](https://github.com/SciML/OrdinaryDiffEq.jl/issues/3381) and [SciML/NonlinearSolve.jl#932](https://github.com/SciML/NonlinearSolve.jl/pull/932)), but with the depth-only fast path replaced by the containment proof discussed in #807's review thread. Opened as a separate PR only because #807's head branch can't be pushed to from this account; @ChrisRackauckas can instead pull this into #807's branch to keep the discussion in one place:

```
git fetch https://github.com/ChrisRackauckas-Claude/ForwardDiff.jl containment-tag-ordering
git push git@github.com:JuliaDiff/ForwardDiff.jl.git FETCH_HEAD:ChrisRackauckas-patch-1 --force-with-lease
```

**Note: please treat this as pending until reviewed by @ChrisRackauckas.**

## The problem

`tagcount` is a `@generated` function whose literal is baked at first compile, so precompilation can bake an outer and an inner tag's counts in inverted order (#714). Tag ordering (`≺`) is a pure `tagcount` comparison, so nested differentiation then composes Duals in the wrong nesting order — the stochastic `DualMismatchError` / `MethodError: no method matching Float64(::Dual{...})` class of failures seen downstream.

## The fix: order by provable containment, fall back to tagcount

#807's original depth-only fast path had two confirmed counterexamples (thanks @devmotion, confirmed by @riftsim-jarod): `tagdepth` read nesting off `V` only, so it both regressed a 3-level case where nesting lives in the closure type `F`, and missed the same-depth case where both tags have `V === Float64`.

This version replaces the depth heuristic with a containment proof. `containstag(T1, T2)` walks `T1`'s type structure at compile time — the seeded value type `V`, plus the function type `F`'s type parameters and concrete field types (closure captures) — and checks whether `T2` genuinely appears inside it. If it does, `T1` was necessarily created while `T2`'s derivative was already in progress, so `Dual{T1}` must compose outside `Dual{T2}`, regardless of what `tagcount` says:

```julia
    containstag(Tag{F1,V1}, Tag{F2,V2}) && return false
    containstag(Tag{F2,V2}, Tag{F1,V1}) && return true
    return tagcount(Tag{F1,V1}) < tagcount(Tag{F2,V2})
end
```

When neither tag provably nests the other, behavior is exactly master's. The walk is a `@generated` function (cycle-guarded, depth-bounded at 32), so `≺` still folds to a constant at compile time. This is the "prove that one tag genuinely nests the other" safety check from the #807 thread — extended through `F` as well as `V`, which is what also fixes the same-`V` case (the nesting there is only visible through the callable's fields).

## Verification (all run locally)

| Check | Result |
|---|---|
| Full FD suite, Julia 1.11, default mode | 9045/9045 |
| Full FD suite, Julia 1.12, NaN-safe mode | 9129/9129 |
| [8-case downstream matrix](https://github.com/JuliaDiff/ForwardDiff.jl/pull/807#issuecomment-4472620435) (6× OrdinaryDiffEq spec/u0 combos + both counterexamples), Julia 1.12 | **8/8** (depth-only #807: 6/8; master: 1/8) |
| @devmotion example 1 (3-level, constant inner seed) | `12.0` — #807's regression gone |
| @devmotion example 2 (same-depth, inverted tagcount) | `cos(0.5) - 0.5·sin(0.5)` — fixed |
| SciML/OrdinaryDiffEq.jl#3381 original MWE | no longer throws; identical to `AutoFiniteDiff` reference (9 digits); FD Jacobian vs analytic: `3e-8` |

Regression tests added to `test/ConfusionTest.jl`, covering containment ordering under explicitly inverted tagcounts in both shapes (`V`-nesting and `F`-capture) plus the 3-level constant-seed case. The `F`-capture test throws `DualMismatchError` on master and passes here.

Two caveats for review:

1. The hybrid ordering (containment first, tagcount fallback) is not a proven total order: a three-way mix where A nests B but A–C and B–C resolve by tagcount could in principle order inconsistently. Master's pure tagcount has the same in-principle exposure once precompilation inverts counts, and the containment direction is the semantically forced one, so this is strictly an improvement — but it is a heuristic layered on tagcount, not a replacement for it.
2. `containstag` only unwraps `Union`/`UnionAll` and walks concrete field types; pathological/truncated cases fall back to tagcount (master behavior) rather than guessing.

The "Julia pre" CI failures are pre-existing — all six pre jobs fail identically on the latest master push run (JET `@test_opt` in `test/QATest.jl` flagging Julia 1.13-rc1 Base broadcast-inference regressions; reproduced locally on unmodified master).

### DIFF
--- a/src/config.jl
+++ b/src/config.jl
@@ -20,10 +20,16 @@ end
 
 Tag(::Nothing, ::Type{V}) where {V} = nothing
 
+@inline tagdepth(::Type) = 0
+@inline tagdepth(::Type{<:Dual{T,V,N}}) where {T,V,N} = 1 + tagdepth(V)
+@inline tagdepth(::Type{<:Tag{F,V}}) where {F,V} = 1 + tagdepth(V)
 
 @inline function ≺(::Type{Tag{F1,V1}}, ::Type{Tag{F2,V2}}) where {F1,V1,F2,V2}
-    tagcount(Tag{F1,V1}) < tagcount(Tag{F2,V2})
-end
+      d1 = tagdepth(Tag{F1,V1})
+      d2 = tagdepth(Tag{F2,V2})
+      d1 != d2 && return d1 < d2
+      return tagcount(Tag{F1,V1}) < tagcount(Tag{F2,V2})
+  end
 
 struct InvalidTagException{E,O} <: Exception
 end
--- a/src/config.jl
+++ b/src/config.jl
@@ -20,16 +20,44 @@ end
 
 Tag(::Nothing, ::Type{V}) where {V} = nothing
 
-@inline tagdepth(::Type) = 0
-@inline tagdepth(::Type{<:Dual{T,V,N}}) where {T,V,N} = 1 + tagdepth(V)
-@inline tagdepth(::Type{<:Tag{F,V}}) where {F,V} = 1 + tagdepth(V)
+# A tag `T1` provably nests a tag `T2` when `T2` appears inside `T1`'s type
+# structure: in the seeded value type `V1`, or captured by the function type `F1`
+# (closure/struct fields and type parameters). In that case `T1` was necessarily
+# created while `T2`'s derivative was already in progress, so `Dual{T1}` must be
+# composed outside `Dual{T2}` — independent of `tagcount`, whose values can be
+# baked in an arbitrary order by precompilation (see #714).
+function _containstag(@nospecialize(T), @nospecialize(target), seen::Base.IdSet{Any}, depth::Int)
+    T === target && return true
+    depth <= 0 && return false
+    if T isa Union
+        return _containstag(T.a, target, seen, depth - 1) ||
+               _containstag(T.b, target, seen, depth - 1)
+    elseif T isa UnionAll
+        return _containstag(Base.unwrap_unionall(T), target, seen, depth - 1)
+    end
+    T isa DataType || return false
+    T in seen && return false
+    push!(seen, T)
+    for p in T.parameters
+        p isa Type && _containstag(p, target, seen, depth - 1) && return true
+    end
+    if isconcretetype(T) && isstructtype(T)
+        for ft in fieldtypes(T)
+            ft isa Type && _containstag(ft, target, seen, depth - 1) && return true
+        end
+    end
+    return false
+end
+
+@generated function containstag(::Type{T1}, ::Type{T2}) where {T1,T2}
+    return _containstag(T1, T2, Base.IdSet{Any}(), 32) ? :(true) : :(false)
+end
 
 @inline function ≺(::Type{Tag{F1,V1}}, ::Type{Tag{F2,V2}}) where {F1,V1,F2,V2}
-      d1 = tagdepth(Tag{F1,V1})
-      d2 = tagdepth(Tag{F2,V2})
-      d1 != d2 && return d1 < d2
-      return tagcount(Tag{F1,V1}) < tagcount(Tag{F2,V2})
-  end
+    containstag(Tag{F1,V1}, Tag{F2,V2}) && return false
+    containstag(Tag{F2,V2}, Tag{F1,V1}) && return true
+    return tagcount(Tag{F1,V1}) < tagcount(Tag{F2,V2})
+end
 
 struct InvalidTagException{E,O} <: Exception
 end
--- a/test/ConfusionTest.jl
+++ b/test/ConfusionTest.jl
@@ -72,5 +72,56 @@ end
     end
 end == 0.0
 
+# Nested differentiation must not depend on tagcount instantiation order (#714) #
+#-------------------------------------------------------------------------------#
+
+# containstag: nesting through the seeded value type V
+struct TagOrderOuterMarker end
+struct TagOrderInnerMarker end
+let Tag = ForwardDiff.Tag, Dual = ForwardDiff.Dual
+    Touter = Tag{TagOrderOuterMarker, Float64}
+    Tinner = Tag{TagOrderInnerMarker, Dual{Touter, Float64, 1}}
+    @test ForwardDiff.containstag(Tinner, Touter)
+    @test !ForwardDiff.containstag(Touter, Tinner)
+    # bake tagcounts in inverted order: containment must win regardless
+    ForwardDiff.tagcount(Tinner)
+    ForwardDiff.tagcount(Touter)
+    @test ForwardDiff.:(≺)(Touter, Tinner)
+    @test !ForwardDiff.:(≺)(Tinner, Touter)
+end
+
+# Second derivative with tagcount baked in inverted order, as precompilation can
+# do: the inner tag nests the outer through V, so ordering must not consult
+# tagcount at all.
+struct TagOrderInnerV end
+(::TagOrderInnerV)(y) = y^3
+struct TagOrderOuterV end
+(::TagOrderOuterV)(x) = ForwardDiff.derivative(TagOrderInnerV(), x)
+ForwardDiff.tagcount(ForwardDiff.Tag{TagOrderInnerV, ForwardDiff.Dual{ForwardDiff.Tag{TagOrderOuterV, Float64}, Float64, 1}})
+ForwardDiff.tagcount(ForwardDiff.Tag{TagOrderOuterV, Float64})
+@test ForwardDiff.derivative(TagOrderOuterV(), 2.0) ≈ 12.0
+
+# Same with the outer perturbation entering through a capture in F (both tags
+# have V === Float64): nesting is only visible through the callable's fields.
+struct TagOrderOuterF end
+struct TagOrderInnerF
+    x_dual::ForwardDiff.Dual{ForwardDiff.Tag{TagOrderOuterF, Float64}, Float64, 1}
+end
+(c::TagOrderInnerF)(y) = sin(c.x_dual * y)
+(::TagOrderOuterF)(x_dual::ForwardDiff.Dual{ForwardDiff.Tag{TagOrderOuterF, Float64}, Float64, 1}) =
+    ForwardDiff.derivative(TagOrderInnerF(x_dual), 1.0)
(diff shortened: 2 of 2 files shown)


devmotion:
I'm on vacation and will think this through more carefully when I'm back end of July.

ChrisRackauckas:
Bump

devmotion:
Since the tag system is one of the most critical parts of ForwardDiff I wanted to go through this more carefully than usual, and did so with the help of Claude. The containment check isn't sound in the form implemented here, and in one case it turns code that works on master into an error.

A few things that do check out: `containstag` only depends on its type arguments, so in contrast to `tagcount` precompilation can't invert it. `≺` still folds to a constant (`code_typed` gives `Any[:(return false)]`), so there's no runtime cost, and the walk doesn't touch global state, unlike `tagcount`. `test/ConfusionTest.jl` passes for me on this branch.

### The containment argument doesn't hold for `F`

The comment in `src/config.jl` claims

> In that case `T1` was necessarily created while `T2`'s derivative was already in progress

AFAICT this isn't true. That `T2` occurs somewhere in `F1` only means the type existed, not that the corresponding perturbation is an ancestor. A callable can hold storage for an AD pass that runs *inside* it:

```julia
using ForwardDiff
const FD = ForwardDiff

struct Inner
    p
end
(i::Inner)(y) = y * i.p
const T2 = FD.Tag{Inner, Float64}

struct Outer
    buf::Vector{FD.Dual{T2, Float64, 1}}
end
(o::Outer)(x) = FD.derivative(Inner(x), 2.0)
const T1 = FD.Tag{Outer, Float64}

FD.derivative(Outer(FD.Dual{T2,Float64,1}[]), 1.0)  # d/dx d/dy (y * x)
```

On master this gives `1.0`, and `tagcount(T1) = 0 < tagcount(T2) = 1`, i.e. the existing ordering is correct here. On this branch it throws

```
Cannot determine ordering of Dual tags ForwardDiff.Tag{Inner, Float64} and ForwardDiff.Tag{Outer, Float64}
```

because `containstag(T1, T2)` is `true` — `T2` is reachable through the parameter of `Vector{Dual{T2,Float64,1}}` — so the correct `tagcount` ordering is overridden.

The ingredients are a preallocated Dual buffer in the differentiated object and something that hides the reverse direction from the walk; in the example that's the untyped field in `Inner`, but an abstract field type, a `Ref{Any}`, or a reinterpreted `Vector{Float64}` buffer have the same effect.

For the `V` part of the check the argument does hold: `Tag{F,Dual{T2,...}}` means the point being differentiated is a `Dual{T2}`. It's only the walk through `F` that's affected.

Possible directions:

1. Fix the comment. The counterexample shows it states a proof where there is only a heuristic, so this applies regardless of the rest.
2. Don't descend into element types of `AbstractArray`/`Ref` when walking `F`. That fixes the example above, and all three new tests still pass (`TagOrderInnerF` reaches its tag through a direct `Dual` field; the `V`-nesting and 3-level cases don't go through containers either). The price is that `ConfusionTest.jl:14`-style cases, where the inner closure captures a `v::Vector{Dual{...}}`, fall back to `tagcount` — those only need the containment path if `tagcount` is already inverted.
3. Keep it as-is and state that the `F` part can invert correct orderings.

### `≺` is not transitive anymore

Caveat 1 in the PR description says a three-way inconsistency is possible in principle. It's reachable. With `containstag(TB, TA) === true` and counts baked as `B = 0, C = 1, A = 2`:

```
A ≺ B = true   # containment
B ≺ C = true   # tagcount
C ≺ A = true   # tagcount
```

`@define_ternary_dual_op` (`src/dual.jl:174-177`) picks the maximum assuming a total order, so with a cycle it can pick a non-maximal tag. `muladd`/`fma` don't error on this input, and the resulting nesting `Dual{TC}(Dual{TB}(Dual{TA}))` respects the one provable constraint while master's `Dual{TA}(Dual{TC}(Dual{TB}))` doesn't. I couldn't construct a case where the intransitivity alone makes things worse than master.

Separately, `promote_rule` (`src/dual.jl:436`) needs exactly one of `T1 ≺ T2` / `T2 ≺ T1` to hold. If both `containstag` directions were `true`, `≺` returns `false` both ways and `promote_type` doesn't converge. That requires a type cycle through both tags and I didn't find a way to construct one, but it's two lines to rule out with values that are already computed:

```julia
    c12 = containstag(Tag{F1,V1}, Tag{F2,V2})
    c21 = containstag(Tag{F2,V2}, Tag{F1,V1})
    c12 !== c21 && return c21
    return tagcount(Tag{F1,V1}) < tagcount(Tag{F2,V2})
end
```

### `containstag` depends on field declaration order

`seen` is a global visited set while the depth budget is per path, so a type first reached via a long path is pruned permanently even if a shorter path later would have had enough budget. Two structs with the same set of reachable types can disagree:

```julia
chain(leaf, k) = k == 0 ? leaf : chain(Tuple{leaf}, k - 1)
X = chain(TT, 20)
LONG = chain(X, 20)

struct RootLongFirst;  a::LONG; b::X;    end
struct RootShortFirst; a::X;    b::LONG; end

containstag(X, TT)               # true
containstag(RootLongFirst,  TT)  # false
containstag(RootShortFirst, TT)  # true
```

`seen` alone already guarantees termination — the set of reachable types is finite, and the guard handles a self-referential `mutable struct` field cycle fine. `depth` isn't needed to terminate; it adds incompleteness and this order dependence. Alternatives would be dropping it, memoizing on `(T, depth)`, or a global budget on the number of visited types. Minor: the budget is also decremented for `Union`/`UnionAll` unwrapping, which isn't structural descent.

### Smaller things

- `Base.IdSet` isn't public API. `Set{Any}` works the same here since types compare by `===` under `isequal`.
- No version bump in `Project.toml`.
- Parameters are walked for every `DataType` but fields only for `isconcretetype && isstructtype`. The reason (`TypeVar`s in non-concrete `fieldtypes`) could use a one-line comment.
- Both caveats from the PR description are only in the PR description. `≺` is unexported, not declared `public`, and doesn't appear anywhere in `docs/src` (`api.md` uses explicit `@docs` blocks, so its docstring isn't rendered), so the source is the only place they'd be visible. The total-order assumption is currently only implicit in `src/dual.jl:174-177` and `src/dual.jl:436`.

### Tests

- The new tests depend on `tagcount` being baked in inverted order but never check it. It does hold currently (`tagcount(Tinner) = 0 < tagcount(Touter) = 1`), but without something like `@test ForwardDiff.tagcount(Tinner) < ForwardDiff.tagcount(Touter)` they can silently turn into no-ops.
- Nothing tests that `≺` still folds to a constant, which is what the performance argument rests on.
- The false-positive case above isn't covered.
- The cycle guard and the `Union` branch work but aren't tested; the depth limit isn't tested at all.
- `ForwardDiff.:≺` would be consistent with `test/DualTest.jl:29` and `test/GradientTest.jl:18`, unlike `ForwardDiff.:(≺)`. And `let Tag = ForwardDiff.Tag, Dual = ForwardDiff.Dual` shadows two names that are used unqualified elsewhere in the test suite.

ChrisRackauckas:
Okay. I don't think it's possible for a fix to keep transitive. Is the answer then to just document that ForwardDiff is incompatible with precompilation? Or do you have another idea?

PR #821, opened 2026-07-03
Unseed only the chunk, and split unseed! out of seed!

ChrisRackauckas-Claude:
## Summary

Standalone against `master` (previously stacked on #816; now independent of it — the two
touch the same function and will need a trivial conflict resolution whichever lands second).

In ForwardDiff 0.10, the chunk-mode "unseed" call `seed!(duals, x, index, seed)` wrote
exactly the `N` chunk elements starting at `index`. The 1.x rewrite changed it to write
from `index` to the **end** of the array:

```julia
idxs = Iterators.drop(structural_eachindex(duals, x), offset)
for idx in idxs   # no take(N) — runs to the end
```

Chunk mode calls this after every chunk (`seed!(xdual, x, i)` in
`chunk_mode_gradient_expr` / `chunk_mode_jacobian_expr`) purely to clear the chunk it
just seeded — the rest of the array is already unseeded (everything is zeroed once
up front, and every other chunk clears itself). So writing to the end does
Σ(n − i) ≈ n²/2N redundant dual writes per gradient/jacobian sweep. At n = 100000 with
chunk 12 and `Dual{…,Float64,12}` (104 bytes each) that is ~40 GB of memory traffic.

This restores the 0.10 behavior: write at most `N` elements starting at `index`
(`Iterators.take(…, N)`). The 3-arg `seed!` form has no other callers in the package.

Second commit, from review: the two `seed!` methods whose `seed::Partials` argument
defaulted to `zero(Partials)` were only ever called to *un*seed, so they are now
`unseed!(duals, x)` and `unseed!(duals, x, index)` with that argument dropped (it was
never passed outside the allocation test). The `NTuple{N,Partials}` seeding methods keep
the `seed!` name and are unchanged.

## Benchmarks

Julia 1.12, `gradient!` of `f(x) = sum(abs2, x)` with default chunk 12, median of 11 runs:

| | master | this PR |
|---|---|---|
| `gradient!` n=1000 | 502 μs | 279 μs |
| `gradient!` n=10000 | 51.3 ms | 30.4 ms |
| `gradient!` n=100000 | 5.39 s | 2.99 s |
| unseed sweep only, n=100000 | 2.49 s | 0.33 s |

The remaining unseed cost is the `Iterators.drop(structural_eachindex(…), index-1)` walk,
which is still O(index) per chunk; #816's dense fast path removes that separately.

## Tests

No new tests: chunk-vs-vector-mode consistency (including non-divisible chunk sizes and
`UpperTriangular`/`Diagonal` inputs) is covered by the existing suite, which passes
locally (`9134/9134` on Julia 1.12). Verified additionally that chunked results match
full-chunk references for n ∈ {5, 12, 13, 24, 25, 100} × chunk ∈ {1, 3, 5, 12}, for
`gradient`, `jacobian`, and the mutating `jacobian(f!, y, x)` form, plus `UpperTriangular`
/ `Diagonal` structural inputs and a `BigFloat` (non-isbits) input.

---

> [!NOTE]
> Opened as a draft by an agent on behalf of @ChrisRackauckas. Please ignore
> until reviewed by @ChrisRackauckas.

### DIFF
--- a/src/apiutils.jl
+++ b/src/apiutils.jl
@@ -27,7 +27,7 @@ end
 function vector_mode_dual_eval!(f!::F, cfg::JacobianConfig, y, x) where {F}
     ydual, xdual = cfg.duals
     seed!(xdual, x, cfg.seeds)
-    unseed!(ydual, y)
+    seed_zero_partials!(ydual, y)
     f!(ydual, xdual)
     return ydual
 end
@@ -70,32 +70,24 @@ function structural_eachindex(x::Diagonal, y::AbstractArray)
     return diagind(x)
 end
 
-# Copies the values of `x` into `duals` with zero partials, i.e. removes any seeds
-# `duals` is currently carrying.
-function unseed!(duals::AbstractArray{Dual{T,V,N}}, x) where {T,V,N}
-    seed = zero(Partials{N,V})
-    if isbitstype(V)
-        for idx in structural_eachindex(duals, x)
-            duals[idx] = Dual{T,V,N}(x[idx], seed)
-        end
-    else
-        for idx in structural_eachindex(duals, x)
-            if isassigned(x, idx)
-                duals[idx] = Dual{T,V,N}(x[idx], seed)
-            else
-                Base._unsetindex!(duals, idx)
-            end
-        end
-    end
-    return duals
+# Copies the values of `x` into `duals` with zero partials. Used both to remove seeds `duals` is
+# currently carrying and to initialize a freshly allocated work buffer, whose elements must all be
+# written before the target function reads them.
+seed_zero_partials!(duals::AbstractArray{Dual{T,V,N}}, x) where {T,V,N} =
+    _seed_zero_partials!(duals, x, structural_eachindex(duals, x))
+
+# Zeroes the partials of `count` elements starting at structural position `index`. Chunk mode only
+# needs to clear the chunk it just seeded, so writing through to the end of the array would be O(n)
+# redundant work per chunk, i.e. O(n^2/N) per sweep. `count` mirrors the `chunksize` argument of
+# `seed!(duals, x, index, seeds, chunksize)`.
+function seed_zero_partials!(duals::AbstractArray{Dual{T,V,N}}, x, index,
+                             count = N) where {T,V,N}
+    idxs = Iterators.take(Iterators.drop(structural_eachindex(duals, x), index - 1), count)
+    return _seed_zero_partials!(duals, x, idxs)
 end
 
-# Unseeds at most `N` elements starting at `index`: chunk mode only ever needs to clear
-# the N-wide chunk it just seeded, so writing through to the end of the array would be
-# O(n) redundant work per chunk (O(n^2) per sweep).
-function unseed!(duals::AbstractArray{Dual{T,V,N}}, x, index) where {T,V,N}
+function _seed_zero_partials!(duals::AbstractArray{Dual{T,V,N}}, x, idxs) where {T,V,N}
     seed = zero(Partials{N,V})
-    idxs = Iterators.take(Iterators.drop(structural_eachindex(duals, x), index - 1), N)
     if isbitstype(V)
         for idx in idxs
             duals[idx] = Dual{T,V,N}(x[idx], seed)
--- a/src/apiutils.jl
+++ b/src/apiutils.jl
@@ -27,7 +27,7 @@ end
 function vector_mode_dual_eval!(f!::F, cfg::JacobianConfig, y, x) where {F}
     ydual, xdual = cfg.duals
     seed!(xdual, x, cfg.seeds)
-    seed!(ydual, y)
+    unseed!(ydual, y)
     f!(ydual, xdual)
     return ydual
 end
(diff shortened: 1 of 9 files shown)


ChrisRackauckas-Claude:
CI: all 19 Julia 1 / lts / min-patch jobs green. The 6 "Julia pre" failures are the pre-existing Julia 1.13-rc1 + JET issue that also fails on master — root-cause analysis in https://github.com/JuliaDiff/ForwardDiff.jl/pull/816#issuecomment-4878116764. This commit introduces no new failures relative to the #816 branch it stacks on.

devmotion:
Marked as draft since it depends on #816. Seemingly could be decoupled if supposed to be reviewed more urgently.

ChrisRackauckas-Claude:
Unstacked: force-pushed `narrow-unseed` to a single commit rebased directly onto current `master` (`a26b375`), dropping #816's commit. The change is now expressed against master's structural-only `seed!` (`Iterators.take(…, N)`); the dense-path variant of the same clamp lives in #816 and the two will conflict trivially in `src/apiutils.jl` depending on merge order.

Re-verified locally on the unstacked branch: full `Pkg.test()` passes (9134/9134, Julia 1.12) and chunked results match full-chunk references across n ∈ {5, 12, 13, 24, 25, 100} × chunk ∈ {1, 3, 5, 12}. Benchmarks in the description were re-measured against master alone (the old table's numbers were on top of #816).

src/apiutils.jl
```diff
@@ -106,10 +106,13 @@ function seed!(duals::AbstractArray{Dual{T,V,N}}, x,
     return duals
 end
 
+# Writes at most N elements starting at `index`: chunk mode only ever needs to
+# clear the N-wide chunk it just seeded, so writing through to the end of the
+# array would be O(n) redundant work per chunk (O(n^2) per sweep).
```
devmotion: This comment is confusing - albeit seemingly correct - since the function is not actually used to "seed" anything but to unseed existing seeds (to replace by zero partials). Can we decouple the unseeding and seeding calls? E.g. use `unseed!(duals, x, index)` instead of `seed!` with an implicit `zero(Partials(...))` argument?

ChrisRackauckas-Claude: Agreed — done in b8a9b17.

`seed!(duals, x)` and `seed!(duals, x, index)` are now `unseed!(duals, x)` and `unseed!(duals, x, index)`, and I dropped the `seed::Partials` argument from them rather than keeping it as an implicit zero: nothing in the package ever passed it (the only caller of the explicit-seed form was `AllocationsTest`, which now exercises `unseed!` instead). The `NTuple{N,Partials}` methods keep the `seed!` name and are otherwise untouched, so the two operations are now separate functions rather than one function distinguished by a default argument.

I renamed the whole-array form too, not just the indexed one flagged here — `seed!(xdual, x)` before the chunk loop and `seed!(ydual, y)` for the output buffer are unseeding as well, so leaving that one as `seed!` would have kept exactly the confusion you're pointing at. Happy to narrow it back to the 3-arg method if you'd rather keep the diff minimal.

Verified locally after the rename: full `Pkg.test()` passes (9134/9134, Julia 1.12), including the allocation tests, and the benchmark numbers are unchanged (unseed sweep at n=100000: 2.49 s on master, 0.29 s here).

devmotion (commented):
LGTM. Can you update the version number?

ChrisRackauckas:
Looks like you got it?

devmotion:
I'm rereviewing and I'm about to push a few more commits.

devmotion:
I pushed a commit that revises the API around the fix.

Regarding the origin: `git log -L 109,127:src/apiutils.jl` traces it to fce7f76 (#739), which replaced `dual_inds = (1:N) .+ offset` with an unbounded `Iterators.drop`. So the missing bound seems to have been an oversight rather than intentional - the 4-arg `seed!` kept its `chunksize` bound.

I renamed `unseed!` to `seed_zero_partials!`. Of the 14 renamed call sites, 10 initialize a buffer that was never seeded (`cfg.duals` is uninitialized `similar` memory, and `ydual` in the `f!` paths never carries partials), and only 4 clear a chunk, so `unseed!` seems to describe the minority of cases. I also added an explicit `count = N` argument, mirroring `chunksize` in `seed!(duals, x, index, seeds, chunksize)`, and moved the shared loop into `_seed_zero_partials!` since the two methods only differed in the index iterator.

Additionally, chunk mode now seeds the first chunk and zeroes only the tail (`N + 1`, `xlen - N`) instead of clearing everything and then overwriting the first `N` elements. `xlen > N` always holds in chunk mode, so the windows are disjoint and still cover `1:xlen`. Mostly this is useful because the four `work_array_definition` quotes in jacobian.jl reduce to just unpacking `cfg.duals`.

I don't think the existing tests are sufficient here. They do cover correctness - a too-narrow window would fail `GradientTest.jl:250` and the `BigFloat` tests - but clearing too much is harmless, which is presumably why this could survive since #739. `test/SeedTest.jl` checks the window directly; reverting to the unbounded version fails 4 of the new tests. I also extended the allocation tests with the 4-arg form, where `count` is a runtime value, and turned the second `BigFloat` test into a loop over the position of the unassigned entry - previously it was always in the last chunk, which is never cleared, so `Base._unsetindex!` was never reached in the windowed method.

Tests pass locally (9208/9208, Julia 1.12).

Benchmarks of `gradient!` of `sum(abs2, x)` with default chunk size, against reverting only the `take` bound:

| n | pre-fix | with fix |
|---|---|---|
| 1000 | 0.21 ms | 0.15 ms |
| 10000 | 23.4 ms | 15.0 ms |
| 100000 | 2360 ms | 1582 ms |

That's 1.5x rather than the 1.8x in the description. All timings are much faster here, so the ratio seems to be hardware dependent.

devmotion (approved):
LGTM. Please double check @ChrisRackauckas-Claude @ChrisRackauckas

ChrisRackauckas (approved):

PR #822, opened 2026-07-03
Fix flaky gamma_inc partials test

ChrisRackauckas-Claude:
## Summary

The `gamma_inc` partials assertions in `test/DualTest.jl` fail sporadically on CI —
roughly one job per full-matrix run, on arbitrary OS/Julia-version combinations (e.g.
"Julia lts - windows-latest" in the 2026-06-22 master run, "Julia 1 - ubuntu-latest" in
the 2026-06-07 run), always of the form

```
partials(pq[i]) ≈ PARTIALS * Calculus.derivative(x -> gamma_inc(a, x, ind...)[i], 1 + PRIMAL) rtol=tol
```

with `V === Float32` and random `a`/`PRIMAL` narrowly missing `rtol = 5f-4`.

## Diagnosis

The reference, not ForwardDiff, is at fault, for two reasons:

1. For `V === Float32`, `Calculus.derivative` central-differences the **Float32**
   evaluation of `gamma_inc`, whose finite-difference error is of the same order as the
   `5f-4` tolerance — so random draws sporadically cross it (~4% of draws, measured).
2. For `ind = 1` / `ind = 2`, `gamma_inc` deliberately evaluates a reduced-accuracy
   approximation (~14 / ~6 significant digits). Finite-differencing a function with
   ~1e-6 intrinsic noise amplifies that noise by `1/h`; the measured reference error
   reaches 2000–3000% of the true derivative in the worst draws. The existing
   `tol^(1/2^ind)` loosening (up to `rtol ≈ 0.15`) was compensating for this.

In every sampled failing draw, ForwardDiff's Float32 derivative was within 4e-8
relative of the true derivative (Float64 dual evaluation), while the finite-difference
reference was off by 2%–3000%.

## Fix

ForwardDiff's rule (`src/dual.jl`) is the analytic `exp(-x)·x^(a-1)/Γ(a)`, independent
of `ind`. So compare the partials against a finite difference of the **full-accuracy
(`ind = 0`) Float64** evaluation, at the **base** tolerance — the `ind`-dependent
loosening remains only on the value comparison, where it belongs (values genuinely
differ by the requested approximation accuracy). `Float64(1 + PRIMAL)` converts after
the V-precision addition, so the reference is evaluated at exactly the primal of the
dual input.

Measured over 300000 random `(a, PRIMAL, PARTIALS)` draws × all `ind` variants
(2.4M assertions): old reference fails ~4% of draws; new reference fails **0**, even at
the base (unloosened) tolerance.

[...]

PR #823, opened 2026-07-03
Skip JET QA tests on Julia 1.13 until JET supports it

ChrisRackauckas-Claude:
## Summary

Every "Julia pre" CI job has failed on master since the JET 0.11 compat bump: the
`JET.@test_opt` calls in `test/QATest.jl` report runtime dispatches on Julia
1.13.0-rc1 whose leaf frames are all in **Base** (broadcast `preprocess`,
`simd_index`/`CartesianIndex`, `SubArray` construction under `view`,
`mightalias`/`_isdisjoint`/`_in_tuple`).

These are false positives in JET's Julia 1.13 analysis, not ForwardDiff (or Base)
problems:

- Native inference (`Base.infer_return_type` / `code_typed`) of every flagged frame,
  with the exact argtypes from the reports, is concrete and dispatch-free — and
  identical on 1.12.6 and 1.13.0-rc1.
- The flagged kernels are allocation-free at runtime on both versions.
- The same reports reproduce with self-contained MWEs that don't involve ForwardDiff
  at all (e.g. `@report_opt view(rand(3,3), :, 1:2)` and a plain
  `CartesianIndex(1, CartesianIndex(2))` are flagged on 1.13-rc1, clean on 1.12.6,
  same JET 0.11.5).
- JET's own Julia 1.13 support is still in progress (aviatesk/JET.jl#815, #819), and
  the pending 1.13-compat PR does not fix these reports (verified against that
  branch).

Filed upstream with the reduced MWEs and evidence: aviatesk/JET.jl#839.

This PR gates the JET QA testset on `isempty(VERSION.prerelease)`, so it runs on
every released Julia version and skips on rc/beta/DEV builds, where JET's analysis
routinely lags the compiler (as it does now on 1.13.0-rc1). Comment links the
upstream issue. Test-skip explicitly authorized by @ChrisRackauckas for this case.

Verified locally: on Julia 1.12.6 (`VERSION.prerelease == ()`) the JET testset still
runs (4/4 pass); on 1.13.0-rc1 (`("rc1",)`) it is skipped and `QATest.jl` loads
cleanly.

---

> [!NOTE]
> Opened as a draft by an agent on behalf of @ChrisRackauckas. Please ignore
> until reviewed by @ChrisRackauckas.

### DIFF
--- a/test/QATest.jl
+++ b/test/QATest.jl
@@ -5,12 +5,20 @@ using Test
 
 using JET: @test_opt
 
-@testset "JET" begin
-    # issue #778
-    @test_opt ForwardDiff.derivative(identity, 1.0)
-    @test_opt ForwardDiff.gradient(only, [1.0], ForwardDiff.GradientConfig(only, [1.0], ForwardDiff.Chunk{1}()))
-    @test_opt ForwardDiff.jacobian(identity, [1.0], ForwardDiff.JacobianConfig(identity, [1.0], ForwardDiff.Chunk{1}()))
-    @test_opt ForwardDiff.hessian(only, [1.0], ForwardDiff.HessianConfig(only, [1.0], ForwardDiff.Chunk{1}()))
+# On Julia 1.13, JET produces false-positive runtime-dispatch reports for
+# concrete Base broadcast/view/CartesianIndex code reached by these calls:
+# native inference of every flagged frame is concrete and the kernels are
+# allocation-free, so nothing is actually mis-inferred in ForwardDiff.
+# Re-enable once JET supports Julia 1.13; tracked in
+# https://github.com/aviatesk/JET.jl/issues/839
+if VERSION < v"1.13.0-"
+    @testset "JET" begin
+        # issue #778
+        @test_opt ForwardDiff.derivative(identity, 1.0)
+        @test_opt ForwardDiff.gradient(only, [1.0], ForwardDiff.GradientConfig(only, [1.0], ForwardDiff.Chunk{1}()))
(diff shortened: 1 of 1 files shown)


ChrisRackauckas-Claude:
CI: the gate does its job — all 6 "Julia pre" jobs are green on this branch (they fail on master and on every other open PR). 24/25 jobs pass; the single failure ("Julia min-patch - macOS - NaN-safe disabled") is the known random `gamma_inc` tolerance flake at `DualTest.jl:638`, unrelated to this change and fixed by #822.

ChrisRackauckas-Claude:
Switched the gate to `isempty(VERSION.prerelease)` per review: the JET tests now run on every released Julia and skip on rc/beta/DEV builds generally, instead of hardcoding a 1.13 bound. Verified: runs 4/4 on 1.12.6, skips on 1.13.0-rc1.

devmotion:
Nobody cares about pre, so nobody cares about these test failures - neither in ForwardDiff nor other packages such as StatsFuns or SpecialFunctions pre CI is required to pass successfully. So instead of tweaking anything for CI on pre arguably the more relevant question is why to run CI on pre at all.

ChrisRackauckas:
Well at least in SciML we upstreamed a lot during the last prerelease cycle. I think it's at least makes sense to disable JET and advanced ADs (Enzyme, Mooncake) since they tend to be broken right up until release. Turning off other prereleaese though is a larger discussion.

devmotion:
There's an official way to skip JET tests on new (unsupported) Julia versions now, but I still think the better approach would be to reduce CI usage and remove pre tests.

PR #824, opened 2026-07-16
Don't widen the primal when differentiating `x^y` with a real base

antoine-levitt:
`2^x` is `Float32` for a `Float32` `x`, but `2^Dual{T,Float32,N}` came back `Dual{T,Float64,N}`. Came up on https://github.com/JuliaMolSim/DftFunctionals.jl/pull/33


devmotion (approved):
LGTM, thank you!

PR #828, opened 2026-07-29
Only run JET tests on explicitly supported versions

devmotion:
Supersedes #827. Closes #827.


test/QATest.jl
```diff
@@ -3,14 +3,16 @@ module QATest
...
-using JET: @test_opt
+import JET
 
-@testset "JET" begin
-    # issue #778
-    @test_opt ForwardDiff.derivative(identity, 1.0)
-    @test_opt ForwardDiff.gradient(only, [1.0], ForwardDiff.GradientConfig(only, [1.0], ForwardDiff.Chunk{1}()))
-    @test_opt ForwardDiff.jacobian(identity, [1.0], ForwardDiff.JacobianConfig(identity, [1.0], ForwardDiff.Chunk{1}()))
-    @test_opt ForwardDiff.hessian(only, [1.0], ForwardDiff.HessianConfig(only, [1.0], ForwardDiff.Chunk{1}()))
+if VERSION < v"1.12" || JET.JET_AVAILABLE
```
andreasnoack: `JET_AVAILABLE` is only defined in version 0.12. Is the `VERSION < v"1.12"` condition sufficient to ensure that this branch is only reached with JET@0.12?

devmotion: Yes, JET < 0.12 does not support Julia 1.12, JET >= 0.12 are the only JET versions that support Julia 1.12.

devmotion: The alternative I considered was
```suggestion
if isdefined(JET, :JET_AVAILABLE) || JET.JET_AVAILABLE
```
but I was worried then it might be easier for tests to silently be skipped if a new JET release removes `JET.JET_AVAILABLE`.

andreasnoack (dismissed):

andreasnoack (approved):

PR #829, opened 2026-08-04
Use `round(::Dual, ::RoundingMode)` for rounding functions on 1.11+

JamesWrigley:
The other functions like floor/trunc are already implemented in terms of that function in 1.11: https://docs.julialang.org/en/v1/manual/interfaces/#man-rounding-interface

Fixes these invalidations seen when loading CurveFit.jl on 1.13:
```julia
 inserting floor(::Type{R}, d::ForwardDiff.Dual) where R<:Real @ ForwardDiff ~/.julia/packages/ForwardDiff/IleGP/src/dual.jl:328 invalidated:                                                                                                  
   backedges: 1: superseding floor(::Type{T}, x) where T @ Base rounding.jl:484 with MethodInstance for floor(::Type{Int64}, ::Any) (658 children)                                                                                             
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ForwardDiff"
 uuid = "f6369f11-7733-5829-9624-2563aa707210"
-version = "1.4.4"
+version = "1.4.5"
 
 [deps]
 CommonSubexpressions = "bbf7d656-a473-5ed7-a52c-81e309532950"
--- a/src/dual.jl
+++ b/src/dual.jl
@@ -325,17 +325,18 @@ end
 
 Base.rtoldefault(::Type{D}) where {D<:Dual} = Base.rtoldefault(valtype(D))
 
-Base.floor(::Type{R}, d::Dual) where {R<:Real} = floor(R, value(d))
-Base.floor(d::Dual) = floor(value(d))
+# Base derives floor/ceil/trunc/round from `round(x, ::RoundingMode)`:
+# https://docs.julialang.org/en/v1/manual/interfaces/#man-rounding-interface
+Base.round(d::Dual, r::RoundingMode) = round(value(d), r)
 
-Base.ceil(::Type{R}, d::Dual) where {R<:Real} = ceil(R, value(d))
-Base.ceil(d::Dual) = ceil(value(d))
-
-Base.trunc(::Type{R}, d::Dual) where {R<:Real} = trunc(R, value(d))
-Base.trunc(d::Dual) = trunc(value(d))
-
-Base.round(::Type{R}, d::Dual) where {R<:Real} = round(R, value(d))
-Base.round(d::Dual) = round(value(d))
+# Julia 1.11 added the generic `f(::Type{T}, x)` fallbacks, so these can be
+# dropped once 1.11 is the minimum supported version.
+if VERSION < v"1.11"
+    Base.floor(::Type{R}, d::Dual) where {R<:Real} = floor(R, value(d))
+    Base.ceil(::Type{R}, d::Dual) where {R<:Real} = ceil(R, value(d))
+    Base.trunc(::Type{R}, d::Dual) where {R<:Real} = trunc(R, value(d))
+    Base.round(::Type{R}, d::Dual) where {R<:Real} = round(R, value(d))
+end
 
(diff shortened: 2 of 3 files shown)


devmotion:
Thanks, this is a nice cleanup — getting rid of these invalidations is definitely worth it.

I think the version bound can be narrower than 1.11 though. Base has defined

```julia
trunc(x::Real; kwargs...) = round(x, RoundToZero; kwargs...)
floor(x::Real; kwargs...) = round(x, RoundDown; kwargs...)
ceil(x::Real; kwargs...)  = round(x, RoundUp; kwargs...)
```

for much longer than that (`base/floatfuncs.jl:153-155` on 1.10), and since `round(x::Real, r::RoundingMode=RoundNearest; ...)` (`base/floatfuncs.jl:128`) has a default for `r`, the one-arg `round(d)` re-dispatches to `round(d, RoundNearest)` as well. The 1.10 docstring already recommends this:

> To extend `round` to new numeric types, it is typically sufficient to define `Base.round(x::NewType, r::RoundingMode)`.

I checked on 1.10.6 with a dummy `<:Real` that only defines `round(x, ::RoundingMode)`, and `floor(d)`, `ceil(d)`, `trunc(d)` and `round(d)` all work — only the `f(::Type{T}, d)` variants throw. So what 1.11 actually adds are the generic `trunc/floor/ceil/round(::Type{T}, x)` fallbacks (`base/rounding.jl:474-477`), and IMO only those four need a fallback here:

```julia
# Base derives floor/ceil/trunc/round from `round(x, ::RoundingMode)`:
# https://docs.julialang.org/en/v1/manual/interfaces/#man-rounding-interface
Base.round(d::Dual, r::RoundingMode) = round(value(d), r)

# Julia 1.11 added the generic `f(::Type{T}, x)` fallbacks, so these can be
# dropped once 1.11 is the minimum supported version.
if VERSION < v"1.11"
    Base.floor(::Type{R}, d::Dual) where {R<:Real} = floor(R, value(d))
    Base.ceil(::Type{R}, d::Dual) where {R<:Real} = ceil(R, value(d))
    Base.trunc(::Type{R}, d::Dual) where {R<:Real} = trunc(R, value(d))
    Base.round(::Type{R}, d::Dual) where {R<:Real} = round(R, value(d))
end
```

Two advantages besides the smaller diff: the invalidations are fixed on the LTS as well (as written, 1.10 users keep all eight methods), and `round(d, RoundUp)` then works on all supported versions instead of only on 1.11+ — with the current version it would be a `MethodError` on 1.10, which seems like an unnecessary difference.

Also, AFAICT `@static` isn't needed since both branches only contain ordinary method definitions — a plain `if` is sufficient. (It would matter if the untaken branch used macros or bindings that don't exist on the other version.)

Could you add a few tests as well? The existing ones only cover `Int` and the no-type variants, so

- the explicit rounding modes are untested, in particular `RoundNearestTiesAway`/`RoundNearestTiesUp`/`RoundFromZero`, which only become available with this PR, and
- `f(::Type{T}, d)` with `T<:AbstractFloat` is untested. That one seems worth covering: `round(::Type{T}, x, r)` dispatches to `Base._round_convert`, which for `T<:AbstractFloat` compares `x_t < x`/`x_t > x` and calls `signbit(x)` on the *original* argument (`base/float.jl:471`) — i.e. on the `Dual`, not on `value(d)`. It works, but only because we define the mixed comparisons, and it would be good to notice if that ever changes.

Something like

```julia
for r in (RoundDown, RoundUp, RoundToZero, RoundNearest,
          RoundNearestTiesAway, RoundNearestTiesUp, RoundFromZero)
    @test round(FDNUM, r) === round(PRIMAL, r)
    @test round(NESTED_FDNUM, r) === round(PRIMAL, r)
end
@test floor(Float32, FDNUM) === floor(Float32, PRIMAL)
@test ceil(Float32, FDNUM) === ceil(Float32, PRIMAL)
```

(The failures on `pre` are unrelated to this PR — they're JET bugs, fixed in JET 0.12. Our compat bound still caps at 0.11, so #827 is what actually resolves them; #828 additionally restricts the JET tests to explicitly supported versions.)

JamesWrigley:
Sure, changed the implementation and added those tests in cdafa99.

devmotion (approved):

PR #840, opened 2026-08-17
Extract derivatives at the indices of `x`, and store those indices in the config

### LINKED ISSUE #838: `gradient!` writes to the wrong entries for structured inputs
`extract_gradient!` and `extract_gradient_chunk!` take their target positions from `structural_eachindex(result)`, i.e. from the *result* container, while the seeds are laid out according to `structural_eachindex(x)`. When the two disagree, the derivatives land in the wrong entries.

```julia
julia> using ForwardDiff, LinearAlgebra, DiffResults

julia> x = UpperTriangular([1.0 2.0 3.0; 0.0 4.0 5.0; 0.0 0.0 6.0]);

julia> f(z) = sum(abs2, z) / 2;   # df/dz[i,j] == z[i,j] on the structural entries

julia> ForwardDiff.gradient(f, x)   # correct -- the result is allocated as `similar(x)`
3×3 UpperTriangular{Float64, Matrix{Float64}}:
 1.0  2.0  3.0
  ⋅   4.0  5.0
  ⋅    ⋅   6.0

julia> out = fill(NaN, 3, 3); ForwardDiff.gradient!(out, f, x); out
3×3 Matrix{Float64}:
 1.0  3.0  NaN
 2.0  5.0  NaN
 4.0  6.0  NaN
```

The six derivatives are written to linear positions `1:6` of the dense result instead of to the upper triangle. No error, no warning, and it happens at every chunk size including vector mode.

With a `DiffResult` it errors instead, because `DiffResults.gradient!` does a linear `copyto!` into the `UpperTriangular` buffer that `GradientResult` allocated:

```julia
julia> ForwardDiff.gradient!(DiffResults.GradientResult(x), f, x)
ERROR: ArgumentError: cannot set index (2, 1) in the lower triangular part of an UpperTriangular matrix to a nonzero value (2.0)
```

Same for `LowerTriangular` and `Diagonal`. Introduced in #739.

The fix is presumably to derive the positions from `x` rather than from `result` — `gradient!` has `x` right there. That is what I ended up doing for the Hessian in #837, where the same delegation would otherwise have mis-scattered the gradient stored in a `HessianResult` (whose gradient buffer is dense even when `x` is not).

ForwardDiff v1.4.5, Julia 1.12.6.

### LINKED ISSUE #839: Chunked `jacobian` throws for `Diagonal`/`LowerTriangular`/`UpperTriangular` inputs
`jacobian` allocates its result with `chunksize(cfg)` columns, which is `structural_length(x)`, but `reshape_jacobian` reshapes it to `length(xdual)` columns. For inputs whose structural length differs from their length the two disagree, so chunk mode always throws while vector mode works.

```julia
julia> using ForwardDiff, LinearAlgebra

julia> g(z) = [sum(z), sum(abs2, z)];

julia> x = UpperTriangular(rand(3, 3));

julia> size(ForwardDiff.jacobian(g, x))   # vector mode
(2, 6)

julia> ForwardDiff.jacobian(g, x, ForwardDiff.JacobianConfig(g, x, ForwardDiff.Chunk{2}()))
ERROR: DimensionMismatch: new dimensions (2, 9) must be consistent with array length 12
```

All three wrapper types, only the full-length chunk survives:

| input | chunk=1 | chunk=2 | chunk=`structural_length(x)` |
| --- | --- | --- | --- |
| `UpperTriangular(rand(3,3))` | `DimensionMismatch` | `DimensionMismatch` | `(2, 6)` |
| `LowerTriangular(rand(3,3))` | `DimensionMismatch` | `DimensionMismatch` | `(2, 6)` |
| `Diagonal(rand(3,3))` | `DimensionMismatch` | `DimensionMismatch` | `(2, 3)` |

Since `Chunk(x)` picks the full structural length for anything up to `DEFAULT_CHUNK_THRESHOLD`, this only shows up for larger inputs or an explicit `Chunk`. Introduced in #739.

Worth deciding the column convention at the same time: `jacobian` currently uses structural columns while `gradient` returns a result shaped like `x`. See the discussion in #837, where I argued for indexing the Hessian by the linear indices of `x` on both axes.

ForwardDiff v1.4.5, Julia 1.12.6.

### LINKED ISSUE #842: Reusing a `GradientConfig`/`JacobianConfig` with a differently structured input silently computes wrong derivatives
Two independent bugs in the structure-aware seeding introduced in #739. Both reproduce on `master`
(v1.4.5, 569af35) and on the branch of #840, which addresses neither. I intend to fix both in #840.

## 1. Reusing a config with a differently structured input is silently wrong

`seed!` and `seed_zero_partials!` take the positions to seed from the **config's work buffer**
(`structural_eachindex(duals, x)`), while extraction takes them from somewhere else — from `result`
on `master`, from `x` after #840. When a config is reused with an input of the same length but a
different structure, the two sets disagree and nothing complains:

```julia
using ForwardDiff, LinearAlgebra
using ForwardDiff: GradientConfig, Chunk, gradient

f(z) = sum(abs2, z) / 2                                    # ∇f(z) == z
U = UpperTriangular([1.0 2.0 3.0; 0.0 4.0 5.0; 0.0 0.0 6.0])
A = reshape(1.0:9.0, 3, 3)

cfgU = GradientConfig(f, U, Chunk{2}())
cfgA = GradientConfig(f, A, Chunk{2}())

vec(gradient(f, U, cfgU))   # [1, 0, 0, 2, 4, 0, 3, 5, 6]   ✅ correct
vec(gradient(f, A, cfgA))   # [1, 2, 3, 4, 5, 6, 7, 8, 9]   ✅ correct

vec(gradient(f, U, cfgA))   # [1, 0, 0, 0, 0, 0, 2, 4, 0]   ❌ should be [1, 0, 0, 2, 4, 0, 3, 5, 6]
vec(gradient(f, A, cfgU))   # [1, 4, 5, 7, 8, 9, 0, 0, 0]   ❌ should be [1, 2, 3, 4, 5, 6, 7, 8, 9]
```

Affected:

- **both modes** — with this `x`, `Chunk{2}` is chunk mode and `Chunk{6}` is vector mode; both
  produce exactly the wrong values above;
- **`gradient`, `jacobian` and `hessian`** — `hessian` through both of its sub-configs, e.g.
  `diag(hessian(f, A, HessianConfig(f, U, Chunk{2}())))` is `[1, 0, 0, 0, 0, 0, 0, 0, 0]` where it
  should be all ones;
- all three structured wrappers (`LowerTriangular`, `UpperTriangular`, `Diagonal`), in either
  direction, **including structured→structured**:

  ```julia
  L = LowerTriangular([1.0 0.0 0.0; 2.0 4.0 0.0; 3.0 5.0 6.0])
  vec(gradient(f, U, GradientConfig(f, L, Chunk{2}())))
  # [1, 0, 0, 0, 0, 0, 4, 0, 6]   ❌ should be [1, 0, 0, 2, 4, 0, 3, 5, 6]
  ```

  Note that these two agree on `size` *and* on `structural_length` (both 6), so no count- or
  size-based check can catch this pair.

### …and the work buffer is left partially uninitialized

The reuse does not merely mislabel the output. The buffer entries that the mismatched structure never
visits are **never written at all**, and the target function reads them. Above, the sweep for an
`UpperTriangular` input seeds and clears 6 positions of a work buffer that has 9. With `Float64` that
is silent garbage; a non-bits element type surfaces it:

```julia
Ub = UpperTriangular(big.([1.0 2.0 3.0; 0.0 4.0 5.0; 0.0 0.0 6.0]))
Ab = big.(collect(reshape(1.0:9.0, 3, 3)))

gradient(f, Ub, GradientConfig(f, Ab, Chunk{2}()))
# ERROR: UndefRefError: access to undefined reference
```

This is also why the fix cannot simply be "take the seeding positions from `x` as well": the buffer
entries outside `x`'s structure would then never be initialized. The config has to be checked against
the input instead.

## 2. Independent: `Base._unsetindex!` is linear-index-only, so the unassigned-entry branch of `seed!` mostly does not work

`seed!` and `seed_zero_partials!` have a branch for non-`isbitstype` value types that propagates
unassigned entries of `x` into the buffer via `Base._unsetindex!(duals, idx)`. Base has **no**
`_unsetindex!(::AbstractArray, ::CartesianIndex)` method at all — not even for `Array` — and its
`AbstractArray` fallback for a linear index calls itself forever, since `to_index(::Int)` is the
identity (`_unsetindex!(A, i::Integer) = _unsetindex!(A, to_index(i))`, `abstractarray.jl:1482`). So
that branch only ever worked for a dense `Array`. With one unassigned entry at a structural position
and `Chunk{2}()`:

| `x`, element type `BigFloat` | `ForwardDiff.gradient(f, x)` |
| --- | --- |
| `Matrix` | `UndefRefError` — raised by `f` reading the hole, i.e. seeding worked |
| `adjoint(Matrix)` | `MethodError: no method matching _unsetindex!(::Matrix{Dual{…}}, ::CartesianIndex{2})` |
| `PermutedDimsArray` | same `MethodError` |
| `UpperTriangular`, `LowerTriangular` | `MethodError: no method matching _unsetindex!(::UpperTriangular{Dual{…}}, ::CartesianIndex{2})` |
| `Diagonal` | `StackOverflowError` |

Between them the two failure modes cover the whole of Base's `_unsetindex!` surface, which has
exactly two concrete methods, for `Array` and for `Memory`:

- **No `CartesianIndex` method exists at all.** Rows 2 and 3 are the informative ones here: the buffer
  is a plain dense `Matrix`, because `similar` does not preserve those wrappers, and it still fails.
  Cartesian indices reach the seeding loop whenever *either* array of `eachindex(duals, x)` is
  `IndexCartesian`, so this is not confined to `LinearAlgebra`'s wrapper types.
- **The `AbstractArray` fallback for a linear index recurses forever.** `Diagonal` is the one
  structured case whose positions are already linear (`structural_eachindex(::Diagonal, _)` returns
  `diagind(x)`), so it gets past the first problem and straight into the second.

## Proposed fix (in #840)

1. Store the structural positions in the config, as an indexable vector of **linear** indices, built
   from the work buffer the config owns. This also removes the `Iterators.drop` walk that currently
   re-traverses the lazy triangular position iterators from the front to reach each chunk — three
   times per middle chunk in the gradient sweep, which measures at 35–49% of `gradient!` for an
   `UpperTriangular` input.
2. Validate the config against the input at every API entry point, next to `checktag`. Comparing the
   *structural kind* is O(1) and a compile-time constant, so it can run on every call; a size or
   count comparison is not sufficient, per the `LowerTriangular`/`UpperTriangular` case above.
3. Unset through the array that actually stores the entry, with a linear index.

Making the stored positions linear is what makes (3) expressible, and it fixes the non-bits path for
`adjoint`, `transpose`, `PermutedDimsArray` and non-strided `view`s as well as for the three wrappers
`similar` preserves.

devmotion:
Fixes #838, fixes #839 and fixes #842 — the same omission from #739 three times over: seeding became structure-aware, extraction and the config did not.

## `gradient!` wrote to the wrong entries (#838)

`extract_gradient!`/`extract_gradient_chunk!` took their positions from the *result*, while the seeds are laid out along the structural positions of `x`. They now take the positions of `x`, which the config holds (see #842 below), and entries that receive no derivative are zeroed — that *is* their derivative.

```julia
julia> x = UpperTriangular([1.0 2.0 3.0; 0.0 4.0 5.0; 0.0 0.0 6.0]);

julia> f(z) = sum(abs2, z) / 2;

julia> out = fill(NaN, 3, 3); ForwardDiff.gradient!(out, f, x); out
3×3 Matrix{Float64}:      # master:  1.0  3.0  NaN
 1.0  2.0  3.0            #          2.0  5.0  NaN
 0.0  4.0  5.0            #          4.0  6.0  NaN
 0.0  0.0  6.0

julia> ForwardDiff.gradient!(DiffResults.GradientResult(x), f, x);   # master: ArgumentError
```

This also fixes the mis-scattered gradient of `hessian!(::DiffResult, f, x)`, whose buffer `DiffResults.HessianResult` allocates densely even for a structured `x` — the case from the #837 review.

## Chunked `jacobian` threw (#839)

The result was allocated with `structural_length(x)` columns while `reshape_jacobian` asked for `length(xdual)`. The allocation was the side that was wrong: the Jacobian is indexed by the **linear indices of `x`** — column `j` holds `∂f(x)[i]/∂x[j]`, exactly as the docstring says — with hard zeros in the columns of the structural zeros.

```julia
julia> g(z) = [sum(z), sum(abs2, z)];

julia> size(ForwardDiff.jacobian(g, x, ForwardDiff.JacobianConfig(g, x, ForwardDiff.Chunk{2}())))
(2, 9)                    # master: DimensionMismatch
```

## Reusing a config across structures was silently wrong (#842)

`seed!` derived the positions to seed from the config's *work buffer*, while extraction derived them from `x`. A config reused with an input of the same length and a different structure therefore seeded one set of positions and extracted another — silently — and left the buffer entries outside the input's structure **uninitialized** for the target function to read (`UndefRefError` for a non-bits element type, garbage for `Float64`).

Taking the positions from `x` instead would leave those buffer entries never written at all. So `structural_indices` is derived from the buffer and stored on the config — a new `indices` field, one position vector per buffer a config owns — and every API entry now checks the config against the arrays it is used with, calling `checkstructure` next to `checktag`. A size or count comparison could not replace that check: `LowerTriangular(n, n)` and `UpperTriangular(n, n)` agree on `size` *and* on `structural_length`, yet their positions are disjoint off the diagonal.

```julia
julia> A = collect(reshape(1.0:9.0, 3, 3));

julia> ForwardDiff.gradient(f, A, ForwardDiff.GradientConfig(f, x, ForwardDiff.Chunk{2}()))
ERROR: ArgumentError: the config was built for an array of type UpperTriangular and cannot be
used with an array of type Array: the two are structured differently
# master: [1.0 4.0 5.0; 7.0 8.0 9.0; 0.0 0.0 0.0], silently
```

Separately, seeding's unassigned-entry branch only ever worked for a dense `Array`: `Base._unsetindex!` has no `CartesianIndex` method at all, and its `AbstractArray` fallback for a linear index recurses forever. With the positions now linear, `adjoint`, `transpose`, `PermutedDimsArray` and non-strided `view` inputs behave like a `Matrix`, their buffers being plain `Array`s; the wrappers `similar` preserves raise an `ArgumentError` naming the entry, in place of a `MethodError` (the triangles) or a `StackOverflowError` (`Diagonal`).

## It is also faster

The lazy position iterators were walked from the front to reach each chunk (`Iterators.drop`, which has no range specialization), three times per gradient chunk. Storing them turns that into `view(indices, index:(index + count - 1))`. Dense inputs paid the same walk over `eachindex(duals, x)`, so they speed up too — best-of-200 `gradient!`/`jacobian!`, µs:

| `x` | n | chunk | `gradient!` | `jacobian!` |
| --- | --- | --- | --- | --- |
| `Matrix` | 20 | 2 | 122.0 → 68.5 | 175.4 → 134.8 |
| `Matrix` | 40 | 2 | 1885 → 1108 | 2728 → 2198 |
| `Matrix` | 40 | 12 | 411 → 282 | 636 → 550 |
| `UpperTriangular` | 20 | 2 | 70.4 → 37.7 | 110.7 → 76.2 |
| `UpperTriangular` | 40 | 2 | 1092 → 581 | 1698 → 1168 |
| `UpperTriangular` | 40 | 12 | 253 → 168 | 389 → 294 |

Triangular buffers are now walked by linear index rather than by `CartesianIndex`, which is also what lets a position double as a Jacobian column number.

## Breaking changes

The result conventions follow #839/#837 — index the result by the indices of `x`, hard zeros off the structure — which is what `gradient(f, x)` has always returned. Consequences:

- `jacobian` gains the zero columns: `(2, 6)` → `(2, 9)` for `UpperTriangular(3×3)`.
- `hessian` inherits both conventions through `jacobian(∇f, x)` and becomes `length(x) × length(x)` with hard-zero rows and columns, instead of mixing linear rows with structural columns: `(9, 6)` → `(9, 9)`. `hessian!(DiffResults.HessianResult(x), f, x)` starts working (it allocates `length(x)^2`), as does `jacobian!(DiffResults.JacobianResult(y, x), ...)`.
- For a `Diagonal` the Jacobian/Hessian now scale with `length(x) = n²`, so differentiating with respect to the diagonal vector is the better choice there.
- `gradient!` requires the result to have the same length as `x` and to store every entry `x` does: `gradient!(zeros(6), f, UpperTriangular(3×3))` throws `DimensionMismatch` rather than packing the derivatives in structural order, that order being an internal detail, and a structured result for a dense `x` throws `ArgumentError`. Shape is not constrained beyond the length, so `zeros(9)` for a `3×3` input and an `UpperTriangular` result for a `Diagonal` one both work.
- `jacobian!` into a matrix whose shape is not `length(y) × length(x)` now throws `DimensionMismatch` in chunk mode, where it used to be reshaped whenever the total length happened to match. Vector mode already used such a result as is; the two modes now agree.
- **Reusing a config across structures throws** rather than computing the wrong derivatives (#842 above).
- **The configs gained a type parameter** for the `indices` field, deliberately not shimmed, since pinning the parameter count means depending on internals. Two known cases: SimpleNonlinearSolve constructs `ForwardDiff.JacobianConfig{tag,V,N,typeof(duals)}(seeds, duals)` directly, a hard error, and Pumas types two struct fields as `JacobianConfig{TJ,V,NJ,DJ}` / `GradientConfig{TG,Dual{…},NG,DG}`, which stop being concrete.
- A config for an **offset-indexed** input throws at construction rather than at the first `gradient`/`jacobian` call, since `structural_indices` requires one-based indexing.

## Smaller changes

**Zeroing belongs to the sweep, not to a chunk.** The chunk extractors recognised the first chunk by `index == 1` and used it to zero the whole result — neither part of extracting one chunk nor something a chunk can decide, the entries at stake belonging to no chunk in particular. Both sweeps now zero once up front.

**`reshape_jacobian` allocated a wrapper on every chunked call.** It reshaped the result even when that was already a matrix, and since 1.12 `reshape` can no longer return its argument. It now shares the short-circuit `extract_jacobian!` was given in #797, with an explicit size check stricter than what `reshape` performed on the way past — that only ruled out a wrong total length, so a matrix of the right length and the wrong shape was silently reinterpreted.

**`x` last, one job per function.** `extract_gradient!` and `extract_jacobian!` take `x` after the derivatives, like every other internal function pairing an output with an input; the chunk extractors drop it, since they only write positions, which they now receive. `extract_jacobian!` duplicated the structured branch of `extract_jacobian_chunk!` at `offset == 0` and now delegates to it, and its redundant `n` argument is gone. Both gradient sweeps reject an array-valued `f` up front, as the Jacobian sweeps do, so `gradient` on one points at `jacobian` instead of raising a `MethodError` from `zero(::Type{<:AbstractArray})`.

## Tests

New testsets in `GradientTest.jl` (#838), `JacobianTest.jl` (#839) and `HessianTest.jl` (both, inherited) cover the three wrappers × sizes × every relevant chunk size against a dense result, a result shaped like `x`, `DiffResults.GradientResult`/`JacobianResult`/`HessianResult`, a dense `DiffResult` gradient buffer, and both Jacobian forms. Results are prefilled with `NaN` so untouched entries fail rather than pass, and the expected nonzero positions are written out by hand so a bug in the position mapping cannot hide inside the reference; against master the Hessian testset gets 0 passed, 3 failed, 4 errored. For #842, a config-reuse testset in each of `DerivativeTest.jl`, `GradientTest.jl`, `JacobianTest.jl` and `HessianTest.jl` runs the full four-kinds × four-kinds grid in both modes — including the `LowerTriangular`/`UpperTriangular` pair no size or count check can catch — and asserts the error *messages*, not just the types. `AllocationsTest.jl` asserts zero allocations outright for all four input types in both modes, which is also what pins the stored positions to being indexed through stack-allocated views.

<details>
<summary>Further test detail</summary>

The Hessian test differentiates a function whose second derivative is `1 + (a == b)` on the structural entries, so the reference is exact and the hard-zero rows and columns are asserted rather than approximated. The config-reuse testsets also assert that a wrong *length* still fails as a `DimensionMismatch` and that the fitting config still returns the right answer. `DerivativeTest.jl` had no structure coverage at all before, so it also pins the value and the derivative for a structured `y`.

The unassigned-entry cases are covered end-to-end in `GradientTest.jl` and per-operation in `SeedTest.jl`, which now pins `structural_indices` against the index sets it writes out by hand, asserts that a window overrunning the end is a `BoundsError` rather than a silently truncated chunk, and covers all four `Array`-buffer input kinds plus the three wrappers.

The first version of the allocation test failed on Julia ≤ 1.10 for `Diagonal`, but none of the allocations came from extraction: the target function reduced with the no-function `sum(z)`, and `Base._sum(::Diagonal, ::Colon)` allocates there — 32 bytes even for a `Diagonal{Float64}`. Reducing with `sum(f, z)` measures ForwardDiff rather than LinearAlgebra. (On arm64 macOS with Julia 1.10.12 the older `Test jacobian! allocations` testset reports 16 bytes both here and on `569af35`; they come from the target function's own broadcast over `Vector{Dual}`, and 1.12 reports zero.)

Also asserted: the out-of-place `gradient` returns the structure of `x`; a structured result cannot hold the gradient of a dense `x`, while a result whose structure contains that of `x` can; a wrongly shaped `jacobian!` result and a `gradient!` result of the wrong length both throw; chunk mode fills the `MVector` gradient buffer of an `ImmutableDiffResult`; the structured Jacobian and Hessian testsets run a chunk size leaving a partial final chunk, which every size they had divided; and the JET tests cover a structured input.

Beyond the suite: a sweep over the three wrappers × `n ∈ {1,2,3,5,8}` × every chunk size × every result container agrees with the closed-form expectation, with the Hessian symmetric and its structural rows/columns exactly zero; empty inputs, flat `Vector` results and `BigFloat` inputs with unassigned entries behave. On the GPU side the dense path is unchanged: `partials_wrap` is kept, and `result[:, cols] .= …` targets a contiguous view of the whole array there, `cols` being a `UnitRange` over every column. A `JLArray` run fails identically on this branch and on master, at the scalar indexing in `seed!` that #816 addresses.

</details>

### DIFF
--- a/src/gradient.jl
+++ b/src/gradient.jl
@@ -52,9 +52,9 @@ gradient(f, x::Real) = throw(DimensionMismatch("gradient(f, x) expects that x is
 #####################
 
 # Only the structurally non-zero entries of `x` are seeded, so only those have a derivative to
-# extract. The positions come from the config, i.e. from `x`, not from `result`: the two may have
-# different structure, as they do when `DiffResults.HessianResult` allocates a dense gradient buffer
-# for a structured `x`. Entries of `result` that no position covers are zeroed. See #838.
+# extract. The positions come from `x`, not from `result`: the two may have different structure, as
+# they do when `DiffResults.HessianResult` allocates a dense gradient buffer for a structured `x`.
+# Entries of `result` that no position covers are zeroed. See #838.
 
 function extract_gradient!(::Type{T}, result::DiffResult, y::Real, x, indices) where {T}
     result = DiffResults.value!(result, y)
@@ -69,30 +69,26 @@ function extract_gradient!(::Type{T}, result::MutableDiffResult, dual::Dual, x, 
     return result
 end
 
-# Immutable results cannot be written to entry by entry. Copying the partials wholesale is correct
-# as long as every entry of `x` is seeded, which holds for the `StaticArray` gradient buffers that
-# are the only source of such results; anything else throws on the length mismatch.
+# Immutable results cannot be written to entry by entry, so the partials are copied wholesale, which
+# is only correct when every entry of `x` is seeded.
 function extract_gradient!(::Type{T}, result::ImmutableDiffResult, dual::Dual, x, indices) where {T}
+    check_structural_indices(x, DiffResults.gradient(result))
     result = DiffResults.value!(result, value(T, dual))
     result = DiffResults.gradient!(result, partials(T, dual))
     return result
 end
 
-# Readies `result` for extraction, once per sweep: checks that it can be indexed by the positions of
-# `x`, then zeroes it unless every entry is going to be written anyway. Comparing counts is exact
-# here -- a `result` with the same count but a different structure, an `UpperTriangular` one for a
-# `LowerTriangular` `x`, cannot hold the gradient at all. `dual` is passed for its value type, which
-# unlike `eltype(result)` is a number type even for an `Any` result.
+# Zeroes `result` unless every entry is going to be written anyway. `dual` is passed for its value
+# type, which unlike `eltype(result)` is a number type even for an `Any` result.
 function zero_unseeded!(::Type{T}, result::AbstractArray, dual, x) where {T}
     check_structural_indices(x, result)
     structural_length(x) == structural_length(result) || fill!(result, zero(valtype(T, dual)))
-    return result
+    return nothing
 end
-# Dispatched on `DiffResult`, not on `MutableDiffResult`: a `StaticArray` gradient buffer makes the
-# result immutable even when the buffer itself can be written to, as an `MVector` can.
+# `DiffResult`, not `MutableDiffResult`: `DiffResult(v, ::MVector)` is an `ImmutableDiffResult`.
 function zero_unseeded!(::Type{T}, result::DiffResult, dual, x) where {T}
     zero_unseeded!(T, DiffResults.gradient(result), dual, x)
-    return result
+    return nothing
 end
 
 extract_gradient!(::Type{T}, result::AbstractArray, y::Real, x, indices) where {T} =
@@ -129,6 +125,7 @@ end
 
 function vector_mode_gradient!(result, f::F, x, cfg::GradientConfig{T}) where {T, F}
     ydual = vector_mode_dual_eval!(f, cfg, x)
+    ydual isa Real || throw(GRAD_ERROR)
     result = extract_gradient!(T, result, ydual, x, cfg.indices)
     return result
 end
--- a/src/gradient.jl
+++ b/src/gradient.jl
@@ -16,6 +16,7 @@ Set `check` to `Val{false}()` to disable tag checking. This can lead to perturba
 function gradient(f::F, x::AbstractArray, cfg::GradientConfig{T} = GradientConfig(f, x), ::Val{CHK}=Val{true}()) where {F, T, CHK}
     require_one_based_indexing(x)
     CHK && checktag(T, f, x)
+    checkstructure(cfg, x)
     if chunksize(cfg) == structural_length(x)
         return vector_mode_gradient(f, x, cfg)
     else
@@ -35,6 +36,7 @@ This method assumes that `isa(f(x), Real)`.
 function gradient!(result::Union{AbstractArray,DiffResult}, f::F, x::AbstractArray, cfg::GradientConfig{T} = GradientConfig(f, x), ::Val{CHK}=Val{true}()) where {T, CHK, F}
     result isa DiffResult ? require_one_based_indexing(x) : require_one_based_indexing(result, x)
     CHK && checktag(T, f, x)
+    checkstructure(cfg, x)
     if chunksize(cfg) == structural_length(x)
         vector_mode_gradient!(result, f, x, cfg)
     else
@@ -49,44 +51,68 @@ gradient(f, x::Real) = throw(DimensionMismatch("gradient(f, x) expects that x is
 # result extraction #
 #####################
 
-function extract_gradient!(::Type{T}, result::DiffResult, y::Real) where {T}
+# Only the structurally non-zero entries of `x` are seeded, so only those have a derivative to
+# extract. The positions come from the config, i.e. from `x`, not from `result`: the two may have
+# different structure, as they do when `DiffResults.HessianResult` allocates a dense gradient buffer
+# for a structured `x`. Entries of `result` that no position covers are zeroed. See #838.
+
+function extract_gradient!(::Type{T}, result::DiffResult, y::Real, x, indices) where {T}
     result = DiffResults.value!(result, y)
     grad = DiffResults.gradient(result)
     fill!(grad, zero(y))
     return result
 end
 
-function extract_gradient!(::Type{T}, result::DiffResult, dual::Dual) where {T}
+function extract_gradient!(::Type{T}, result::MutableDiffResult, dual::Dual, x, indices) where {T}
+    result = DiffResults.value!(result, value(T, dual))
+    extract_gradient!(T, DiffResults.gradient(result), dual, x, indices)
+    return result
+end
+
+# Immutable results cannot be written to entry by entry. Copying the partials wholesale is correct
+# as long as every entry of `x` is seeded, which holds for the `StaticArray` gradient buffers that
+# are the only source of such results; anything else throws on the length mismatch.
+function extract_gradient!(::Type{T}, result::ImmutableDiffResult, dual::Dual, x, indices) where {T}
     result = DiffResults.value!(result, value(T, dual))
     result = DiffResults.gradient!(result, partials(T, dual))
     return result
 end
 
-extract_gradient!(::Type{T}, result::AbstractArray, y::Real) where {T} = fill!(result, zero(y))
-function extract_gradient!(::Type{T}, result::AbstractArray, dual::Dual) where {T}
-    idxs = structural_eachindex(result)
-    for (i, idx) in zip(1:npartials(dual), idxs)
+# Readies `result` for extraction, once per sweep: checks that it can be indexed by the positions of
+# `x`, then zeroes it unless every entry is going to be written anyway. Comparing counts is exact
+# here -- a `result` with the same count but a different structure, an `UpperTriangular` one for a
+# `LowerTriangular` `x`, cannot hold the gradient at all. `dual` is passed for its value type, which
+# unlike `eltype(result)` is a number type even for an `Any` result.
+function zero_unseeded!(::Type{T}, result::AbstractArray, dual, x) where {T}
+    check_structural_indices(x, result)
+    structural_length(x) == structural_length(result) || fill!(result, zero(valtype(T, dual)))
+    return result
+end
+# Dispatched on `DiffResult`, not on `MutableDiffResult`: a `StaticArray` gradient buffer makes the
+# result immutable even when the buffer itself can be written to, as an `MVector` can.
+function zero_unseeded!(::Type{T}, result::DiffResult, dual, x) where {T}
+    zero_unseeded!(T, DiffResults.gradient(result), dual, x)
+    return result
+end
+
+extract_gradient!(::Type{T}, result::AbstractArray, y::Real, x, indices) where {T} =
+    fill!(result, zero(y))
+function extract_gradient!(::Type{T}, result::AbstractArray, dual::Dual, x, indices) where {T}
+    zero_unseeded!(T, result, dual, x)
+    return extract_gradient_chunk!(T, result, dual, indices, 1, npartials(dual))
+end
+
+function extract_gradient_chunk!(::Type{T}, result, dual, indices, index, chunksize) where {T}
+    for (i, idx) in enumerate(structural_chunk(indices, index, chunksize))
         result[idx] = partials(T, dual, i)
     end
     return result
 end
 
-function extract_gradient_chunk!(::Type{T}, result, dual, index, chunksize) where {T}
-    offset = index - 1
-    idxs = Iterators.drop(structural_eachindex(result), offset)
-    for (i, idx) in zip(1:chunksize, idxs)
-        result[idx] = partials(T, dual, i)
-    end
+function extract_gradient_chunk!(::Type{T}, result::DiffResult, dual, indices, index, chunksize) where {T}
+    extract_gradient_chunk!(T, DiffResults.gradient(result), dual, indices, index, chunksize)
     return result
 end
-
-function extract_gradient_chunk!(::Type{T}, result::DiffResult, dual, index, chunksize) where {T}
-    extract_gradient_chunk!(T, DiffResults.gradient(result), dual, index, chunksize)
-    return result
-end
-
-extract_gradient_chunk!(::Type, result, dual::AbstractArray, index, chunksize) = throw(GRAD_ERROR)
-extract_gradient_chunk!(::Type, result::DiffResult, dual::AbstractArray, index, chunksize) = throw(GRAD_ERROR)
 
 const GRAD_ERROR = DimensionMismatch("gradient(f, x) expects that f(x) is a real number. Perhaps you meant jacobian(f, x)?")
 
@@ -98,12 +124,12 @@ function vector_mode_gradient(f::F, x, cfg::GradientConfig{T}) where {T, F}
     ydual = vector_mode_dual_eval!(f, cfg, x)
     ydual isa Real || throw(GRAD_ERROR)
     result = similar(x, valtype(T, ydual))
-    return extract_gradient!(T, result, ydual)
+    return extract_gradient!(T, result, ydual, x, cfg.indices)
 end
 
 function vector_mode_gradient!(result, f::F, x, cfg::GradientConfig{T}) where {T, F}
     ydual = vector_mode_dual_eval!(f, cfg, x)
-    result = extract_gradient!(T, result, ydual)
+    result = extract_gradient!(T, result, ydual, x, cfg.indices)
     return result
 end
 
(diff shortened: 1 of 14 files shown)


andreasnoack:
Looks like your PRs need some JET related adjustments

src/gradient.jl
```diff
@@ -49,45 +49,75 @@ gradient(f, x::Real) = throw(DimensionMismatch("gradient(f, x) expects that x is
 # result extraction #
 #####################
 
-function extract_gradient!(::Type{T}, result::DiffResult, y::Real) where {T}
+# Derivatives are only computed with respect to the structurally non-zero entries of `x`, since only
```
KristofferC: Having these huge comments on almost every function getting changed is not great. It's a good practice to read through the comments and really think about if they have a reasonable noise-to-signal ratio and are written so a human actually understands them.

devmotion: I know 🙂 

I noticed a few more issues (eg #842) so I'm still iterating quite a bit on the PR locally (actually it has diverged completely). It's not useful at all to spend time on reviewing it right now.

ChrisRackauckas:
> A JLArray run fails identically on this branch and on master, at the scalar indexing in seed! that https://github.com/JuliaDiff/ForwardDiff.jl/pull/816 addresses.

Can we make sure that gets handled too?

devmotion:
No, this is intentionally not addressed in this PR. The PR's sole purpose is to fix bugs with structural indices (#838, #839, #842).

andreasnoack:
How essential is it to compute derivatives wrt to these structured matrices? It seems to be a bit ambiguous what it actually means, and presumably the current behavior is broken. Is supporting it worth the extra complexity?

devmotion:
I don't think there's any ambiguity for these types.

A `LowerTriangular` has `n(n+1)/2` free numbers, the stored entries. The other positions aren't
variables, they're structural zeros. So the derivative is just the partials wrt the stored entries,
and each of those sits at exactly one position of the matrix. Returning them shaped like `x` is then
the obvious packaging, with zeros elsewhere, which is also what you need for `dot(G, H)` to give the
directional derivative for admissible `H`.

It's also the only well-defined option. With

```julia
f(z) = sum(abs2, z) / 2
k(z) = sum(abs2, z) / 2 + z[1, 2]
```

`f` and `k` are the same function on `LowerTriangular` since `z[1,2]` is always `0`, but their dense
gradients differ at `(1,2)`. So there is no well-defined dense gradient at a triangular point, only
a structured one. If you want the dense one you can pass `Matrix(A)` or `parent(A)`.

I checked the other AD packages on 1.12.7 (Zygote 0.7.12, ReverseDiff 1.17.0, Mooncake 0.5.49,
Enzyme 0.13.199, FiniteDifferences 0.12.34).

`gradient(z -> sum(abs2, z)/2, x)` for `Diagonal`, `LowerTriangular` and `UpperTriangular`:

| | |
| --- | --- |
| Zygote | the structured wrapper |
| Enzyme | the structured wrapper |
| FiniteDifferences | the structured wrapper |
| Mooncake | `Mooncake.Tangent{@NamedTuple{data::Matrix}}` / `{diag::Vector}` |
| ReverseDiff | fails on all three, `AssertionError: IndexStyle(value) === IndexLinear()` |

Same values and same convention everywhere, zeros off-structure. ChainRulesCore does it explicitly
in `ProjectTo`, `$UL(project.parent(dx))` and `Diagonal(project.diag(diag(dx)))`.

Jacobians of `g(z) = [sum(z), sum(abs2, z)/2]`:

| | `Diagonal` | `LowerTriangular` | `UpperTriangular` |
| --- | --- | --- | --- |
| master | `(2, 3)` | `(2, 6)` | `(2, 6)` |
| this PR | `(2, 9)` | `(2, 9)` | `(2, 9)` |
| Zygote | `(2, 9)` | `(2, 9)` | `(2, 9)` |
| ReverseDiff | error | error | error |
| `Enzyme.jacobian` | `MethodError: unsafe_copyto!` | same | same |

With this PR the Jacobians are entrywise identical to Zygote's for all three types, both with the
default chunk size and with `Chunk{2}()`. On master they match neither Zygote nor ForwardDiff's own
`gradient`, which already returns something shaped like `x`.

On whether it's worth the complexity: everything except ReverseDiff supports these inputs, so I
think it's reasonable for users to expect ForwardDiff to as well, and at the moment they get a mix
of wrong values and errors instead (#838, #839, #842).

That said, your question made me wonder whether it would be less complex to follow the design of
FiniteDifferences and Zygote and handle the structure at the API boundary only, with something like
`to_vec` or `ProjectTo`: flatten `x` to its free parameters, run the existing chunk-mode paths on a
plain `AbstractVector`, and unflatten the result at the end. Then seeding, extraction and the config
could stay structure-agnostic, `checkstructure` wouldn't be needed, and other wrappers would work
without further changes. The costs would be an extra reconstruction of `x` per evaluation, and that
`f` would see a rebuilt wrapper rather than `x` itself.

devmotion:
> your question made me wonder whether it would be less complex to follow the design of
FiniteDifferences and Zygote and handle the structure at the API boundary only

A very rough and quick initial experiment leads me to believe that - while it might still be a worthwhile option - it would not really reduce complexity significantly. One would also have to trade performance for simplicity, as the most simple approaches likely lead to reduced performance (due to the unflattening involved in every user-function evaluation). I also noticed that most of the additional LoC in this PR actually also seems to be caused by tests (+438 LoC out of +570 LoC).

Regardless, master is broken in many significant cases (#838, #839, #842), so keeping the current state is not really an option. Additionally, I suspect that the loose typing with `AbstractVector`/`AbstractArray` does not only lead to `setindex!` etc. errors on array types currently not handled explicitly but could in the worst case even lead to silently wrong results for array types where the current mode of seeding violates the structure of the array. The core of ForwardDiff is fundamentally based on the assumption of functions operating on dense arrays without any domain/structure constraints - without enforcing it (or without even being able to enforce it?).

PR #843, opened 2026-08-21
Exploit symmetry in the Hessian, on top of #840

### LINKED ISSUE #253: Hessian should be symmetric
The Hessian of a function f is a symmetric matrix if f is smooth enough.
Is there a way of having this information made available? I guess not just by returning a SymmetricMatrix, due to type stability.

### LINKED ISSUE #836: Take advantage of symmetry in `hessian`

devmotion:
Fixes #253. Fixes #836.

Carries #837's two commits unmodified on top of #840, and answers its second review round.
Supersedes #841. `kc/symmetric_hessian` is untouched.

Based on #840 because the sweep needs its structural positions: `structural_indices` cached on
the config, and an `extract_gradient_chunk!` that takes its positions from `x` rather than from
the result (#838). On master the same review asks would need three new helpers that #840 already
has, all of which would then conflict with it.

## Breaking

For structured inputs the result is now `length(x)` by `length(x)`, both axes indexed by the
linear indices of `x`, with zeros in the rows and columns of the structurally zero entries. So
`hessian(f, UpperTriangular(rand(3, 3)))` is 9x9 with rows and columns 2, 3, 6 zero. That is the
Jacobian's convention, and it makes `DiffResults.HessianResult(x)` the right buffer, so
`hessian!(::DiffResult, f, x)` works for structured inputs instead of erroring. It also makes
`hessian(f, ::Diagonal)` quartic in the size of the diagonal.

`HessianConfig` holds `iseeds`, `oseeds`, `duals` and `indices` instead of a `JacobianConfig` and
a `GradientConfig`. The sweep never reads the Jacobian config's buffer, which only existed because
the Hessian used to be `jacobian(gradient(f), x)`. Dropping it takes `HessianConfig(f, x)` from
1.472 MB to 1.368 MB at `length(x) == 1000`, and the result-aware constructor (which built two
dead buffers) from 1.576 MB. Separate commit, easy to drop.

## Fixes

The result was not reproducible across chunk sizes: diagonal blocks read one triangle,
off-diagonal blocks the other, and the two round differently. At `n = 16` every chunk size below
16 differed from single-block by ~1e-16. Every chunk size is now bitwise equal to the StaticArrays
path.

`hessian!` had stopped checking the result shape when the sweep replaced `extract_jacobian!`, so
`hessian!(fill(NaN, 4, 4), f, rand(3))` silently left row and column 4 untouched.
`reshape_hessian` restores it, mirroring `reshape_jacobian`.

Smaller items: `seed_hessian_chunk!` no longer builds a zero for a layer that was given seeds (288
to 0 bytes per call for `BigFloat` at `N = 3`), and the extension's `reshape` branch, its two
`HESSIAN_ERROR` throws and empty inputs are now covered.

## Numbers

Suite 10420/10420. #837's table re-measured against the old algorithm rebuilt on this branch,
`hessian!` into a preallocated matrix:

| n | rosenbrock | ackley | allocations |
| --- | --- | --- | --- |
| 10 (single chunk) | 1.13x | 1.17x | 0 vs 19856 B |
| 30 | 1.58x | 1.57x | 0 vs 28224 B |
| 100 | 1.86x | 1.86x | 0 vs 142624 B |

## Three review items done differently

- `reshape_hessian` checks `size` like `reshape_jacobian` rather than `axes`, plus one
  `require_one_based_indexing` for the `DiffResult` buffer that no entry point checks.
- The `Diagonal` cost is not in the docstring. Neither `gradient` nor `jacobian` documents
  structured inputs, and the new shape sentence (worded like `jacobian`'s) makes it derivable.
- `SeedTest` does not gain the mixed seed forms. Which layer a seed lands in is enforced by the
  types, and both mixed forms run in every multi-block sweep, so the bitwise chunk-size test
  already covers them. That differs from `seed_zero_partials!`, whose testset exists because
  over-clearing is invisible through the public API.

### DIFF
--- a/ext/ForwardDiffStaticArraysExt.jl
+++ b/ext/ForwardDiffStaticArraysExt.jl
@@ -7,7 +7,7 @@ using ForwardDiff: Dual, partials, npartials, Partials, GradientConfig, Jacobian
                    gradient, hessian, jacobian, gradient!, hessian!, jacobian!,
                    extract_gradient!, extract_jacobian!, extract_value!,
                    vector_mode_gradient, vector_mode_gradient!,
-                   vector_mode_jacobian, vector_mode_jacobian!, valtype, value
+                   vector_mode_jacobian, vector_mode_jacobian!, HESSIAN_ERROR, valtype, value
 using DiffResults: DiffResult, ImmutableDiffResult, MutableDiffResult
 
 @generated function dualize(::Type{T}, x::StaticArray) where T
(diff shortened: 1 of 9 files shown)


devmotion:
Closed in favour of #844.

